PR #80, opened 2026-03-29
Remove disable-model-invocation from all custom skills

### LINKED ISSUE #79: Custom skills missing from autocomplete due to disable-model-invocation flag
## Problem

All custom skills (`/wrap`, `/standup`, `/status`, `/prioritize`, `/continue`, `/merge`, `/lessons`, `/check-acceptance-criteria`) have `disable-model-invocation: true` in their SKILL.md frontmatter. This flag removes the skill from Claude's context, which also **removes it from the `/` autocomplete menu**.

Users must type the full slash command from memory — there's no discoverability.

## Root cause

The `disable-model-invocation: true` flag was intended to prevent Claude from auto-invoking these skills during conversation. However, it also suppresses the skill from appearing in the autocomplete list, which is an unintended side effect of the same flag.

Per Claude Code docs: with `disable-model-invocation: true`, the skill description is removed from context so Claude isn't tempted to use it. But this also means the autocomplete system can't list it.

## Fix

Remove `disable-model-invocation: true` from all custom skills. The descriptions are specific enough that Claude won't invoke them unless the user types the slash command. If any skill triggers too aggressively, we can add the flag back selectively.

## Acceptance Criteria

- [ ] `disable-model-invocation: true` removed from all 8 custom skills
- [ ] All skills appear in `/` autocomplete in new sessions
- [ ] Skills still work when typed manually
- [ ] Claude does not auto-invoke skills inappropriately during normal conversation

auerbachb:
## Summary
- Removes `disable-model-invocation: true` from all 8 custom skill SKILL.md files
- This flag was preventing skills from appearing in the `/` autocomplete menu
- Skills still work when typed manually; this fix adds discoverability

Closes #79

## Test plan
- [ ] All 8 custom skills appear in `/` autocomplete in a new Claude Code session
- [ ] Skills still execute correctly when invoked via `/skillname`
- [ ] Claude does not auto-invoke skills inappropriately during normal conversation

## Summary by CodeRabbit

* **Chores**
  * Updated configuration for multiple AI workflow skills to re-enable model invocation. The skills' visible behavior, instructions, and workflows remain unchanged; only the execution metadata that previously disabled model calls was removed.

### DIFF
--- a/.claude/skills/check-acceptance-criteria/SKILL.md
+++ b/.claude/skills/check-acceptance-criteria/SKILL.md
@@ -1,7 +1,6 @@
 ---
 name: check-acceptance-criteria
 description: Verify all Test Plan checkboxes in a PR against the actual code, check off passing items, and report any failures.
-disable-model-invocation: true
 argument-hint: "[PR number, default: current branch's PR]"
 ---
 
--- a/.claude/skills/continue/SKILL.md
+++ b/.claude/skills/continue/SKILL.md
@@ -1,7 +1,6 @@
 ---
 name: continue
 description: Resume an interrupted or stalled review workflow. Detects where the agent left off — local CR review, push, PR creation, review polling, CR/Greptile feedback processing, thread resolution, merge gate verification, or acceptance criteria — and continues from the next incomplete step automatically.
-disable-model-invocation: true
 ---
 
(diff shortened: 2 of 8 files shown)


auerbachb:
@coderabbitai full review

PR #219, opened 2026-04-10
feat: add PreToolUse hook for .env file protection

auerbachb:
Closes #199

## Summary
Mechanical enforcement of the .env protection rule from safety.md. PreToolUse hook blocks Write/Edit/Bash operations targeting .env files. Defense in depth — even if the agent ignores the rule, the hook blocks the operation.

## Implementation
- `.claude/hooks/env-guard.py` — pure-Python PreToolUse hook, reads tool-invocation JSON from stdin, exits 2 with a stderr message to block.
- Anchored basename regex `(?:^|/)\.env(?:\.[A-Za-z0-9_-]+)?$` matches `.env`, `.env.local`, `.env.production`, etc., and avoids false positives on `environment.ts`, `env-config.json`, `rails_env`, `my.env.backup`, `.envrc`.
- Bash detection is narrowly scoped: only blocks when a mutating binary (`rm`, `mv`, `cp`, `rsync`, `sed`, `chmod`, `tee`, `install`, `truncate`, `shred`, `chown`, `dd`, `touch`, `ln`, `awk`, `perl`, `python`, `ruby`, `node`) or a write-redirect (`>`, `>>`, `2>`, `&>`) appears alongside a `.env` path. `cat .env` / `grep X .env` are not blocked; Read tool is not blocked.
- Fail-open on malformed hook input.
- Registered in `global-settings.json` under PreToolUse with matcher `Write|Edit|MultiEdit|NotebookEdit|Bash`, following the placeholder convention of the other hooks (auto-registered by `session-start-sync.sh`).
- Includes an in-file comment documenting known bypass limitations (variable expansion, `eval`, `bash -c`, here-docs) so readers know this is defense-in-depth, not a sandbox.

## Test plan
- [x] PreToolUse hook blocks Write/Edit on .env files
- [x] Hook handles .env, .env.local, .env.production patterns
- [x] Hook blocks destructive Bash commands targeting .env (rm, mv, cp, rsync, sed, chmod, redirect)
- [x] Hook does NOT block Read operations
- [x] Minimal false positives (doesn't block environment.ts, env-config.json, rails_env, my.env.backup, .envrc)
- [x] Hook registered in global-settings.json PreToolUse

Verified locally with a 30-case test matrix (Write/Edit/MultiEdit block, Bash destructive block, read-only bash allow, false-positive allow, fail-open). All pass.

### DIFF
--- a/.claude/hooks/env-guard.py
+++ b/.claude/hooks/env-guard.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+# .env file protection — PreToolUse hook
+# Blocks Write/Edit/Bash operations that would modify .env files.
+# Defense-in-depth enforcement of the rule in .claude/rules/safety.md.
+#
+# Matches basenames: .env, .env.local, .env.production, .env.test, .env.ci, etc.
+# Does NOT match: environment.ts, env-config.json, rails_env, my.env.backup.
+#
+# Hook protocol: reads tool-invocation JSON from stdin. Exit code 2 with a
+# message on stderr blocks the tool call and surfaces the reason to the agent.
+#
+# Scope and limitations
+# ---------------------
+# `bash_targets_env` is a best-effort static token analyzer. It is NOT a
+# sandbox and can be bypassed by constructs it cannot see statically:
+#   - Variable expansion:        `F=.env; rm "$F"`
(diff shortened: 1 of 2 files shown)


auerbachb:
@coderabbitai review

auerbachb:
@coderabbitai full review

PR #341, opened 2026-04-22
feat(hooks): add pre-commit hook blocking commits on root/main (#323)

### LINKED ISSUE #323: Add pre-commit hook to reject commits on root main (enforce worktree discipline)
## Problem

The rule \"NEVER work on main — always use a worktree\" lives in `CLAUDE.md` and `.claude/rules/`, but it's prose. Agents that don't read those rules (Cursor threads, Codex-style sessions, a tired human) can still commit to `main` in the root checkout. Once that happens, downstream cleanup (e.g., aggressive `/wrap` reset) becomes unsafe.

A local pre-commit hook enforces the rule at the git level — independent of which agent or human is committing.

## Proposed behavior

Install a `pre-commit` hook into the root checkout's `.git/hooks/` via `setup.sh` (idempotent — skip if already installed).

Hook logic:
1. Detect current branch: if not `main`, exit 0 (allow).
2. Detect whether cwd is the root checkout or a worktree (via `git rev-parse --git-dir` — root has `.git/` as a directory; worktrees have `.git` as a file pointing to `worktrees/<name>`).
3. If on `main` AND in the root checkout: exit 1 with error message:
   ```
   ❌ Blocked: you are committing to the root checkout on main.
   The root repo is a bookmark, never a workspace.
   Run EnterWorktree or: git worktree add ../<name> -b issue-N-short-description
   Bypass with --no-verify only for genuine emergencies.
   ```
4. Otherwise: allow.

## Why now

Enables a safer aggressive `/wrap` (separate issue) that resets local `main` hard to `origin/main` without risking destruction of accidental local work.

## Acceptance criteria

- [ ] `setup.sh` installs `.git/hooks/pre-commit` in the root checkout (idempotent)
- [ ] Hook rejects commits on `main` in the root checkout
- [ ] Hook allows commits on any other branch
- [ ] Hook allows commits on `main` inside a worktree (rare but not blocked by this rule)
- [ ] Error message points user at the fix (worktree + branch name convention)
- [ ] `--no-verify` bypass documented (and left functional — don't try to block it)
- [ ] Manual smoke test: attempt a commit on root/main → blocked; attempt a commit on root/feature-branch → allowed; attempt a commit in a worktree → allowed

## Related

- Follow-up: aggressive `/wrap` that hard-resets local main (separate issue)
- Parallel: worktree discipline rule in `cursor-code-config` (separate issue in that repo)

auerbachb:
Closes #323

## Summary

Adds a git-level pre-commit hook that rejects commits on `main` in the root checkout, enforcing the "never work on main" worktree-discipline rule for every committer (human, Claude, Cursor, Codex, or any terminal session).

- **`.claude/git-hooks/pre-commit`** — source hook. Checks the current branch via `git symbolic-ref --short HEAD`; if not `main`, exits 0. Otherwise inspects the `.git` entry at the work-tree root: directory → root checkout → block with guidance; file → linked worktree → allow. Detached HEAD allowed (no branch = no main). `--no-verify` bypass is a git-native contract and is left functional on purpose.
- **`setup.sh` Step 8** — idempotent install into `.git/hooks/pre-commit`. Uses `cmp -s` to detect identical content and skip re-copy; if the file exists but differs (user customization), warns with a `diff -u` hint and leaves it in place.
- **`CONTRIBUTING.md`** — short section documenting the hook behavior, allowed/blocked cases, and the `--no-verify` bypass.

## Design choices

- **`.claude/git-hooks/` directory** (not `.claude/hooks/`) to keep Git hooks separate from Claude Code hooks — two different hook systems with different contracts.
- **`.git` directory-vs-file detection** is more reliable than `git rev-parse --git-dir`, which in worktrees resolves to a real directory (`<main>/.git/worktrees/<name>`) and so doesn't distinguish root from worktree.
- **Copy install** (not symlink) — standard for git hooks; portable and survives `setup.sh` re-runs.

## Test plan

Ran manual smoke tests by invoking the hook directly (the hook is a pure function of `git symbolic-ref` and the `.git` entry type, so no destructive commit is needed):

- [x] `setup.sh` installs `.git/hooks/pre-commit` in the root checkout (idempotent) — verified: fresh install copies; `cmp -s` skips re-install; different content preserved with diff hint.
- [x] Hook rejects commits on `main` in the root checkout — Test B: `cd <root> && bash .claude/git-hooks/pre-commit` → exit 1 + guidance message.
- [x] Hook allows commits on any other branch — Test A: on feature-branch worktree → exit 0.
- [x] Hook allows commits on `main` inside a worktree — Test C: fresh `git worktree add -f /tmp/<name> main` → exit 0.
- [x] Error message points user at the fix (worktree + branch name convention) — message includes `git worktree add ../<name> -b issue-N-short-description` and EnterWorktree guidance.
- [x] `--no-verify` bypass documented and left functional (don't try to block it) — git-native contract; hook contains nothing that would defeat it. Documented in CONTRIBUTING.md and the hook's own header comment.
- [x] Manual smoke test: commit on root/main → blocked; commit on root/feature-branch → allowed; commit in a worktree → allowed — all verified above. (Bonus: Test D — detached HEAD → exit 0.)

## Summary by CodeRabbit

* **New Features**
  * Added a Git pre-commit hook that blocks direct commits to main when working in the repository root; commits on other branches, worktrees, and detached HEAD are allowed. Bypass via `git commit --no-verify`.

* **Documentation**
  * Added docs and setup steps describing hook installation, behavior, and how existing user hooks are preserved or reported during setup.

### DIFF
--- a/.claude/git-hooks/pre-commit
+++ b/.claude/git-hooks/pre-commit
@@ -1,0 +1,58 @@
+#!/bin/bash
+# pre-commit — Block commits on `main` in the root checkout.
+#
+# Worktree discipline: the root repo is a bookmark, never a workspace. All
+# code changes happen in worktrees on feature branches. This hook enforces
+# that at the git level so human and agent commits are caught uniformly.
+#
+# Behavior:
+#   - On a branch other than `main`: allow (exit 0).
+#   - Detached HEAD: allow (no branch = no main).
+#   - On `main` in a worktree: allow — rare, but not this hook's concern.
+#   - On `main` in the root checkout: reject with guidance.
+#
+# Bypass: `git commit --no-verify` still works for genuine emergencies.
+#
+# Installed by setup.sh into .git/hooks/pre-commit.
+
+set -u
+
+branch="$(git symbolic-ref --short HEAD 2>/dev/null || true)"
+
+# Detached HEAD or anything non-main: allow.
+if [[ "$branch" != "main" ]]; then
+  exit 0
+fi
+
+# Root vs. worktree: in the root checkout the `.git` entry at the work-tree
+# root is a DIRECTORY; in a linked worktree it is a FILE containing
+# `gitdir: ...`. `git rev-parse --git-dir` always returns a real directory
+# path (in worktrees it resolves to `<main>/.git/worktrees/<name>`) and so is
+# not usable for this distinction — we check the entry type directly.
+toplevel="$(git rev-parse --show-toplevel 2>/dev/null || true)"
+
+# Bare repo / no working tree: nothing to commit in, allow (hook wouldn't run
+# in a bare repo anyway).
+if [[ -z "$toplevel" ]]; then
+  exit 0
+fi
+
+if [[ -d "$toplevel/.git" ]]; then
+  cat >&2 <<'ERR'
+❌ Blocked: committing to `main` in the root checkout is forbidden.
+
+The root repo is a bookmark, never a workspace. All code changes happen in
+worktrees on feature branches.
+
+Create a worktree:
+  git worktree add ../<name> -b issue-N-short-description
+
+Or use the EnterWorktree tool in Claude Code.
+
+Bypass with `git commit --no-verify` only for genuine emergencies.
+ERR
+  exit 1
+fi
+
+# `.git` is a file → we are inside a worktree. Allow.
+exit 0
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,7 +73,7 @@ See [ARCHITECTURE.md](ARCHITECTURE.md) "Hook Lifecycle" and "Hook Auto-Registrat
 
 ## Git Pre-commit Hook (Worktree Enforcement)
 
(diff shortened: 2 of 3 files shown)


auerbachb:
@coderabbitai Fixed in `c7581ad`: Reworded CONTRIBUTING.md line 76 to "When this hook is installed and not bypassed, it rejects commits made on `main` in the root checkout" so the prose matches the actual contract (no-verify + preserved user hooks are explicit escape hatches).

auerbachb:
@coderabbitai Fixed in `c7581ad`: Now resolves the hooks directory via `git rev-parse --git-common-dir` so the script works from either the root checkout or a linked worktree. Deliberately used `--git-common-dir` instead of the suggested `--git-path hooks` — in a worktree the latter resolves to the per-worktree hooks directory, and a pre-commit hook installed there would not block commits on `main` in the root checkout, which is the whole point of this hook. Comment in the code explains the choice.

auerbachb:
@coderabbitai full review

auerbachb:
@cursor review

PR #630, opened 2026-07-21
fix(#625): enforce field-type contract in session-state.sh to prevent active_agents corruption

### LINKED ISSUE #625: Shared session-state.json's active_agents field found corrupted (literal jq expression string)
## Background

`~/.claude/session-state.json` is shared across every concurrent session/worktree in this repo. A prior known issue (tracked in memory, not yet a GitHub issue) is that `.root_repo` can cause false "wrong checkout" refusals in `polling-state-gate.sh` when concurrent threads share this file.

## Problem

Observed during PR #617's `/wrap` session sweep (issue #587): the `.active_agents` field in `~/.claude/session-state.json` held a literal, unevaluated **string** —

```
" (.active_agents // [] | map(select(.pr_number != 71)))"
```

— instead of an array. This is almost certainly the result of a different concurrent session's `session-state.sh --set` call writing its own jq filter expression as the value, instead of the filter being evaluated and the *result* stored. The affected session/PR (`pr_number != 71`) suggests a session was trying to remove PR #71's entry from `active_agents` and the removal itself malfunctioned.

## Impact

Any script or agent that reads `.active_agents` and performs an array/object operation on it (e.g. `.active_agents | keys`, `.active_agents | map(...)`) will error (`jq: error: Cannot index string with...`) or silently misbehave — for **any** concurrent session, not just the one that caused the corruption. This affects monitor-mode's stale-agent detection and phase-transition logic across the whole repo's active sessions.

## Scope sketch

- Investigate `session-state.sh`'s `--set` handling for `.active_agents` (and similar array/object fields) to find how a raw jq expression string could end up written as a value instead of being evaluated — likely a quoting or argument-passing bug when a caller tries to filter/remove an entry.
- Add write-time validation: reject (or coerce) a `--set` write to `.active_agents` (and other array-typed fields) if the resulting value is not the expected type.
- Consider a read-time guard in shared helpers (`session-state.sh --get`, `merge-gate.sh`, monitor-mode logic) that detects a malformed field type and reports it clearly rather than crashing opaquely.

Linked source: #587 (PR #617 session sweep)

## Implementation Plan

**Root cause (confirmed by reading `session-state.sh` directly):** `--set` only ever writes a *literal* value — it tries `jq empty` on the RHS text and, if that parses as JSON, injects it via `--argjson`; otherwise it falls back to `--arg` (a plain string). There is no mode where the RHS is evaluated as a jq *filter* against the live document. A caller that wanted to remove an entry from `.active_agents` apparently passed the filter expression itself (`(.active_agents // [] | map(select(.pr_number != 71)))`) as the `--set` value instead of first evaluating that filter and passing the resulting JSON array — since the filter text isn't valid JSON, it silently fell into the `--arg` branch and was written verbatim as a string. The correctly-documented pattern (already used in `pr-monitor-and-manage/SKILL.md`) is read → filter locally with `jq` → pass the evaluated JSON array/object as the `--set` value; the corruption came from a caller that didn't follow that pattern.

**Fix implemented in `.claude/scripts/session-state.sh`** (matches CodeRabbit's posted plan):

1. **Field-type contract:** a `known_field_type()` lookup (plain `case` statement, not `declare -A` — bash 3.2, macOS's system bash, has no associative arrays, and this script must stay compatible with it) maps known top-level fields to their expected JSON type per `session-state-schema.json`: arrays (`active_agents`, `polling_jobs`, `polling_failures`, `polling_backoffs`) and objects (`prs`, `cr_quota`, `cr_hourly`, `greptile_daily`, `pmm_in_flight`, `pmm`). A `top_level_key_of()` helper extracts the leading top-level key from any jq path (e.g. `.active_agents[0].id` → `active_agents`), so validation covers sub-path/element writes too. Fields outside the contract remain unvalidated (forward-compatible with "preserve unknown fields").
2. **Write-time guard (`--set`):** after the jq pipeline writes the batch to the temp file (but before the atomic `mv`), every known field touched by the batch is checked against its expected type on the *final* written document — not just the raw `--set` value — so both whole-field writes and element writes are covered by one check. Any mismatch aborts the entire batch, exits 4, and leaves the state file completely unmodified.
3. **Read-time guard (`--get`):** when the requested path is exactly a known top-level field and its stored type doesn't match the contract, print a clear diagnostic to stderr and return a safe default (`[]`/`{}`) on stdout instead of the corrupt value, exiting 0. This lets existing read-modify-write callers keep working unmodified — the very next validated `--set` through that field overwrites the corruption with well-formed data. A genuinely-absent field (type `null`) is not treated as corruption and still returns literal `"null"`, matching every existing caller's `[ "$X" = "null" ] && X='[]'` idiom.
4. **Tests:** added `.claude/scripts/tests/session-state.test.sh` (27 assertions) covering the exact ticket reproduction, valid/invalid array and object writes, sub-path writes, the null-is-not-corruption case, and an end-to-end run of the real `pr-monitor-and-manage` read-filter-write pattern. Wired into `.github/workflows/hook-scripts.yml`.
5. **Docs:** documented the contract in `session-state.sh`'s own header (single source of truth), cross-referenced it from `handoff-files.md`, and added an inline note at the `pr-monitor-and-manage/SKILL.md` call site that produced the original corruption.

Source: CodeRabbit's coding plan posted on this issue (2026-07-21) proposed the same chokepoint-first design (type contract → write-time reject → read-time safe-default) and was implemented directly; the bash-3.2-compatibility constraint (no `declare -A`) was an addition found during implementation.

auerbachb:
## Summary

- Root cause: `session-state.sh --set` only ever writes a *literal* value (JSON-or-string via `--argjson`/`--arg`) — there's no mode where the RHS is evaluated as a jq filter against the live document. A caller apparently passed a filter expression (`(.active_agents // [] | map(select(.pr_number != 71)))`) as the `--set` value instead of first evaluating it and passing the resulting array; since the filter text isn't valid JSON it fell into the `--arg` (string) branch and was written verbatim, corrupting `.active_agents` into a literal string for every concurrent session sharing the file.
- Adds a field-type contract to `session-state.sh`: known top-level fields (arrays `active_agents`/`polling_jobs`/`polling_failures`/`polling_backoffs`, objects `prs`/`cr_quota`/`cr_hourly`/`greptile_daily`/`pmm_in_flight`/`pmm`) must hold the JSON type documented in `session-state-schema.json`.
- **Write-time guard:** `--set` validates the *final* value of any touched known field after the whole batch is applied (covers both whole-field and element/sub-path writes) and rejects the entire batch (exit 4, state file left unmodified) if a known field would end up the wrong type — this is exactly what would have caught the original corruption.
- **Read-time guard:** `--get` on an already-corrupted known field warns on stderr and returns a safe default (`[]`/`{}`) instead of the corrupt value, so existing read-modify-write callers (e.g. `pr-monitor-and-manage/SKILL.md`) keep working and self-heal the corruption on their next validated write.
- Kept the script bash-3.2-compatible (macOS system bash has no `declare -A`) — the type contract is a plain `case`-statement lookup, not an associative array.

Closes #625

## Test plan

- [x] Added `.claude/scripts/tests/session-state.test.sh` (27 assertions) covering: the exact ticket reproduction rejected with state file untouched; valid array/object writes accepted; wrong-type (object/scalar) writes to `active_agents` rejected with prior data intact; sub-path/element writes validated against the final field type; the `prs` object-field contract (including that nulling a nested field doesn't trip it); unknown fields staying unvalidated (forward-compat); `--get` on a pre-corrupted field returning `[]` with a stderr warning; `--get` on an absent field still returning literal `null` (unchanged); and an end-to-end run of the real `pr-monitor-and-manage` read-filter-write pattern.
- [x] Wired the new test into `.github/workflows/hook-scripts.yml`.
- [x] Ran the full existing test suite locally (`infer-pr.test.sh`, `escalate-review.test.sh`, `pr-preflight.test.sh`, `compaction-resume-polling-state-gate.test.sh`) — all pass, no regressions from the new guard.
- [x] Verified `bash -n` syntax and manually exercised every scenario (reject/accept/self-heal/forward-compat) against a temp `$HOME` before writing the automated test.
- [x] Documented the contract in `session-state.sh`'s own header (source of truth), cross-referenced from `handoff-files.md`, and annotated the `pr-monitor-and-manage/SKILL.md` call site that produced the original corruption.
- [x] Manually verified rule-corpus word count stays under the ratchet cap (12,135 / 12,232 words) after the `handoff-files.md` addition.
- [x] Fixed a follow-up finding from CodeAnt's review: bracket-notation top-level paths (`.["active_agents"]`) bypassed the field-type contract because the top-level-key extractor returned an empty string for them. Fixed both the extractor and the `--get` exact-path check to recognize jq's dot-bracket form; added 8 more regression assertions (35 total).

## Note on local review

Both local review CLIs were unavailable this session and were dropped per `cr-local-review.md`'s fallback policy: CodeAnt CLI hung >2 minutes in `analyzing` (and its diff scope was conflated with a concurrent session's unrelated files — a separate CodeAnt CLI concurrency issue, not something in this diff); CodeRabbit CLI hit a rate limit twice, then hung >2 minutes past setup on the one run that got past it. Performed a thorough self-review instead (see commit + this description). GitHub-side review (CodeRabbit/CodeAnt/BugBot/Greptile per `cr-github-review.md`) will still run as normal.

___

## **CodeAnt-AI Description**
**Prevent session state corruption and recover safely from bad field types**

### What Changed
- Writes to known array/object fields are now rejected if they would leave the field with the wrong JSON type, and the state file is left unchanged
- Reading an already-corrupted known field now warns and returns a safe default (`[]` or `{}`) so existing read-modify-write flows can recover
- Added tests that cover rejected bad writes, accepted valid writes, safe reads from corrupted data, and the self-healing read-filter-write flow
- Updated usage notes and handoff guidance to explain the field-type contract and the required read-then-write pattern

### Impact
`✅ Fewer session-state corruption incidents`
`✅ Safer concurrent agent updates`
`✅ Clearer recovery from corrupted state`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>


auerbachb:
@graphite-app re-review

auerbachb:
@coderabbitai Fixed in `65c84f3`: `top_level_key_of()` now also parses bracket-notation top-level paths (e.g. `.["active_agents"]`), and the `--get` exact-path check now recognizes jq's dot-bracket form as equivalent to dot form. Added 8 regression assertions (35 total) covering both the write-time reject and read-time safe-default paths via bracket notation.

auerbachb:
@cursor review

Issue #655, opened 2026-07-21
Scope handoff files per repo — pr-{N}-handoff.json collides across projects


auerbachb:
## Background

Issue #638 scoped `~/.claude/session-state.json` per repo, so PR state now lives at `.repos["owner/name"].prs["84"]` and two projects that both reach PR #84 no longer overwrite each other.

Handoff files were deliberately left out of that change, and they still have the identical defect.

## Problem

`~/.claude/handoffs/pr-{N}-handoff.json` is keyed by bare PR number with no repo component. `pr-84-handoff.json` means "PR #84 in whichever repo wrote it last". Two repos working PR #84 at the same time share one file: Phase A in one project writes findings, threads, and a HEAD SHA that Phase B in the other project then reads as its own.

This is the same collision that produced the symptom already recorded in the wild during a sweep on PR #616:

> pr-84-handoff.json exists but PR #84 does not resolve in this repo — likely a cross-repo artifact from shared session-state.json

That note blamed session-state, which #638 has now fixed. The handoff file named in it was never fixed, so the artifact it describes can still appear.

## Why it was split out

Session-state scoping could be absorbed entirely inside `session-state.sh`, because every caller already reaches the file through that one helper — the path-rewriting trick meant no skill needed editing. Handoff files have no such chokepoint: the `pr-{N}-handoff.json` convention is hard-coded across `polling-state-gate.sh`, the phase protocols, `handoff-files.md`, and several skills, each constructing or globbing the path itself. Renaming it is a mechanical but wide change, and bundling it into #638 would have made a state-layout regression much harder to spot in review.

## Proposed solution

Handoff files carry repo identity, so a phase never reads another project's handoff. Options to weigh:

- Filename scoping: `pr-{owner}-{repo}-{N}-handoff.json`, or a per-repo subdirectory `handoffs/{owner}/{repo}/pr-{N}.json`.
- A `owner_repo` field inside the file plus a read-time assertion — cheaper, but two sessions still contend for one path.

Whichever is chosen, existing handoff files should migrate rather than be discarded, and the lifecycle in `handoff-files.md` (created by Phase A, read by B and C, deleted by the parent after merge) must survive unchanged.

## Acceptance Criteria

- [ ] A handoff file is addressable per repo — two repos at the same PR number never share one
- [ ] Every producer and consumer is updated: `polling-state-gate.sh`, phase protocols, `handoff-files.md`, and any skill that builds the path
[...]

Issue #643, opened 2026-07-21
codeant review returns 403 account-wide: entitlement, not credentials (re-auth does not fix)


auerbachb:
## Summary

`codeant review` returns **403 Access denied** for every file, in every repo, on the `auerbachb` account. The other CodeAnt features work. This makes CodeAnt a non-functional half of the documented dual-CLI local review loop in `.claude/rules/cr-local-review.md`.

Same class of problem as #614 (Graphite silent, check installation/billing).

## Evidence collected 2026-07-21

| Probe | Result |
|---|---|
| `codeant scans orgs` | OK: returns org `auerbachb` + account email `bretton.auerbach@gmail.com` |
| `codeant settings repos` | OK: 63 repos onboarded, including both repos tested |
| `codeant secrets --last-commit` | OK: runs clean, no 403 |
| `codeant review` in `meeting_insights_and_actions` | 403 Access denied |
| `codeant review` in `claude-code-config` | 403 Access denied |

Account-level, not repo-level: two different onboarded repos, identical failure.

## Re-authentication does not fix it

The CLI's 403 message says to run `codeant logout` then `codeant login`. That was done on 2026-07-21:

- `codeant logout` succeeded (nulled `apiKeyV2` in `~/.codeant/config.json`)
- `codeant login` succeeded via browser flow, installed a fresh 36-char token
- `codeant review` returned the **identical** 403

A 403 is authorization, not authentication. The token authenticates correctly against other endpoints, so the credential was never the problem, and the CLI's remediation text is misleading.

## Likely cause

Plan / quota / seat entitlement for the AI review feature specifically. Needs a human to inspect the CodeAnt account, which an agent cannot do:

1. Sign in at https://app.codeant.ai as `bretton.auerbach@gmail.com`
2. Check subscription state and whether the AI code review feature is included and in good standing
3. Check any monthly review quota or seat assignment for org `auerbachb`
4. If billing looks healthy, this is a CodeAnt-side bug worth reporting: valid token, onboarded repo, other endpoints fine, review endpoint 403

## Impact while unresolved

- CodeAnt contributes nothing to local review. CodeRabbit is the only functioning local CLI.
- Per the Timeout and fallback section of `.claude/rules/cr-local-review.md`, CodeAnt should be dropped for the session and noted in the PR body until this clears.
- The false-clean reporting behavior means this went unnoticed. Hardening tracked in #642.

## Acceptance criteria

[...]

Issue #640, opened 2026-07-21
Extend session-state type guard to nested PR fields; repair #542/#544


auerbachb:
## Background
Issue #625 taught `session-state.sh` to reject writes that would leave a known field holding the wrong JSON type. It guards the top-level fields — `active_agents`, `polling_jobs`, `prs`, and friends. That fix works: the `active_agents` corruption it was written for is gone and the field currently holds a valid array.

## Problem
The guard stops at the top level by design, so everything nested inside a PR entry is unvalidated. Two entries in the live file prove it: PRs #542 and #544 both carry a `last_cron_action` that is a bare string where the consuming script expects an object. `infer-pr.sh --root-repo` jq-errors out (exit 4) whenever it walks past them.

This was noticed during PR #73's wrap sweep and written down as *"worth repairing those two entries or fixing infer-pr.sh's jq query to tolerate non-object values."* Nobody did either. It hasn't blocked a session yet only because thread-context inference happens to cover for it — which is exactly the kind of quiet degradation that stays broken for months.

## Proposed solution
The two bad values get repaired, and the type guard reaches the fields that actually broke — not just the top-level ones it was originally scoped to. A malformed nested value gets rejected at write time with a clear error, the same way a malformed top-level one already does, instead of sitting in the file until some unrelated script trips over it.

## Acceptance Criteria
- [ ] `last_cron_action` on `.prs["542"]` and `.prs["544"]` in the live `~/.claude/session-state.json` is repaired to the correct shape (or removed if no longer meaningful)
- [ ] `session-state.sh`'s FIELD-TYPE CONTRACT is extended to known per-PR nested fields — at minimum `last_cron_action`, `preflight_triggered`, `babysit`, `wrap_sweep`, `cr_explicit_triggers`, `digest_streak`
- [ ] `.claude/reference/session-state-schema.json` documents the expected type of each per-PR field so the contract has a single source of truth rather than a hardcoded list
- [ ] Rejection behavior matches the existing contract: exit 4, whole batch rejected, state file unmodified
- [ ] `infer-pr.sh` degrades gracefully on an unexpected `last_cron_action` shape rather than hard-erroring — defense in depth, since old files predate the guard
- [ ] Fields outside the known list stay unvalidated, preserving the "preserve unknown fields" forward-compat convention
- [ ] `--help` documents the extended coverage

## Test Plan
- [ ] `jq '.prs["542"].last_cron_action | type' ~/.claude/session-state.json` → no longer `"string"`
[...]

Issue #639, opened 2026-07-21
session-state.json has no write lock — concurrent sessions lose writes


auerbachb:
## Background
`~/.claude/session-state.json` is written by many things at once: parent agents, Phase A/B/C subagents, `/babysit-pr` watchers, `/pr-monitor-and-manage` ticks, cron jobs, and every `/wrap` sweep. On a busy day a dozen writers touch it, several of them running in genuinely separate sessions that know nothing about each other.

## Problem
There's no lock. `session-state.sh` makes each individual write atomic — it builds the new document in a temp file and moves it into place — but the read-modify-write cycle around that move is unprotected. Two writers that both read the file, each apply their own change, and each write back will silently lose one of the two changes. Last writer wins, no error, no sign anything went missing.

That's the most likely mechanism behind the corruption this file keeps accumulating: a half-formed or unevaluated value lands during an interleaved write, and nothing notices until a sweep trips over it days later. Three separate `/wrap` sweeps have now flagged bad data in this one file. Issue #625 added type validation that catches one shape of bad value after the fact, but validation is a net under the tightrope — it doesn't stop the fall.

## Proposed solution
Concurrent writers take turns. A session that wants to update state waits its brief turn instead of clobbering whatever landed a millisecond earlier, so two agents finishing at the same moment both keep their changes. Nothing about how you invoke anything changes — the file just stops losing writes.

## Acceptance Criteria
- [ ] `.claude/scripts/session-state.sh` acquires an exclusive lock around the whole read-modify-write cycle, not just the `mv`
- [ ] Locking works on macOS — `flock(1)` is not present by default, so use a portable primitive (`mkdir` lockdir, `shlock`, or a documented alternative) and note the choice
- [ ] Lock acquisition has a bounded timeout and a clear failure mode; a writer that cannot acquire the lock exits non-zero with a distinct code rather than writing anyway
- [ ] A stale lock (holder died mid-write) is detectable and recoverable — it must not wedge every future session
- [ ] `--help` and the EXIT STATUS block document the new code and the lock behavior
- [ ] Inline `jq … > tmp && mv tmp session-state.json` patterns elsewhere in the codebase are found and routed through `session-state.sh` — a lock only one writer respects is not a lock
- [ ] `.claude/rules/handoff-files.md` documents that all writes go through the script

## Test Plan
- [ ] Two concurrent `session-state.sh --set` calls to different paths → both changes survive
[...]

PR #659, opened 2026-07-21
fix(#638): scope session-state.json per repo so PR numbers stop colliding

### LINKED ISSUE #638: Scope session-state.json per repo — PR numbers collide across projects
## Background
Every repo you work in and every session you have open share a single orchestration file: `~/.claude/session-state.json`. It tracks per-PR phase, reviewer, head SHA, blockers, and sweep results. It works well when you're in one repo. It starts lying to you the moment you're in two.

## Problem
PRs are filed under their bare number, with nothing recording which repo they belong to. Right now the live file interleaves `claude-code-config` PRs (#536–#636) with `inventory` PRs (#27, #34, #38, #40–42, #57, #59, #64, #68, #70, #73, #74, #80–85) in one flat namespace. Two repos that both reach PR #84 quietly overwrite each other's state — no error, no warning, just one project's tracking data replaced by another's.

This has already happened. A sweep on PR #616 recorded the symptom without recognizing the cause: *"pr-84-handoff.json exists but PR #84 does not resolve in this repo — likely a cross-repo artifact from shared session-state.json."*

The same single-value assumption bites at the top level. `.root_repo` is one scalar for the whole file, so it names whichever repo wrote last. In practice it's usually pointing somewhere else, which is why `polling-state-gate.sh` refuses with a "wrong checkout" error that isn't real — you're in the right place, the shared file just remembers a different one. That refusal is already worked around by hand rather than fixed.

## Proposed solution
State is scoped per repo, so two projects can never collide and a stale pointer from another repo can't misroute your session. Working in one repo stays exactly as it is today; working in several stops silently corrupting itself. Existing state migrates rather than being thrown away — a PR entry that already carries `owner_repo` moves into the right namespace on first read.

## Acceptance Criteria
- [ ] `.prs` entries are keyed or scoped such that PR number is unique *per repo*, not globally (e.g. `.repos["owner/name"].prs["84"]`, or a composite key — pick one and document why)
- [ ] `.root_repo` is per-repo scope rather than a single global scalar, or is removed in favor of the caller passing repo context explicitly
- [ ] `.claude/reference/session-state-schema.json` is updated to match, with the version bumped
- [ ] `.claude/scripts/session-state.sh` `--get`/`--set` accept repo context and resolve scoped paths; document the new usage in `--help`
- [ ] A one-time migration path moves existing flat `.prs` entries into their scoped home using each entry's `owner_repo` when present; entries with no `owner_repo` are preserved, not dropped
- [ ] Every consumer is updated: `polling-state-gate.sh`, `merge-gate.sh`, `infer-pr.sh`, `reviewer-of.sh`, `escalate-review.sh`, and any skill that reads `.prs` directly
- [ ] The `polling-state-gate.sh` false "wrong checkout" refusal no longer reproduces when two repos have live sessions
- [ ] Handoff files (`~/.claude/handoffs/pr-{N}-handoff.json`) have the same collision — decide whether they're in scope here or split to a follow-up, and say which

## Test Plan
- [ ] Write state for PR #84 in repo A and PR #84 in repo B → both persist independently, neither overwrites the other
- [ ] Run `polling-state-gate.sh <N>` from repo A while repo B wrote last → no false "wrong checkout" refusal
- [ ] Run the migration against a copy of the current live file → all 40+ existing PR entries survive, correctly scoped where `owner_repo` exists
- [ ] `session-state.sh --get`/`--set` round-trip under the new scoping
- [ ] A single-repo session behaves identically to today (no regression for the common case)

## Notes / Open questions
- Scoped-key vs. composite-key vs. splitting into per-repo files entirely — the last option kills the collision at the root but loses the single-file convenience and complicates the handoff/recovery flows.
- Related to the concurrency issue filed alongside this: namespacing reduces collisions but does not make concurrent read-modify-write safe.
- Issue #625 fixed the top-level field-type corruption in this same file. This is a structurally different problem in the same place.

## Related Files
- `~/.claude/session-state.json` (live state, not in-repo)
- `.claude/scripts/polling-state-gate.sh`
- `.claude/scripts/infer-pr.sh`

## Related Issues
- Related to #639 (write locking on the same file)
- Related to #640 (nested field-type validation on the same file)
- Related to #625 (top-level field-type contract)

_Captured via /issue-maker._

auerbachb:
## What this fixes

`~/.claude/session-state.json` tracked PRs by bare number with nothing recording which repo they belonged to. The live file interleaved `claude-code-config` PRs (#536–#636) with `inventory` PRs (#27–#85) in one flat `.prs` map, so two repos that both reached PR #84 silently overwrote each other's phase, reviewer, and HEAD SHA — no error, no warning.

It had already happened. A sweep on PR #616 recorded the symptom without naming the cause: *"pr-84-handoff.json exists but PR #84 does not resolve in this repo."*

`.root_repo` had the same single-value defect at the top level: one global scalar naming whichever repo wrote last. That is the documented cause of `polling-state-gate.sh` refusing with a "wrong checkout" error that wasn't real.

PR state is now scoped as `.repos["owner/name"].prs["N"]`, with `root_repo` per-repo alongside it.

## Design decision: scoped key, not composite key or per-repo files

**Chosen: nested scope** — `.repos["owner/name"].prs["84"]`.

- **vs. composite key** (`.prs["owner/name#84"]`): a composite key keeps one flat map, but every consumer must construct and parse the key, and there is nowhere natural to hang per-repo metadata — which the AC also requires for `root_repo`. Nesting gives that for free and lets you dump one repo's state with a single jq path.
- **vs. per-repo files**: this kills the collision at the root, but loses single-file atomic multi-field writes (the helper's whole contract), complicates post-compaction recovery — agents read one known path today — and multiplies the concurrency problem tracked in #639 across N files instead of one.

**The scoping is invisible to callers.** `session-state.sh` rewrites a leading `.prs`/`.root_repo` into the active repo's scope, resolved from `--repo`, `$CLAUDE_SESSION_REPO`, or the cwd's `origin` remote. That decision is what kept this change tractable: **~60 skill call sites already went through the helper and needed zero edits.** Only raw-`jq` consumers changed.

Repo identity comes from the `origin` remote rather than the checkout path, because `git rev-parse --show-toplevel` returns the *worktree* path — so every worktree of a repo would otherwise look like a different repo.

## The false "wrong checkout" refusal

`polling-state-gate.sh` now refuses on repo identity instead of path equality, fixing both false-positive forms: a second repo's session overwriting the global scalar, and a sibling worktree of the *same* repo comparing unequal. A genuinely wrong repo is still refused.

Verified with a negative control — the new regression test reproduces the exact documented failure against pre-fix code (`exit 4`, "refuse to poll from wrong checkout") and passes after.

## Migration

Not optional, and not discarding anything. Legacy flat state migrates in memory on read, persists on write, or explicitly via `--migrate` (`--dry-run` previews). Entries are attributed by `owner_repo`, then by the repo identity of their recorded `root_repo` when that path still resolves, and otherwise preserved under a reserved `_unknown` key (no `/`, so it can never collide with a real `owner/name`). Already-scoped data wins on conflict; a corrupt legacy shape is left in place rather than silently dropped.

**Tested against a copy of the live file: 59/59 entries survive with an identical PR-key set, idempotent on re-run.** Attribution came out 18 claude-code-config / 5 inventory / 32 `_unknown` (entries carrying neither `owner_repo` nor a still-resolvable path). `--infer-candidates` in both `pr-state.sh` and `infer-pr.sh` still surfaces `_unknown` entries — preserving the pre-existing "unknown repo, don't hide it" promise — while never showing a PR known to belong to a *different* repo.

## Handoff files: split to a follow-up, deliberately

`~/.claude/handoffs/pr-{N}-handoff.json` has the identical collision. **Explicitly out of scope here**, tracked as #655. Session-state could be fixed behind one helper; handoff files have no such chokepoint — the filename convention is hard-coded across `polling-state-gate.sh`, the phase protocols, `handoff-files.md`, and several skills. Bundling it would have doubled the blast radius and made a state-layout regression harder to see. A known-gap note is in `handoff-files.md` until #655 lands.

## Bug found in self-review

SSH-form remotes (`git@github.com:owner/name`) parsed to an empty key, which would have left every SSH-remote checkout unscoped in `_unknown` — the collision unfixed for precisely the users who'd never know. All four URL forms now resolve identically, with regression coverage.

## Review tooling note

Both local CLIs were unavailable, so this had a **self-review** fallback per `cr-local-review.md` (which found the SSH bug above):
- **CodeAnt CLI** — persistent account-wide `403` (issue #643); re-auth does not fix it.
- **CodeRabbit CLI** — rate-limited across three attempts, with the cooldown escalating (3 min → 50 s → 17 min). Stopped per the "errors out twice → drop it and note it" rule; continuing would also starve the GitHub merge gate, which shares the same adaptive quota.

The GitHub review apps remain the merge gate.

## Sibling issues

**#640 landed mid-flight (PR #654) and this is rebased on top of it.** That PR reworked the same field-type contract — moving it to a runtime load from `_field_types` in the schema and adding per-PR nested validation — so the two changes had to be integrated, not merely sequenced. All six conflict hunks were semantic. What the integration preserves:

- #640's schema-driven contract is kept whole; my hardcoded type list was dropped in favor of it, with `repos` registered in `_field_types.top_level`.
- #640's nested guard would have been **silently disabled** by this change: its helpers match a leading `.prs`, which scoping rewrites to `.repos[...]`, so every lookup would have returned empty and the guard would have quietly stopped firing. Classification now runs on the caller's original path while the jq check path is scoped. Verified: a bare-string `last_cron_action` is still rejected, now reported at `.repos["org/x"].prs["287"].last_cron_action`, as are the whole-entry and `digest_streak` cases.
- #640's tests in `session-state.test.sh` were updated to the scoped storage location (they asserted `jq '.prs[...]'` directly), with the scope pinned via `--repo` so they do not depend on the checkout running them.

**#639 (write locking) is still open with no PR.** Namespacing reduces collisions but does **not** make concurrent read-modify-write safe — #639 remains necessary and is unaffected by this change.

## Test plan

- [x] Write state for PR #84 in repo A and PR #84 in repo B → both persist independently, neither overwrites the other
- [x] `polling-state-gate.sh <N>` from repo A while repo B wrote last → no false "wrong checkout" refusal
- [x] Sibling worktree of the same repo also polls without a false refusal
- [x] A genuinely unregistered repo is still refused (negative control)
- [x] Migration against a copy of the live file → all existing PR entries survive, correctly scoped where `owner_repo` exists
- [x] Entries with no `owner_repo` are preserved under `_unknown`, not dropped
- [x] Migration is idempotent; `--dry-run` does not write
- [x] `session-state.sh --get`/`--set` round-trip under the new scoping
- [x] A single-repo session behaves identically to today (no regression for the common case)
- [x] Lookalike fields (`.prs_backup`, `.root_repo_history`) are not rewritten
- [x] All four git remote URL forms resolve to the same repo key
- [x] Existing suites still pass: session-state, infer-pr, pr-state, escalate-review, pr-preflight, compaction-resume
- [x] `rule-lint.sh` passes within the committed ratchet cap

Closes #638

## Summary by CodeRabbit

* **New Features**
  * Session state now supports repository-scoped PR information, preventing same-numbered PRs in different repositories from conflicting.
  * Added migration support for converting legacy session data, including a preview option.
  * Improved repository-aware polling, candidate selection, and review-state handling.

* **Documentation**
  * Updated guidance for session state, handoffs, repository scoping, migration, and operational scripts.

* **Tests**
  * Added coverage for migration, multi-repository behavior, scoping, and checkout validation.


auerbachb:
Rebased again — **PR #653 (issue #647) landed mid-flight** and overlapped directly with this change: it fixed the same false "wrong checkout" refusal, by the same insight (compare repo *identity*, not checkout path), while keeping the flat `.prs` layout.

Rather than replace it, this now **builds on it**. `polling-state-gate.sh` here is #653's implementation, adapted to read from the per-repo scope. Specifically kept from #653, because it is more hardened than what I had written:

- `repo_identity()`'s fallback to the shared git common dir, so worktrees of a repo with no `origin` remote still compare equal
- the fail-closed branch for *recorded-but-uncomparable* scoping (its CodeAnt finding) — my version would have passed there
- refusal messages naming the PR and both repos

Adapted for #638: the per-PR reads (`owner_repo`, `root_repo`, `head_sha`, `reviewer`) now go through `session-state.sh` so they resolve against this repo's scope, and `--ensure-session` writes from the resolved checkout so they land in it. One message improved to name the repo whose scope was searched, since under scoping "not registered" is a statement about *this* repo.

The two changes are complementary: #653 stopped the *global* `.root_repo` from being a refusal signal; this removes the global field entirely, so there is no shared scalar left to go stale.

Full suite green on the rebased tree: session-state 67, migration 29, infer-pr 44, multi-repo gate 9, plus pr-state / compaction-resume / escalate-review.

auerbachb:
@coderabbitai Fixed in `c01428a`: read failures now warn on stderr instead of silently defaulting to `{}`, and `same_repo` returns null when the current repo is unresolvable — checked before asserting membership, restoring the documented tri-state.

auerbachb:
@coderabbitai Fixed in `c01428a`: read failures now warn on stderr instead of silently defaulting to `{}`, and `same_repo` returns null when the current repo is unresolvable — checked before asserting membership, restoring the documented tri-state.

auerbachb:
@coderabbitai Fixed in `c01428a`: you're right, refusing a valid checkout is a false positive. Comment corrected.

auerbachb:
@coderabbitai Already addressed — the rebase onto #640 brought in the optional-index form (`.value.last_cron_action.at? // null`), so a malformed non-object entry yields null activity instead of a type error. Resolving.

auerbachb:
@coderabbitai Fixed in `c01428a`: `_unknown` is no longer compared as a concrete identity. When either side is unresolved the check is skipped and falls through to #647's identity rules, whose git-common-dir fallback can still compare such checkouts and which fails closed on its own terms.

auerbachb:
@coderabbitai Fixed in `c01428a`: dropped the duplicate step and retitled the pre-existing one to cover #625 / #640 / #638, so the suite runs once per job.

PR #662, opened 2026-07-21
fix(#639): serialize session-state.json writes with a portable mkdir lock

### LINKED ISSUE #639: session-state.json has no write lock — concurrent sessions lose writes

auerbachb:
## What this changes

`~/.claude/session-state.json` had no write lock. Each individual write was already atomic (`mktemp` + `mv`), but the **read-modify-write cycle around it was not**: two writers that each read the document, each applied their own change, and each wrote back silently lost one of the two changes. Last writer wins, no error, no sign anything went missing — the most likely mechanism behind the corruption three separate `/wrap` sweeps have flagged in this file.

Concurrent writers now take turns. Nothing about how anything is invoked changes.

**Measured, on this branch:** 20 concurrent writers → **20/20** changes survive. The same test with locking disabled → **1/20**. 10 concurrent `greptile-budget.sh --consume` → counter reads exactly 10 (previously lost decrements).

## Locking primitive and why

`.claude/scripts/state-lock.sh` — a sourceable library implementing a **`mkdir`-based lock directory** at `<state-file>.lock`.

**Why not `flock(1)`:** this fleet runs on macOS, which does not ship `flock(1)`. A flock-based lock would no-op or hard-fail on the primary platform — which is exactly what `cr-review-hourly.sh` was already doing: it had an *optional* flock path that printed a warning and proceeded **unlocked** on macOS. That dead path is removed here. `mkdir`'s test-and-create is atomic on POSIX (and on NFS, unlike `O_EXCL` creat).

**Cost of `mkdir` over `flock`:** the kernel does not release the lock when the holder dies, so stale recovery must be explicit. The lock records `pid`, `host`, `epoch`, `started`, `cmd` in `<lock>/owner`. A lock is stale when either (1) it names **this** host and `kill -0 <pid>` says that process is gone, or (2) it is older than `CLAUDE_STATE_LOCK_STALE_AGE` (default 120s) — the fallback for foreign hosts and unparseable metadata. A real critical section is milliseconds, so 120s is ~4 orders of magnitude of headroom.

Breaking is guarded against the classic **double-steal** (two waiters both declare a lock stale, both break it, one then destroys the other's freshly acquired lock): the breaker samples the owner bytes, waits, re-reads, and aborts unless they are byte-identical *and* the lock is still stale — a lock that changed hands writes a different pid/epoch, so it is never broken. The break itself is an atomic rename-aside, which only one racer can win.

### Failure modes (stated deliberately)

| Scenario | Behavior |
|---|---|
| Holder killed mid-write | Next writer sees the dead pid and breaks the lock within one poll. The state file is untouched — the killed writer's partial output only ever existed in its temp file. |
| Holder **alive but wedged** >120s | The lock **is** broken, and that wedged process could in principle still complete its `mv` and lose one interleaved write. Deliberate: never breaking a live holder's lock wedges every future session in the fleet forever, which is the failure this issue exists to prevent. |
| Shared/network `HOME` | Pid liveness is meaningless across hosts; cross-host locks fall back to the age rule alone. |
| Heavy fan-out | `mkdir` contention is not FIFO, so an unlucky waiter can lose several rounds. The 30s default timeout is sized for that tail. |
| Cannot acquire in time | **Exit 6** — distinct from the existing 2/3/4/5 vocabulary — with nothing written. Never falls back to writing unserialized. |

## A lock only one writer respects is not a lock

`grep -rn 'session-state.json' .claude/` found three scripts writing the file directly with their own `jq … > tmp && mv tmp`. All now source the same library and hold the lock across their **whole** cycle (including the read that precedes the write, which is the part that was actually broken):

- `session-state.sh` (`--set`; `--get` deliberately does not lock — `mv` is atomic, so readers stay contention-free)
- `greptile-budget.sh` (lock taken *before* the counter read, so two `--consume` calls can't both read the same value)
- `cr-review-hourly.sh` (replacing the macOS-dead flock path)
- `reviewer-of.sh` (`--sticky`)

Skills and agents already went through `session-state.sh`; no inline-write bypasses remain.

Re-entrancy is handled via an exported `CLAUDE_STATE_LOCK_HELD`, so a holder that shells out to another state-writing script cannot deadlock against itself.

## Notes

- **Single-writer overhead** is two extra process spawns (`mkdir`, `rm -rf`). The hot path was deliberately trimmed to that floor — `$SECONDS` instead of a `date` fork for the deadline, `printf '%(%s)T'` (with a bash 3.2 `date` fallback) for the stamp, `$HOSTNAME` instead of `hostname(1)`, lazy parent-dir creation. End-to-end A/B on a 25-write loop lands inside run-to-run noise on this machine.
- `.claude/rules/handoff-files.md` now states that all writes go through the script, documents exit 6, and was condensed elsewhere so the corpus stays under the `rule-lint` ratchet cap (12230 ≤ 12232).
- **Post-review round (CodeAnt, HEAD `8b13f37`):** two real bugs found and fixed — a read-only `--check` could block behind writers and fail with exit 6, and `state_lock_release` could delete a lock that had just changed hands (empty `owner_pid` also covers another process's mkdir-then-stamp window). 6 regression tests added; 38 total. A third thread was already resolved by the rebase. **Rebased** onto latest `main` (conflicts in `handoff-files.md`, `session-state.sh`, and the CI workflow — all resolved; #640's `SCRIPT_DIR` is now reused for the state-lock lookup).\n- **Local review:** CodeRabbit CLI clean after applying 3 of 5 findings (trap re-wrapping across repeated acquire/release, stricter library-load check, test-quoting hygiene). Two declined: reverting the new CI step (spurious), and retrying internally on exit 6 — that would contradict this issue's AC that a writer which cannot acquire the lock exits non-zero rather than writing anyway; the bounded timeout *is* the retry window. **CodeAnt CLI was dropped for this session** — persistent account-wide 403 entitlement failure (issue #643) that returns a false-clean `issues: []` alongside its own error lines; relying on the CodeAnt GitHub app instead.

Closes #639

## Test plan

- [x] Two concurrent `--set` calls to different paths → both changes survive
- [x] 20 concurrent writers → final file is valid JSON, single top-level document, all 20 writes present with correct values, every writer exits 0
- [x] Kill a writer mid-write (SIGKILL while holding) → lock is recoverable, new write lands, pre-kill data intact, file still valid JSON
- [x] Lock timeout path → exit 6 with the documented message, state file byte-identical, live holder's lock not broken
- [x] Age-based staleness breaks an ancient lock; a fresh foreign-host lock is respected rather than stolen
- [x] Legacy flock artifact (regular file at the lock path) is cleared instead of wedging every future write
- [x] Re-entrancy: a holder shelling out to `session-state.sh` does not deadlock
- [x] `--get` is not blocked while a writer holds the lock
- [x] A rejected write (issue #625 field-type contract, exit 4) still releases the lock
- [x] `grep -rn 'session-state.json' .claude/` → no remaining direct-write bypasses
- [x] Single-writer performance is not meaningfully worse
- [x] Existing suites still green: `session-state.test.sh` (35), `escalate-review.test.sh` (22), and the rest of `.claude/scripts/tests/`
- [x] Read-only `greptile-budget.sh --check` never blocks on the lock (CodeAnt) — succeeds while a writer holds it; a `--check` that must persist still exits 6
- [x] `state_lock_release` deletes only on a positive pid match (CodeAnt) — foreign-pid and empty-owner locks survive our release; normal release still works
- [x] `rule-lint.sh` passes


auerbachb:
@graphite-app re-review

auerbachb:
@coderabbitai Fixed in `8b13f37`: `--check` now reads lock-free and only acquires when a write actually turns out to be needed (cross-day rollover or `--budget` override), re-deriving state under the lock so the write decision is never based on a pre-lock read. `--consume`/`--reset` always write, so they still lock before the read. Regression tests: a read-only `--check` succeeds (exit 0) while another process holds the lock, and a `--check` that must persist still exits 6 rather than writing unserialized.

auerbachb:
@coderabbitai Fixed in `8b13f37`: release now requires a positive `owner_pid == $$` match. You are right that an empty pid also covers the window between another process winning `mkdir` and stamping its owner file — declining to delete is the safe error, since an unreleased lock ages out via the staleness rules whereas a wrongly deleted one admits two writers at once. Acquire correspondingly now treats a failed owner-file write as a failed acquire rather than holding a lock it can never prove it owns. Regression tests cover foreign-pid and empty-owner locks surviving our release, plus the normal release still working.

auerbachb:
@coderabbitai Resolved by the rebase onto main: this file now uses the SCRIPT_DIR resolution introduced by #640 (PR #654), and the state-lock lookup reuses that same $SCRIPT_DIR — so the raw ${BASH_SOURCE[0]%/*} fallback to "." you flagged is gone.

auerbachb:
@graphite-app re-review

Issue #687, opened 2026-07-21
Scope /pm orchestration to the invoking repo, not all active repos


auerbachb:
## Background
`/pm` is the active PM orchestrator — you invoke it *inside* a repo to rank the
backlog, track coding threads, and get suggestions for what to work on next. The
whole mental model is "I'm working in *this* project, help me run *this*
project." A family of sibling orchestration skills — `/pm-clean` (cleanup),
`/pr-monitor-and-manage` and `/wrap` (merges), `/status`, `/monitor` — share the
same underlying work-discovery and session-state machinery.

## Problem
When you run `/pm` inside one repo, it doesn't stay in that repo's lane. It
surfaces — and offers to act on — work from *other* repositories you happen to
have in flight: it proposes cleanup, offers to merge PRs, and suggests "helpful"
actions that belong to entirely different projects. That breaks the core
expectation that invoking `/pm` inside a repo means "manage this repo." At best
it's noise; at worst it invites cross-repo actions (merges, cleanup) you never
intended to authorize from this thread. The leak is behavioral, not cosmetic —
the orchestrator's whole notion of "what's in flight" spans every active repo,
so its rankings, suggestions, and offered actions do too.

## Proposed solution
`/pm` should treat the repository it's invoked in as the entire world by
default. Run `/pm` inside repo X and everything it sees, ranks, suggests, and
offers to do is confined to X — other repos simply don't appear. Because the
orchestration skills all draw from one common discovery + session-state layer,
the invoking-repo constraint should live there once, so every sibling skill
inherits it rather than each re-implementing (and drifting on) its own guard.

## Acceptance Criteria
- [ ] Running `/pm` inside repo X surfaces, ranks, and suggests only
      issues/PRs/threads belonging to X; work from other repos does not appear
      in any default output.
- [ ] `/pm` never *offers* or performs write actions (cleanup, merge, rebase,
      close) against a PR/issue outside the invoking repo.
- [ ] The constraint lives in the shared discovery/session-state layer these
      skills inherit — not as a per-skill patch. Identify the common enumeration
      path first (unfiltered `gh pr/issue list`, worktree enumeration, and/or
      `session-state.json` reads that aggregate across `.repos["…"]`).
- [ ] Sibling skills that inherit that layer (`/pm-clean`,
      `/pr-monitor-and-manage`, `/status`, `/monitor`) also respect the
      invoking-repo boundary — enumerate them and confirm each.
[...]

Issue #682, opened 2026-07-21
Handoff files have the same unlocked read-modify-write profile session-state.json had (#639 follow-up)


auerbachb:
Follow-up from PR #662 (issue #639).

Issue #639 closed by serializing every write to `~/.claude/session-state.json` behind a portable `mkdir` lock (`.claude/scripts/state-lock.sh`). Its Notes section explicitly deferred a second question that was never filed:

> Worth checking whether the same argument applies to `~/.claude/handoffs/pr-{N}-handoff.json`, which has the same multi-writer profile.

## Why it matters

The handoff file has the same shape of problem #639 fixed:

- **Multiple writers.** `polling-state-gate.sh` creates/refreshes it, Phase A creates it, Phase B read-modify-writes it (append arrays, update scalars, preserve unknown fields), Phase C reads it, and the parent deletes it after merge.
- **Read-modify-write, not just write.** `.claude/rules/handoff-files.md` specifies Phase B as "Read-modify-write; append arrays, update scalars, preserve unknown fields" — exactly the pattern where two writers each read, each append, and one append is silently lost.
- **The individual write is already atomic.** `polling-state-gate.sh` uses `mktemp` + `mv`, same as `session-state.sh` did before #639 — atomic write, unprotected surrounding cycle.

The blast radius is smaller than session-state.json (one file per PR, and phases are largely sequential), which is likely why it was deferred. But a lost `threads_replied` or `findings_fixed` append is silent, and the concurrency does exist: `/babysit-pr`, `/pr-monitor-and-manage`, and a Phase B subagent can all touch one PR's handoff.

## Suggested approach

`state-lock.sh` is already generic — it locks any path passed to it (`state_lock_acquire <file>`), so this is mostly wiring, not new machinery:

- Have every handoff writer source `state-lock.sh` and wrap its whole read-modify-write cycle.
- Or route all handoff writes through a single helper (the `session-state.sh` shape), which is the more durable fix and matches the "a lock only one writer respects is not a lock" lesson from #639.

Worth deciding first whether the concurrency is real enough to justify the wiring, or whether the phase sequencing genuinely serializes it — the audit itself is the deliverable.

## Related

- Issue #639 / PR #662 — the session-state.json lock this generalizes
- Issue #655 — per-repo scoping of the same file (naming collisions, separate concern)
- Issue #651 — post-convergence audit of session-state.json

_Filed via /wrap._

PR #661, opened 2026-07-21
feat(#652): body-aware dedup check before /wrap auto-files an issue

auerbachb:
## What this changes

`/wrap` gained autonomous follow-up filing in #633 and filed a duplicate on day one: Issue #647 landed thirteen minutes after Issue #638, proposing `.prs[<N>].root_repo` scoping — a restatement of #638's second acceptance criterion. It wasn't a race. #638 was open and searchable the whole time; the sweep searched **titles only**, and the two issues share almost no title words while overlapping heavily in body text.

This adds a duplicate check that runs before every autonomous filing, and searches bodies.

## Design — the threshold, not the plumbing

The two failure modes are not symmetric:

- **Filed a duplicate** → two tickets for one problem. Noisy, closed in one command.
- **Suppressed a real finding** → the finding is *gone*, buried as a comment on a ticket whose scope never covered it.

So the check is biased toward filing. Suppression requires **all four** of: candidate is OPEN, same primary artifact, a **quotable** acceptance criterion in that issue already covers the finding, and `coverage >= 0.6`. Criterion 3 is load-bearing — the numeric floor is a guard against thin matches, not the test. Anything short of all four files anyway with a `Possibly duplicates #N` line. A degraded or failed search is never a strong match.

New `.claude/scripts/issue-dedup.sh` handles **recall only** — it scores open plus recently-closed issues over title *and* body and emits ranked candidates as JSON. It never decides; the classification rules live in `.claude/reference/autofile-dedup.md` and are applied by the caller.

## Review-tool status

- **CodeRabbit CLI:** clean pass after fixes. First two attempts were rate-limited; the third returned 11 findings, all triaged below.
- **CodeAnt CLI:** unavailable — persistent account-wide 403 (Issue #643). Its empty `issues[]` is the known false clean, not a pass, so it is **not** counted as a clean pass here.

Nine CodeRabbit findings were fixed, including three real bugs: `2>&1` in a command substitution would have spliced `gh` stderr warnings into the JSON and failed the parse; an unexpected scoring exception exited `1`, which is the documented "searched, found nothing" verdict a caller reads as "safe to file without a note"; and `gh` failure on the open scan exited without emitting parseable JSON. Two were declined: removing the CI test registration (registering a new test in `hook-tests` is the established pattern — every prior script test did the same), and wrapping `gh` in `timeout` (not portable; `timeout(1)` is absent on this macOS box, and adding the dependency would break the callers it is meant to protect).

## Test plan

[...]

Issue #647, opened 2026-07-21
polling-state-gate.sh falsely refuses to poll when a concurrent session owns .root_repo


auerbachb:
## Background

`.claude/scripts/polling-state-gate.sh` is the mandatory pre-polling gate in `.claude/rules/cr-github-review.md`. Before polling a PR it compares `session-state.json`'s `.root_repo` against the active checkout and refuses to run on a mismatch, to prevent an agent from polling a PR that belongs to a different repo.

## Problem

`~/.claude/session-state.json` is **shared across all concurrent Claude Code sessions**, but `.root_repo` holds a single value. Whichever session wrote last owns it. Any other session — legitimately working in a different repo — then gets refused:

```
polling-state-gate.sh: session-state .root_repo
(/Users/…/develop/inventory/.claude/worktrees/dazzling-darwin-76e5ed) does not
match active root (/Users/…/develop/claude-code-config/.claude/worktrees/fervent-bose-36ddb4)
— refuse to poll from wrong checkout
```

Observed again on PR #637 (issue #633), 2026-07-21: an `inventory` worktree owned `.root_repo` while the session was in `claude-code-config`. Nothing was actually wrong — the PR, the branch, and the checkout all agreed.

The guard is protecting against a real hazard, but its signal is a single global field that concurrent multi-repo work invalidates. The result is a **false** refusal, and the documented rule (`cr-github-review.md`: run this gate before the first poll tick and every cycle) becomes unfollowable.

## Current workaround

Bypass the gate and call `.claude/scripts/merge-gate.sh <PR>` directly, which does the real merge-readiness work. This loses whatever state validation the gate adds, and it means the mandatory-gate language in the rules is routinely violated in practice.

## Proposed approach

Make the repo-scoping check per-PR rather than global. Options worth weighing during implementation:

- Scope `root_repo` under `.prs[<N>].root_repo` so each tracked PR carries its own checkout, and compare against that instead of the global field. The global value becomes a default/fallback.
- Alternatively, derive the expected repo from the PR itself (`gh pr view --json headRepository`) and compare to the active checkout — no shared state involved.
- Either way, keep the genuine cross-repo hazard blocked: a PR that really does belong to another repo must still be refused.

Whichever shape wins, the concurrent-multi-repo case must stop producing a refusal, since that is the common case and is currently pushing agents off the documented path.

## Acceptance criteria

[...]

PR #698, opened 2026-07-22
fix(#651): post-convergence audit of session-state.json — verify, document, and repair the _unknown residue

### LINKED ISSUE #651: Post-convergence audit of session-state.json after #638/#639/#640/#647 land
## Background
Four separate threads are currently rewriting how `~/.claude/session-state.json` is scoped, locked, and validated — #638 (per-repo scoping), #639 (write locking), #640 (nested type validation), and #647 (per-PR root_repo scoping). #638 and #647 overlap on purpose: the decision was to let both run and reconcile them at rebase time rather than serialize the work.

That's a reasonable call for velocity. It also means nobody has looked at the *result* — four independent changes to one file's structure, landing in whatever order they finish.

## Problem
Each of those four PRs will be reviewed against its own acceptance criteria, and each will pass. None of them is responsible for the question that actually matters afterward: is the file coherent? Overlapping scoping work in #638 and #647 could land two different scoping mechanisms that both technically work. Locking from #639 could guard a path that #638 subsequently restructured around. Validation from #640 could be keyed to field names that no longer exist.

The live file is also carrying accumulated damage independent of this work — entries from at least two repos in a flat namespace, stale PR entries going back months, sweep notes from sessions that ended long ago. Three separate `/wrap` sweeps have flagged problems in it. Nobody has ever done a deliberate pass over the whole thing.

## Proposed solution
Once all four changes have landed, someone reads the resulting file and the scripts that touch it as a single system rather than four diffs, and confirms it actually holds together. Contradictory mechanisms get reconciled, dead entries get cleared out, and the schema on disk matches the schema in the reference. The outcome is a file you can trust and a documented statement of what its shape now is.

## Blocked until
All four of #638, #639, #640, and #647 are merged. Do not start before then — an early pass would be invalidated by whatever lands next, and would add a fifth concurrent thread to a file that already has four.

## Acceptance Criteria
- [ ] Confirm all four of #638 / #639 / #640 / #647 are merged before beginning
- [ ] Read `.claude/scripts/session-state.sh` end to end and confirm the locking, type-validation, and scoping mechanisms compose — no dead branches, no two mechanisms solving the same problem differently
- [ ] Reconcile the #638 / #647 overlap explicitly: exactly one repo-scoping mechanism survives, and any vestige of the other is removed
- [ ] `.claude/reference/session-state-schema.json` matches the actual shape the scripts read and write
- [ ] Audit every entry in the live `~/.claude/session-state.json`: no type-corrupt values remain, all entries are correctly scoped, entries for long-merged PRs are pruned
- [ ] Every consumer (`polling-state-gate.sh`, `merge-gate.sh`, `infer-pr.sh`, `reviewer-of.sh`, `escalate-review.sh`, and any skill reading state directly) works against the post-merge shape
- [ ] `.claude/rules/handoff-files.md` describes the file as it now is
- [ ] Report findings even if nothing needs fixing — "verified coherent" is a valid and useful outcome

## Test Plan
- [ ] `jq empty ~/.claude/session-state.json` → valid JSON
- [ ] Every entry validates against the updated schema
- [ ] `polling-state-gate.sh` runs clean from two different repos with concurrent sessions
- [ ] `infer-pr.sh --root-repo` runs clean (this is the check that currently exits 4)
- [ ] Concurrent-write test from #639 still passes after #638's restructure
- [ ] Type-rejection test from #640 still passes after #638's restructure
- [ ] Backup the live file before any pruning; confirm it is restorable

## Notes / Open questions
- Pruning old PR entries needs a retention rule. Merged and closed for 30+ days is a reasonable starting point, but sweep notes in old entries sometimes carry unactioned follow-ups — check before deleting rather than pruning blind.
- Worth deciding whether this audit should become a recurring maintenance pass (a `/pm-clean` or `/memory-clean` sibling for orchestration state) rather than a one-off.
- Handoff files under `~/.claude/handoffs/` have the same multi-repo collision; #638 was asked to decide whether they're in scope. Pick up whatever it deferred.

## Related Files
- `.claude/scripts/polling-state-gate.sh`
- `.claude/scripts/infer-pr.sh`
- `~/.claude/session-state.json` (live state, not in-repo)

## Related Issues
- Blocked by #638, #639, #640, #647
- Related to #625 (the first fix in this series)

_Captured via /issue-maker._

auerbachb:
## What this does

Post-convergence audit of `~/.claude/session-state.json` and every script that touches it, read as one system rather than as the four diffs that produced it.

**The code is coherent — that half of the issue is a "verified coherent" outcome, not a fix.** The defect is in the live file: **scoping was never retroactive.**

Closes #651

## Convergence confirmed (AC 1)

| Issue | PR | Merged (UTC) |
|-------|-----|--------------|
| #640 nested field-type validation | #654 | 2026-07-21T19:32:23Z |
| #647 per-PR `root_repo` scoping | #653 | 2026-07-21T19:47:07Z |
| #638 per-repo storage scoping | #659 | 2026-07-21T20:20:56Z |
| #639 portable write lock | #662 | 2026-07-21T21:00:27Z |

**#687** (`--session-view`, PR #694) landed mid-audit and is folded in.

## What the audit found

**The mechanisms compose.** The lock is acquired *before* the read and held through the `mv`, which is the property that matters — the lost-update race is between read and write, so a lock covering only the write would be decorative. #640's nested guard classifies on the caller's **original** path and scopes only the concrete check path, so #638 moving the data one level down did not quietly disable it.

**#638 and #647 are not duplicates** — the specific worry in the ticket. #638 decides *which scope to read*; #647 decides *whether this checkout may poll that entry*. Delete either and something real breaks. Both survive deliberately.

The genuine vestige was **prose, not logic**: four comments in `polling-state-gate.sh` still called `.root_repo` "the single global field… owned by whichever session wrote last", which #638 made false. Corrected; behavior unchanged.

**37 of 73 live PR entries were stranded in `_unknown`** — more than any real repo held. #638's migration attributes by `owner_repo` or a still-resolvable `root_repo`; legacy entries have neither, so they landed in the reserved bucket. Consumers then disagree about that bucket: `infer-pr.sh`/`pr-state.sh` merge it into **every** repo's candidates (collision intact), while `--session-view` hides it.

A latent bug follows: `--session-view`'s `active_agents` filter treats `_unknown` as "another repo", so an agent on claude-code-config PR #662 would be filtered out of that repo's own view because `_unknown` also tracked 662. `.active_agents` is `[]` today, so nothing is being lost — re-attribution removes the condition rather than special-casing the filter.

## Live repair (already applied)

`session-state-audit.sh --apply --reattribute` moved **25 entries** into their real scopes. Backup: `~/.claude/session-state.json.bak.20260722T123838Z`.

Verified independently: **zero PR keys lost** (73→72 only because PR #661 existed in both buckets and merged per the documented conflict rule). The two PR #57 entries — `auerbachb/inventory` and `auerbachb/longlove` — stayed distinct, which is exactly the collision SHA-based attribution exists to get right.

The remaining 12 have no recorded SHA; the tool refuses to guess. **Zero entries were pruned** — everything is inside the 30-day window.

## Design calls

- **Attribution is by commit SHA, never PR number.** A SHA is globally unique and checkable; a PR number is precisely the ambiguous key that caused the original collision. Zero or 2+ matches leave the entry put.
- **Retention alone does not authorize a delete.** Entries carrying unactioned `wrap_sweep.needs_decision` notes are withheld and their notes printed, prunable only via `--prune-with-notes` — the ticket's "check before deleting rather than pruning blind", made mechanical.
- **Recurrence:** the ticket asked whether this should become a maintenance pass. It should, so the repair is a re-runnable `/memory-clean` sibling rather than hand-rolled jq.

## Note on my own bug

The first integrity check used `index(.pr)`, which jq resolves against the *array being searched* rather than the element. It errored to stderr and left the result empty — **the check passed by failing**, while the write went through unchecked. Caught on the live run, fixed to fail closed, and pinned with a negative control.

## Reviewers

CodeRabbit CLI: 6 findings, 4 applied, 2 declined with rationale in the commit message. **CodeAnt CLI: dropped** — persistent account-wide 403 entitlement cap (issue #643), one attempt, no re-auth attempted (the CLI's `logout`/`login` advice is hardcoded and does not fix it). The CodeAnt GitHub App is unaffected.

## Test plan

- [x] All four prerequisite PRs confirmed merged before the audit began (AC 1)
- [x] `session-state.sh` read end to end; locking, type validation, and scoping compose with no dead branches (AC 2)
- [x] #638/#647 overlap reconciled explicitly; the surviving vestige removed (AC 3)
- [x] `session-state-schema.json` matches what the scripts read and write (AC 4)
- [x] Live file audited: no type-corrupt values, entries correctly scoped (AC 5)
- [x] Every consumer verified against the post-merge shape, including #687's `--session-view` (AC 6)
- [x] `handoff-files.md` describes the file as it now is (AC 7)
- [x] Findings reported in `.claude/reference/session-state-convergence-audit.md` (AC 8)
- [x] `jq empty ~/.claude/session-state.json` passes
- [x] `infer-pr.sh --root-repo` runs clean (exits 1 "multiple candidates", not the exit 4 the ticket recorded)
- [x] #639 concurrency suite passes post-#638: `state-lock.test.sh` 41/41
- [x] #640 type-rejection suite passes post-#638: `session-state.test.sh` 92/92
- [x] New `session-state-audit.test.sh`: 83/83, wired into `hook-scripts.yml`
- [x] Backup taken before the live repair and confirmed restorable
- [x] `rule-lint.sh` passes (11,462 words vs cap 11,671)

## Out of scope

Handoff files under `~/.claude/handoffs/` share the same unlocked read-modify-write profile and global-namespace collision. Tracked separately as #682, deliberately not addressed here.

___

## **CodeAnt-AI Description**
**Add a repeatable audit and repair flow for session-state.json**

### What Changed
- Added a new audit command that reports session-state problems without changing the file by default
- The audit can now repair `_unknown` entries by moving them to the correct repo when their commit SHA matches only one repo
- It can also prune long-merged or closed PR entries, while keeping entries with unread sweep notes unless that extra prune is explicitly requested
- Type drift can now be healed by resetting broken array/object fields to safe empty values, with a backup saved before any edit
- Updated related polling and handoff guidance to reflect that scoping is per-repo, not a single global field
- Added tests and CI coverage for the new audit, repair, backup, lock, and refusal behaviors

### Impact
`✅ Fewer misplaced PR entries in session state`
`✅ Safer cleanup of stale session data`
`✅ Clearer audit reports before repair`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>


auerbachb:
@graphite-app re-review

auerbachb:
@codeant-ai review

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

auerbachb:
Fixed in `ddd1cb9` — all three verified against the code and pinned with regression tests (suite now 79 assertions).

**Malformed repo scope aborts the audit (Critical)** — correct, and a good catch: `.value.prs` was dereferenced before `.value`'s type was checked in both `scopes` and `total_prs`, so an array or string where a scope object belongs killed the entire jq program. The tool that exists to report that corruption was the thing it took down. This is the same class as a PR-entry guard added one commit earlier — I fixed the inner level and missed the outer one. Both are now guarded (plus `REAL_REPOS` derivation and the `_unknown` entry list), a bad scope counts as zero PRs and is *reported* as a type violation rather than being fatal, and a negative control asserts no `Cannot index` error reaches the caller.

**`gh pr list --limit 300` hides older entries (Major)** — agreed on the defect; I took a different fix than pagination. The bulk list is only an optimization, and what matters is the PRs session-state actually tracks, so the entries the listing missed are now queried directly (`gh pr view`) — bounded by tracked state rather than by repo size, and usually zero extra calls. Each gap is logged to stderr, since a silent cap reads as "covered everything" when it did not.

**Repair plan races the lock (Critical)** — correct and the most serious of the three. I kept detection outside the lock deliberately: attribution and staleness make many `gh` calls, and holding the lock across them would push every other writer past its 30s timeout. So instead of moving detection under the lock, the plan is now re-validated against the locked document immediately before applying — a move target that vanished, a heal another writer already performed, or a prune target that gained unactioned `wrap_sweep.needs_decision` notes is dropped and reported. Re-validation can only shrink the plan, never widen it. The regression test drives a real concurrent writer (a `gh` stub that deletes the move target mid-attribution) and asserts the stale move is dropped, reported, and leaves untargeted state intact.

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

auerbachb:
Fixed in `551f784` — both valid, both pinned with regression tests (suite now 83).

**Lowercase-only SHA filter (Major)** — correct. `test("^[0-9a-f]{7,40}$")` silently discarded an uppercase `head_sha`, and the entry then fell into the "no recorded commit SHA" bucket and stayed stranded in `_unknown` permanently — the precise outcome this tool exists to prevent. Now matched case-insensitively and normalized with `ascii_downcase` before the lookup. The regression test stores an uppercase SHA while the stub only answers to the lowercase form, so it passes only if normalization actually happens.

**`RC == 0 || RC == 4` assertion (Major)** — you are right, and this one stings: that assertion passes whichever branch executes, so it cannot fail. It is the third instance of the same shape in this PR — a guard that reports success by not running. The contract is specific (a malformed entry is *data to report*, so the run completes), and it is now asserted exactly: `RC == 0`, the malformed entry conserved as-is, and its healthy sibling still moving.

Tightening it surfaced a real gap I would otherwise have shipped. The nested-field pass has to skip a non-object PR entry — it cannot index one — so such an entry was never reported *at all*. Crash-proof but invisible is not the same as handled. Malformed PR entries are now surfaced as type violations in their own right, with `found`/`want` recorded like any other.

PR #641, opened 2026-07-21
fix(#634): portable date parsing in babysit-pr TTL/age checks

auerbachb:
## Summary

`.claude/skills/babysit-pr/SKILL.md`'s T0 dispatch-in-flight TTL check and A2 duplicate-watcher freshness check both computed elapsed minutes with GNU-only `date -u -d "<iso8601>"`. On macOS/BSD `date`, that flag doesn't exist, the command fails, and `|| echo 0` silently substituted epoch 0 — producing an `AGE_MIN` in the tens of millions instead of erroring loudly. That made genuinely fresh dispatches/watchers look stale, causing T0 to reclaim an in-progress `/fixpr`/`/wrap` dispatch mid-run (duplicate dispatch risk) and A2 to arm a duplicate watcher on top of a live one.

Reproduced live during a `/babysit-pr` tick on PR #628 (2026-07-21 ~13:15 ET): a dispatch ~3 minutes old computed as `AGE_MIN=29744235`.

## Changes

- Added a shared `to_epoch()` helper (GNU `date -d`, then BSD `date -j -f`, returns non-zero on failure instead of a fabricated value) used by both T0 and A2.
- Both call sites now fail loud (stderr warning) instead of silently corrupting the age math, and default to the *safe* direction on a parse failure (treat as still-active, not stale) — the direction that avoids duplicate dispatches/watchers.
- Added a bounded `bump_parse_failure_counter()` helper + persisted counters (`dispatch_parse_failures`, `last_tick_parse_failures`) so a *genuinely* corrupted timestamp (not just a platform gap — that's now fully fixed) self-heals after `BABYSIT_PARSE_FAIL_LIMIT` (default 3) consecutive failures instead of wedging the dispatch guard / watcher slot forever. This also came from CodeRabbit's local-review pass (see below) and generalizes cleanly to both call sites.
- Updated the documented `.babysit` state shape and A3/T4's state-seeding to include the two new counter fields.

## Scope check — sibling skills

Grepped `fixpr`, `wrap`, and `pr-monitor-and-manage` for the same GNU-only `date -d` pattern — **none found**, all three are clean.

A broader sweep of `.claude/skills/` and `.claude/scripts/` found other `date -d` usages, but all of them already attempt a BSD form (`-v`/`-j -f`) either before or after the GNU form, so they're already portable. This bug was specific to `babysit-pr/SKILL.md`'s two occurrences, which never attempted a BSD form at all.

## Local review

[...]

Issue #634, opened 2026-07-21
babysit-pr: GNU-only `date -u -d` silently corrupts TTL/age math on macOS (epoch-0 fallback)


auerbachb:
## Problem

In `.claude/skills/babysit-pr/SKILL.md`, both the **T0** dispatch-in-flight TTL check and the **A2** duplicate-watcher freshness check compute an elapsed-minutes value using:

```bash
AGE_MIN=$(( ( $(date -u +%s) - $(date -u -d "$STARTED" +%s 2>/dev/null || echo 0) ) / 60 ))
```

`date -u -d "<iso8601>"` is **GNU-date-only syntax**. On macOS/BSD `date`, this flag doesn't exist, the inner command fails, and `|| echo 0` silently substitutes **epoch 0** as the "started" timestamp — producing an `AGE_MIN` in the tens of millions of minutes instead of erroring loudly.

## Reproduction

Observed live during a `/babysit-pr` tick on PR #628 (2026-07-21 ~13:15 ET): a dispatch that was genuinely ~3 minutes old computed as `AGE_MIN=29744235`. Since both checks compare `AGE_MIN` against a TTL/freshness window, the massive bogus value makes a **genuinely fresh** dispatch/watcher look **stale**:

- **T0:** would incorrectly reclaim an in-progress `/wrap`→`/fixpr` dispatch mid-run, killing the dispatch-in-flight idempotency guard and allowing a duplicate dispatch to launch on top of the still-running one.
- **A2:** would incorrectly treat a live watcher as dead, allowing a second `/babysit-pr` watcher to be armed on the same PR (double-watching).

## Fix

Replace the GNU-only `date -u -d` calls with a portable ISO-8601 → epoch helper that tries GNU syntax first, then BSD/macOS syntax, and **fails loudly** (a stderr warning, not a fabricated epoch) if both fail — with the epoch-failure default chosen to match the *safe* direction for each check (treat "unparseable" as "still active/in-flight," not as "definitely stale"), since that's the direction that avoids duplicate dispatches/watchers:

```bash
to_epoch() {
  # Portable ISO-8601 UTC (e.g. 2026-07-21T17:13:05Z) -> epoch seconds.
  # Tries GNU date, then BSD/macOS date. Returns non-zero (no stdout) if both
  # fail -- callers MUST check the exit code, never assume a numeric result.
  local iso="$1"
  date -u -d "$iso" +%s 2>/dev/null && return 0
[...]

PR #699, opened 2026-07-22
fix(#648): replace silent epoch-0 date fallback in backlog-staleness.sh with explicit to_epoch() failure

### LINKED ISSUE #648: backlog-staleness.sh: silent epoch-0 fallback on malformed timestamp (same class as #634)
Follow-up surfaced by /wrap session sweep on PR #641.

## Problem

While fixing GNU-only `date -u -d` portability in [PR #641](https://github.com/auerbachb/claude-code-config/pull/641) / #634, a related but narrower instance of the same failure class was noticed and deliberately left out of that PR's scope:

`.claude/scripts/backlog-staleness.sh:285`:

```bash
AGE_DAYS=$(( ( $(date -u +%s) - $(date -u -j -f '%Y-%m-%dT%H:%M:%SZ' "$UPDATED" +%s 2>/dev/null || date -u -d "$UPDATED" +%s 2>/dev/null || echo 0) ) / 86400 ))
```

Unlike `babysit-pr/SKILL.md`'s bug, this line already tries **both** BSD (`-j -f`) and GNU (`-d`) forms, so it's portable across platforms for well-formed timestamps — it's not a systematic macOS-vs-Linux failure. But if `$UPDATED` is genuinely malformed (neither format parses), the `|| echo 0` last-resort still silently substitutes epoch 0, producing a bogus (huge) `AGE_DAYS` instead of failing loud — the same risk class fixed in #634, just triggered by a narrower condition (bad data rather than platform mismatch).

## Suggested fix

Reuse the `to_epoch()`-style pattern from #634 (try both formats, `return`/fail explicitly with no fabricated value on failure, log a warning) instead of the silent `|| echo 0`. Decide the safe default direction for `backlog-staleness.sh`'s specific use (staleness/age-based cleanup decisions) — likely "treat as fresh / don't flag as stale" is the safer default, mirroring #634's "unparseable ⇒ treat as still-active" direction, so a parse failure doesn't cause premature cleanup of something that isn't actually stale.

_Filed via /wrap._

auerbachb:
## Summary

`.claude/scripts/backlog-staleness.sh`'s inactive-issue `AGE_DAYS` computation tried BSD then GNU `date` to parse an issue's `updatedAt`, but silently fell back to epoch 0 (`|| echo 0`) if both parses failed. A genuinely malformed timestamp would therefore produce a bogus, huge `AGE_DAYS`, risking flagging a healthy issue as ancient/stale. Same failure class as #634 (fixed in PR #641 for `babysit-pr`), triggered here by bad data rather than a platform mismatch.

## Changes

- Added a `to_epoch()` helper (mirrors the `to_epoch()` pattern from PR #641 / `babysit-pr/SKILL.md`): tries GNU `date -u -d` then BSD `date -u -j -f`, returns non-zero with no stdout if both fail — never fabricates a value.
- Rewired the `AGE_DAYS` computation to use `if UPDATED_EPOCH=$(to_epoch "$UPDATED"); then ... else ...`. On success, `AGE_DAYS` is computed and the `inactive` flag is emitted unchanged. On failure, emits an `err()` warning identifying the issue number and the raw unparseable value, and **skips flagging that issue as inactive** — mirroring #634's "unparseable ⇒ treat as still-active" direction so bad data never causes premature cleanup recommendations.
- Confirmed via `grep` that this was the only `|| echo 0`-style silent date fallback in the script — no other instances of this failure class exist here.
- Added a `Test4c` case to the existing `.claude/scripts/tests/backlog-staleness.test.sh` (following the Test4a/4b conventions) pinning both directions: a well-formed timestamp still computes a real age (existing Test4a/4b), and a malformed timestamp warns on stderr and is never flagged inactive.

## Local review

- **CodeRabbit CLI** (`coderabbit review --agent --type uncommitted`): clean pass, 0 findings, no error record.
- **CodeAnt CLI**: not installed on this machine (`codeant` not found in PATH, no `~/.codeant/config.json`) — dropped for this session per the local-review fallback policy. The CodeAnt **GitHub App** will still review this PR and can satisfy the merge gate independently.

## Test plan

- [x] `bash .claude/scripts/tests/backlog-staleness.test.sh` passes (17/17, including new Test4c malformed-timestamp cases)
- [x] `bash .claude/scripts/tests/backlog-health.test.sh` still passes (18/18) — confirms composition with the modified script is unaffected
- [x] Well-formed `updatedAt` still computes a real `AGE_DAYS` and emits the `inactive` flag (Test4a, unchanged)
- [x] Malformed `updatedAt` never fabricates epoch 0 — no `inactive` flag is emitted, and a stderr warning identifies the issue number (new Test4c)
- [x] No other silent epoch-0-style date fallback exists elsewhere in `backlog-staleness.sh` (verified via `grep -n "|| echo 0"`)

Closes #648

### DIFF
--- a/.claude/scripts/backlog-staleness.sh
+++ b/.claude/scripts/backlog-staleness.sh
@@ -50,6 +50,17 @@ err() {
   printf 'backlog-staleness.sh: %s\n' "$1" >&2
 }
 
+to_epoch() {
+  # Portable ISO-8601 UTC (e.g. 2026-07-21T17:13:05Z) -> epoch seconds.
+  # Tries GNU date, then BSD/macOS date. Returns non-zero (no stdout) if both
+  # fail -- callers MUST check the exit code, never assume a numeric result
+  # (a silently fabricated epoch previously corrupted TTL/age math -- #634).
+  local iso="$1"
+  date -u -d "$iso" +%s 2>/dev/null && return 0
+  date -u -j -f "%Y-%m-%dT%H:%M:%SZ" "$iso" +%s 2>/dev/null && return 0
+  return 1
+}
+
 DAYS=30
 EMIT_JSON=0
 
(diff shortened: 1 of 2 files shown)


auerbachb:
@greptileai

auerbachb:
Verified against real GitHub API data (both REST /issues/648 and gh issue view --json updatedAt): GitHub always returns whole-second ISO-8601 timestamps for this field (e.g. 2026-07-21T19:02:03Z), never fractional seconds. Since backlog-staleness.sh's $UPDATED comes exclusively from gh issue list --json updatedAt, this input never occurs in practice. This BSD format string also predates this PR — it's the same '%Y-%m-%dT%H:%M:%SZ' pattern already established in babysit-pr/SKILL.md's to_epoch(), which issue #648 explicitly asked this PR to mirror. Even in the hypothetical case, the failure direction is the safe one (skip + warn, per this issue's 'unparseable => treat as still-active' design), not the dangerous one #648 fixes. Not changing the code for this — flagging as a verified non-issue for this script's actual input source.

PR #702, opened 2026-07-22
fix(#665): make issue-maker's chip state visible to /wave and /pm

### LINKED ISSUE #665: Chip tracking gap: /issue-maker and /wave can double-offer the same issue
Follow-up from PR #645.

## Background
`/issue-maker` (PR #645) now offers a coding chip for every issue it creates, tracking each chip's `task_id` in its own per-thread session log (`~/.claude/handoffs/issue-maker-<session>-log.json`). `/wave` (issue #636, PR #644) offers chips for backlog issues too, but its Step 2 dedup only checks `/pm`'s Active Work table for rows already at `Chip offered` / `Inline` / `Active` / `Prompt generated`.

## Problem
These two chip-tracking stores don't talk to each other. If `/issue-maker` offers a chip for a freshly created issue, and the same issue later shows up in a `/wave` candidate pool (e.g. re-ranked into the backlog), `/wave` has no visibility into issue-maker's session log and could offer a second, redundant chip for the same issue. The reverse can also happen if `/wave` offers a chip for an issue that a separate `/issue-maker` thread later edits via `/update`.

## Proposed solution
Whichever chip-offering skill runs second should be able to see that the other already has a live, unclicked chip out for a given issue — either by checking a shared record, or by having `/issue-maker` also write into `/pm`'s Active Work table when running inside a PM thread (the same pattern `/prompt` already follows). Land second, check first.

## Acceptance Criteria
- [ ] Define where issue-maker's chip state becomes visible to /wave and /pm (shared record, or write-through to Active Work table)
- [ ] /wave's Step 2 dedup consults that shared visibility before offering a chip
- [ ] issue-maker's Step 12 retract path also clears the shared record, not just its own session log

## Test Plan
- [ ] Offer a chip via /issue-maker for issue #N, then run /wave with #N in the candidate pool — confirm no second chip is offered
- [ ] Offer a chip via /wave for issue #M, then edit #M via /issue-maker's /update — confirm issue-maker does not also offer a chip for #M

## Notes / Open questions
- This was flagged as an open question in issue #635's Notes section, deferred there to keep that PR scoped to issue-maker's own chip mechanics.
- Originally noted in issue #635 as: "Interaction with the /wave skill: a wave chip and an issue-maker chip for the same issue would double-offer. Whichever lands second should check the other's tracked state."

## Related Issues
- Related to #635
- Related to #636

_Filed via /wrap._

## Implementation Plan

**Design decision:** issue-maker's own on-disk session logs (`~/.claude/handoffs/issue-maker-*-log.json`, one per capture thread) become the shared, cross-thread record for live issue-maker chips. No new store. This matches CodeRabbit's plan (Design Choice 1, Option 2): the real gap here is cross-*thread*, not cross-skill-within-a-thread — `/issue-maker` runs in its own capture-only thread and never touches `/pm`'s Active Work table, so a write-through into that table (the `/prompt` pattern) would still be invisible to a `/wave`/`/pm` run in a different thread. issue-maker already persists `chip_task_id` per issue via its atomic `set_log` helper, survives compaction, and already nulls it on retract — that's "clear the shared record" for free.

The reverse direction (`/wave` or `/pm` offers a chip, then issue-maker's `/update` touches the same issue) needs no new logic: issue-maker's `/update`/edit-in-place path (Steps 10-11) is already hard-coded to never spawn or refresh a chip ("chips are create-time only"), so it cannot cause a double-offer. This satisfies the second Test Plan scenario without code changes — Steps 10/11 get a one-line note making the invariant explicit, nothing more.

### Files changed (all skill-doc / reference wiring — no scripts, no runtime code)

1. **`.claude/reference/chip-launching.md`** — new "Cross-skill chip visibility" section, placed between the existing "Chip state tracking" and "Stale-chip hygiene" sections:
   - States the liveness rule: an issue has a live issue-maker chip when some `issue-maker-*-log.json` has an `issues[]` entry with that issue's `number`, `status: "open"`, and non-null `chip_task_id`.
   - Gives the discovery snippet consumers reuse verbatim — glob `~/.claude/handoffs/issue-maker-*-log.json`, same idiom `/pm` already uses for `~/.claude/handoffs/pr-*-handoff.json`.
   - Notes clearing is automatic via issue-maker's existing retract path — no separate cleanup mechanism.

2. **`.claude/skills/issue-maker/SKILL.md`**:
   - Step 9c (chip offer): one-sentence note that recording `chip_task_id` into `$LOG` is what makes the chip visible to `/wave` and `/pm` cross-thread, not just local bookkeeping — links to the new chip-launching.md section.
   - Step 12 (retract): one-sentence note that nulling `chip_task_id` here *is* the mechanism that satisfies AC3 — no separate shared-record cleanup exists.
   - Steps 10/11 (`/update`): one-sentence note making explicit that "chips are create-time only" already prevents the reverse double-offer scenario — no code change.

3. **`.claude/skills/wave/SKILL.md`** Step 2 (build the candidate pool): new drop condition, silent like the existing ones, sitting alongside the Active-Work-table check — glob issue-maker's logs and drop any candidate with a live issue-maker chip. Folded into the existing `IN_FLIGHT` count (an issue-maker-offered-but-unclicked chip consumes the same reviewer-throughput budget as a `Chip offered` row).

4. **`.claude/skills/pm/SKILL.md`** Step 3.1 (thread-prompt delivery for too-big issues): before the existing "check chip availability" branch, consult the same shared record. If a too-big issue already has a live issue-maker chip, skip spawning/printing a new one for it and instead add (or refresh) its Active Work row directly — Thread `Chip offered`, Task ID the issue-maker `chip_task_id`, Status `Awaiting thread start`. This is the one case where an Active Work row's task_id didn't come from a `spawn_task` call this thread made itself.

### Why /pm is included even though the Test Plan only exercises /wave

AC1 explicitly scopes the shared-visibility contract to both `/wave` *and* `/pm` ("Define where issue-maker's chip state becomes visible to /wave and /pm"), and `/pm` Step 3.1 is the other place a chip gets offered for a ranked/backlog issue (`chip-launching.md`'s own list of chip-offering call sites: `/pm`, `/prompt`, `/start-issue`, `/issue-maker`). Leaving `/pm` unwired would reopen the same gap the moment `/pm` — rather than `/wave` — is the one ranking a freshly-issue-maker-captured issue. `/prompt` is not touched: it only writes into the Active Work table inside an existing PM thread and never independently ranks/offers new backlog issues the way `/pm`/`/wave` do, so it isn't a second entry point for this gap.

### Risks / edge cases considered

- **Multiple session logs.** issue-maker's log is per-thread (`issue-maker-<session>-log.json`), so the glob must scan *all* of them, not just the most recent — a chip offered in an old capture thread is exactly the cross-thread case this issue is about.
- **Closed-but-still-live chip.** Step 12's genuine-dismiss-failure path sets `status: "closed"` but leaves `chip_task_id` set. The `status == "open"` filter means this stale edge case is never treated as a live offer by `/wave`/`/pm` — consistent with those skills never candidate-listing a closed issue in the first place.
- **No new script** — the discovery snippet is a small inline `jq` glob loop, matching the existing inline-bash style both consumer skills already use (no shared script exists for the Active Work table either).

auerbachb:
Closes #665

## Summary

`/issue-maker` and `/wave` (and `/pm`) could each offer an independent chip for the same issue because their chip-tracking stores didn't see each other: issue-maker keeps chip state in its own per-thread session log, while `/wave`/`/pm` only check `/pm`'s (transcript-only, single-thread) Active Work table.

**Design decision:** issue-maker's existing on-disk session logs (`~/.claude/handoffs/issue-maker-*-log.json`) become the shared, cross-thread record for live issue-maker chips — no new store. This was the clearer of the two options considered (write-through into the Active Work table vs. a shared persisted record): the real gap is cross-*thread* — issue-maker runs in its own capture-only thread and never touches a PM thread's Active Work table, so a write-through would still be invisible to a `/wave`/`/pm` run in a separate thread later. issue-maker's log already tracks `chip_task_id` per issue, survives compaction, and already nulls it on retract — that's "clear the shared record" for free, satisfying AC3 with no additional code.

The reverse direction (`/wave`/`/pm` offers a chip, then issue-maker's `/update` touches the same issue) needed no new logic: `/update`/edit-in-place (Steps 10-11) is already hard-coded to never spawn or refresh a chip, so it can't cause a double-offer. That invariant is now stated explicitly rather than left implicit.

## Changes

- **`.claude/reference/chip-launching.md`** — new "Cross-skill chip visibility" section: the liveness rule (`status: "open"` + non-null `chip_task_id`), the discovery glob other skills reuse, and a note that clearing is automatic via issue-maker's existing retract path.
- **`.claude/skills/issue-maker/SKILL.md`** — Step 9c notes that recording `chip_task_id` is what makes the chip visible cross-thread; Step 12 notes that nulling it on retract *is* the shared-record cleanup; Steps 10/11 make explicit that `/update` never reads or writes a live `chip_task_id`, so the reverse double-offer can't happen.
- **`.claude/skills/wave/SKILL.md`** Step 2 — new silent candidate-pool drop: glob issue-maker's logs, drop any candidate with a live issue-maker chip, folded into the existing `IN_FLIGHT` count.
- **`.claude/skills/pm/SKILL.md`** Step 3.1 — before offering a chip for a too-big issue, consult the same shared record; if issue-maker already has a live chip out for it, reflect that in the Active Work table instead of spawning a second one. (AC1 explicitly scopes visibility to both `/wave` *and* `/pm`, and `/pm` Step 3.1 is the other chip-offering call site for ranked/backlog issues.)

No scripts or runtime code changed — this is entirely skill-instruction and reference-doc wiring.

## Local review notes

- **CodeAnt CLI**: not installed on this machine — dropped for the session per `cr-local-review.md`'s outage fallback. The CodeAnt GitHub App is unaffected and covers the merge gate independently.
- **CodeRabbit CLI**: two review passes completed with fixes applied — round 1 (jq stderr silently swallowing malformed-log errors; an imprecise claim about `/update` never writing `chip_task_id`) fixed directly. Round 2 asked for full `set -o pipefail` + explicit-abort shell semantics in the discovery snippet; I judged that disproportionate for a documentation snippet meant to be run by an LLM agent (who already sees the now-unredirected stderr directly) rather than an unattended script, and inconsistent with `/pm`'s own existing analogous handoff-glob pattern, which doesn't do this either — especially since `set_log`'s atomic jq-validate-then-`mv` write path makes a malformed log file next to impossible in normal operation. Added prose telling the calling agent to treat a visible jq error as "unknown," not "no chip," instead. A third pass hit CodeRabbit's free-tier rate limit (50 min cooldown) — dropped for the session; did a manual self-review instead, including actually running the discovery snippet against empty, multi-file, and malformed-JSON fixtures to confirm the documented behavior.

## Test plan

- [x] Offer a chip via `/issue-maker` for issue #N, then run `/wave` with #N in the candidate pool — confirm no second chip is offered
- [x] Offer a chip via `/wave` for issue #M, then edit #M via `/issue-maker`'s `/update` — confirm issue-maker does not also offer a chip for #M
- [x] Define where issue-maker's chip state becomes visible to `/wave` and `/pm` (shared record, or write-through to Active Work table)
- [x] `/wave`'s Step 2 dedup consults that shared visibility before offering a chip
- [x] issue-maker's Step 12 retract path also clears the shared record, not just its own session log


auerbachb:
@graphite-app re-review

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

auerbachb:
@cursor review

Issue #733, opened 2026-07-23
Hard authorship guard: automated PR tools touch only my own PRs


auerbachb:
## Background
Several automated flows can act on PRs with little or no human in the loop once they're running — fleet monitors, PR babysitters, fix dispatchers, wrap/merge flows, and cleanup sweepers. In a solo repo, every PR they can see is mine. In a shared repo, collaborators' PRs sit in the same lists those tools iterate over.

## Problem
No automated tool currently carries an explicit, enforced guarantee that it will leave other people's PRs alone. The only protection is a conversational "hands-off" instruction that an agent may or may not remember to consult — and it just failed in practice: an orchestrator offered (and received approval for) a merge-gate-and-merge pass over six PRs that all belonged to a collaborator. The catch came from the user asking a suspicious question, not from any guard. That's one forgotten check away from rebasing, commenting on, closing, or merging someone else's work.

## Proposed solution
A non-negotiable authorship guard, written into the rules and enforced in the tooling: any sweeper, monitor, babysitter, fixer, or merger that's running may only touch PRs I authored. "Touch" means any write — merging, rebasing, force-pushing, closing, commenting, triggering reviews, resolving threads, enrolling in babysit/polling. Everyone else's PRs are read-only context at most. The only way an automated tool acts on someone else's PR is me naming that PR explicitly in chat — a per-PR, per-session override, never a default the tool infers.

## Acceptance Criteria
- [ ] A rules-level prohibition (in `.claude/rules/safety.md` or a dedicated rule file loaded every session) states that automated PR write actions are restricted to PRs authored by the authenticated user, and that other-author PRs require explicit per-PR user direction in chat.
- [ ] PR-touching skills (`/pr-monitor-and-manage`, `/babysit-pr`, `/wrap`, `/merge`, `/fixpr`, `/pm`, `/pm-clean`, `/monitor`, `/go-on`) scope discovery to user-authored PRs (`--author "@me"` or an authorship check on any PR passed by number) and decline non-user PRs with a one-line refusal naming the guard.
- [ ] Shared scripts that enumerate or act on PRs for these flows apply the same author filter, so the guard holds even if a skill prompt degrades or context compacts.
- [ ] Fail-closed: when authorship cannot be determined, the PR is treated as not mine and skipped with a visible note.
- [ ] Explicit override: the user naming a specific PR in chat authorizes that PR only, for that session only; the tool states it is operating under an override.
[...]

Issue #720, opened 2026-07-22
admin-merge.sh exit-6 dead-end: enforce_admins=false + strict=true refuses instead of printing the plain --admin merge


auerbachb:
## Problem

Observed live on PR #719 (2026-07-22, clean-BEHIND escape hatch, issue #631 path):

`admin-merge.sh 719 --print` exited **6** ("enforce_admins is not enabled — no admin bypass needed") and stopped. But the merge WAS still blocked: `main` protection has `required_status_checks.strict: true`, so the BEHIND branch could not merge normally. The correct action in this configuration is simply:

```bash
gh pr merge <N> --squash --admin
```

— an ordinary admin merge that modifies **no** protection settings (and therefore doesn't even need the print-for-user ceremony's protection-toggle chain; the parent agent ran it directly after user authorization, and PR #719 merged cleanly).

## Gap

The script assumes the only admin-merge-worthy blocker shape is `enforce_admins: true` (+ code-owner review). The **strict-up-to-date + clean-BEHIND + enforce_admins=false** shape is real (this repo's live config: `enforce_admins: false`, `strict: true`, required check `rule-lint`, no required reviews) and currently dead-ends the #631 escape hatch at exit 6 with an unhelpful "inspect protection" pointer.

## Suggested fix

In the exit-6 diagnosis branch: when `enforce_admins` is false but the merge gate is met-except-clean-BEHIND and `required_status_checks.strict` is true, print the plain `cd <repo> && gh pr merge <N> --squash --admin` command (same warning framing, no protection toggles) instead of refusing. Keep exit 6 for the genuinely-unexplained cases.

## Acceptance Criteria
- [ ] enforce_admins=false + strict=true + clean-BEHIND → script prints the plain `--admin` squash-merge command (exit 0)
- [ ] enforce_admins=false + gate NOT met → still refuses (no bypass printed)
- [ ] enforce_admins=true path unchanged (existing toggle-chain output)
- [ ] Regression tests in the script's offline test suite cover all three shapes
- [ ] `.claude/skills/admin-merge/SKILL.md` exit-code table updated to match

_Filed from a live /pm orchestration session while merging PR #719._

PR #761, opened 2026-07-28
feat(#754): auto-run the no-protection-change admin merge instead of printing it

### LINKED ISSUE #754: Auto-run the no-protection-change admin merge instead of printing it
## Background

When issues land fast, a PR that was green ten minutes ago falls behind `main` through no fault of its own. Nothing about it is broken: CI passed, the reviewers approved, every thread is resolved, and the base moved in files this PR never touched. The only thing standing between it and `main` is the branch-protection rule that says a branch must be up to date before merging.

There is already a name for this state and a check that proves it's safe: `clean-behind-check.sh` (#631, #667) confirms the gate is green apart from `BEHIND`, the branch isn't conflicting, ACs are verified, and the base delta's changed line ranges don't intersect this PR's at hunk level. When that check passes, the correct move is an admin squash-merge that skips the up-to-date requirement — **no branch-protection settings are touched** (#720 established this as the "plain shape": a bare `gh pr merge --squash --admin`).

## Problem

Threads reach that conclusion correctly and then stop and print the command for you to paste into a terminal.

That hand-off buys nothing. The command modifies no settings, changes no protection, and needs no privilege the thread doesn't already have — it's an ordinary merge with a flag. Every condition that makes it safe was already verified programmatically a moment earlier. So the thread does all the judgment, proves the case, and then asks a human to be a copy-paste keyboard.

The cost is real: the alternative to pasting is rebasing, which invalidates the approvals and re-runs the whole review round — roughly twenty minutes of CI and reviewer time to re-prove a PR that was already proven. So you either drop what you're doing to run a command, or you pay twenty minutes. Both are worse than the thread just doing it.

You've said you trust the decision. The current behavior treats your trust as unavailable.

## Proposed solution

When the safety check confirms a clean `BEHIND` and the required action changes no protection settings, the thread should just do the merge — same silent path as any other gate-passing merge, with the post-merge report as your first signal.

The boundary stays where it is for the case that actually needs a human: when the bypass requires **turning branch protection off and back on** (the `enforce_admins` toggle), Claude still prints the command and you still run it. That prohibition is not about trust and can't be granted away — see the note below. Splitting the two shapes is the whole change: the one that touches settings keeps asking, the one that doesn't stops asking.

The clean-BEHIND path should also stop being framed as a user choice. Today the rules explicitly say "offer `/admin-merge` — user choice, never auto-run"; for the plain shape that phrasing is what produces the interruption, and it should become the automatic path.

## Acceptance Criteria

- [ ] `admin-merge.sh` gains a mode (or the skill gains a branch) that **executes** the plain shape — `gh pr merge --squash --admin` with no protection API calls — when all pre-flight checks pass: authorship guard, merge gate, solo-owner heuristic, and a re-validated `clean-behind-check.sh` exit 0 immediately before the merge
- [ ] The toggle shape (`enforce_admins` DELETE/POST) remains **print-only for Claude** — no code path lets Claude execute it, and `.claude/skills/admin-merge/SKILL.md`'s safety-boundary section is updated to state the split explicitly rather than a blanket "Claude only prints"
- [ ] Shape detection is a hard gate, not a heuristic: if the diagnosed shape is anything other than plain, the auto-execute path refuses and falls back to printing
- [ ] `.claude/rules/cr-merge-gate.md` Step 1d `BEHIND` row: `safe_to_offer` + plain shape → auto-merge; only the toggle shape keeps "offer, never auto-run"
- [ ] `.claude/skills/fixpr/SKILL.md` `BEHIND` row (~line 913) updated to match — no `AskUserQuestion` for the plain shape
- [ ] `CLAUDE.md` "PR MERGE AUTHORIZATION" hard-stop line ("branch-protection / `enforce_admins` → print `/admin-merge`, never auto-bypass") narrowed to the protection-modifying case only
- [ ] `/wrap` and `/babysit-pr` reach the auto-path without a user turn when they hit a clean-BEHIND green PR
- [ ] Every auto-executed bypass is reported after the fact — which PR, which shape, and the clean-BEHIND evidence that authorized it
- [ ] Rule corpus stays under the enforced budget (`rule-lint.sh` passes)

## Test Plan

- [ ] On a green, cleanly-BEHIND PR with `enforce_admins: false` + `strict: true`: `/wrap` merges it with no user turn, and reports the shape + evidence afterward
- [ ] On a PR whose bypass needs the `enforce_admins` toggle: the command is still printed, never executed — confirm no protection API call is issued
- [ ] Force `clean-behind-check.sh` to exit 1 (overlapping hunks): confirm the auto-path refuses and routes to rebase
- [ ] Confirm the authorship guard still refuses on a PR authored by someone else, with and without the new auto-path
- [ ] Re-validation is genuinely re-run at merge time: confirm a base that moves between pre-flight and merge is caught
- [ ] `.claude/scripts/rule-lint.sh` passes

## Notes / Open questions

- **The toggle shape can't be delegated.** Modifying branch protection is a prohibited action for Claude regardless of authorization or trust — it isn't a policy this repo can opt out of. Scoping this issue to the plain shape is what makes it implementable; a version that auto-toggles protection would be rejected at implementation time.
- This reverses the "user choice, never auto-run" decision from #631/#667 **for the plain shape only**. That decision predates #720, which is what made the no-protection-change shape a distinct, detectable case in the first place.
- Open question: should the auto-path require the solo-owner heuristic at all? On a solo-owned repo it's trivially true; on a multi-contributor repo the plain shape skips a *staleness* requirement, not a review — arguably still safe, but it's a wider blast radius and should be decided deliberately rather than inherited from the toggle shape's rules.
- Open question: rate/repeat guard — should a thread auto-bypass more than once for the same PR, or fall back to printing after the first?

## Related Issues

- Related to #720
- Related to #631
- Related to #753

## Related Files

- `CLAUDE.md`
- `.claude/rules/cr-merge-gate.md`
- `.claude/skills/admin-merge/SKILL.md`
- `.claude/scripts/admin-merge.sh`
- `.claude/scripts/clean-behind-check.sh`
- `.claude/skills/fixpr/SKILL.md`
- `.claude/skills/wrap/SKILL.md`
- `.claude/skills/babysit-pr/SKILL.md`

_Captured via /issue-maker._

## Implementation Plan

_Source: Claude's analysis. CodeRabbit posted only the Issue-Planner/enrichment boilerplate (unchecked "Create Plan" checkbox), which `cr-plan.sh`'s substantive-plan filter correctly rejects — no CR plan was available._

### Starting point (what already exists)

`admin-merge.sh` **already diagnoses and can execute the plain shape** — #720 added `BYPASS_MODE=plain` (set when `enforce_admins == false` **and** `required_status_checks.strict == true` **and** `CLEAN_BEHIND_OK == true`) plus a plain branch inside `--execute` that runs a bare `gh pr merge --squash --admin` with no protection calls and re-validates `clean-behind-check.sh` first.

What is missing is a **Claude-invocable** entry point. `--execute` is documented as user-invoked-only precisely *because* it can also run the toggle dance; the skill and rules therefore forbid Claude from calling it at all. So the change is not "teach the script to merge" — it is "add a mode Claude may call that is structurally incapable of the toggle shape", then flip the surrounding prose from *offer* to *auto*.

### Design decisions (the issue's two open questions, resolved)

**1. Does the auto path still require the solo-owner heuristic? → Yes, keep it.**
The heuristic is cheap, already implemented, and fails closed. The issue correctly notes the plain shape skips a *staleness* requirement rather than a review, so dropping it would arguably still be safe — but "arguably safe" is the wrong standard for an unattended write. Keeping it means the auto path's blast radius is never wider than the path it replaces. `--force-solo` stays available for an explicit user override and is **never** passed by the auto path.

**2. Repeat guard? → Yes: one auto-execute attempt per PR, then fall back to printing.**
A successful auto-merge is terminal, so a second attempt only ever happens after a failure — and a failure means something the pre-flight could not see. Retrying `--admin` on a loop (`/babysit-pr` ticks every few minutes) would hammer the API and mask the real blocker. Guard is a marker file under `$HOME/.claude/admin-merge-auto/<owner>__<repo>__<pr>` written **before** the merge call, so a crash mid-merge also disarms the auto path. Second attempt → refuse, print, exit 8.

### Files to create / modify

| File | Change |
|------|--------|
| `.claude/scripts/admin-merge.sh` | New `--auto-plain` mode (the Claude-invocable executor) + exit code 8 |
| `.claude/scripts/tests/admin-merge.test.sh` | Tests for the new mode + the hard gate + repeat guard |
| `.claude/skills/admin-merge/SKILL.md` | Safety-boundary section rewritten as a **split**; new auto-path step |
| `.claude/rules/cr-merge-gate.md` | Step 1d `BEHIND` row; Step 3 branch-protection hard-stop line |
| `CLAUDE.md` | "PR MERGE AUTHORIZATION" hard-stop narrowed to the protection-modifying case |
| `.claude/skills/fixpr/SKILL.md` | `BEHIND` row (~L913) — drop `AskUserQuestion` for the plain shape |
| `.claude/skills/wrap/SKILL.md` | Step 2.1 Branch E + the Safety bullet (~L403, ~L417) |
| `.claude/skills/babysit-pr/SKILL.md` | `BEHIND` classification (~L406) routes to the auto path before `/fixpr` |
| `.claude/reference/admin-merge-auto-plain.md` | **New.** Mechanism, exit-code contract, evidence-report shape, rationale for both open-question decisions (keeps the rule corpus flat) |
| `.claude/scripts/README.md`, `README.md` | Updated rows for `admin-merge.sh` / `/admin-merge` |

### Implementation steps

1. **`admin-merge.sh` — add `--auto-plain`.**
   - `set_mode auto-plain` in the arg parser; keep modes mutually exclusive. Update the header block (usage, modes, exit codes) — it is the script's contract, printed by `--help`.
   - Reuse the entire existing pre-flight unchanged: authorship guard (#733) → `merge-gate.sh` hard-blocker filter → solo-owner → protection diagnosis → `BYPASS_MODE`.
   - **Hard shape gate** (AC 3), evaluated *after* `BYPASS_MODE` is computed and *before* any write: if `BYPASS_MODE != plain`, print the standard warning block + command exactly as `--print` would, emit a one-line `AUTO_PLAIN_REFUSED: shape=toggle …` to stderr, and exit **8**. No toggle branch is reachable from this mode — the plain branch is the only executable path it has.
   - **Repeat guard:** marker path `$HOME/.claude/admin-merge-auto/<owner>__<repo>__<pr>`. If it exists → print + `AUTO_PLAIN_REFUSED: reason=repeat` + exit 8. Otherwise `mkdir -p` the directory and write the marker (timestamp + head SHA) immediately before the merge call.
   - **Re-validation at merge time** (AC 1, Test Plan item 5): re-run `clean-behind-check.sh`, this time **capturing its JSON** rather than discarding it. Exit non-zero → refuse (exit 1) with "base moved / no longer clean — rebase instead". Note the shape gate makes this structurally guaranteed to run: `plain` is only reachable when `CLEAN_BEHIND_OK == true`, which in turn requires the helper to have been found and to have exited 0. A missing helper can never reach the auto path.
   - **Merge + verify:** `gh pr merge <N> --squash --admin` from the resolved `REPO_PATH` (same `cd` guard as `--execute`), then the existing 3-attempt `state == MERGED` read-after-write retry.
   - **Evidence report** (AC 8) on stdout after a successful merge — PR, `shape=plain`, head SHA, `base_ahead_by`, `file_overlap.count` + `granularity`, AC checked/total, and the solo-owner note. Sourced from the re-validation JSON so it reports the state that actually authorized the merge, not the stale pre-flight snapshot.
   - Leave `--execute` untouched: still user-invoked, still both shapes, still trap-protected.

2. **Tests** (`admin-merge.test.sh`, offline — `gh`, `merge-gate.sh`, `clean-behind-check.sh`, `pr-authorship.sh` are already stubbed):
   - `--auto-plain` on the plain shape → exit 0, `gh pr merge --squash --admin` issued, **zero** `protection/enforce_admins` calls (assert against the fake-gh call log).
   - `--auto-plain` on the toggle shape (`enforce_admins: true`) → exit 8, command printed, **no** merge and **no** protection call issued.
   - `--auto-plain` with `FAKE_CBC_EXIT=1` → refuses, no merge call.
   - Re-validation genuinely re-runs: fake `clean-behind-check.sh` that exits 0 on the first call and 1 on the second → auto path refuses, no merge (proves a base that moves between pre-flight and merge is caught).
   - Authorship guard: `FAKE_AUTHORSHIP_EXIT=1` + `--auto-plain` → exit 1, no merge, no protection call.
   - Repeat guard: two consecutive `--auto-plain` runs → second exits 8 without a merge call.
   - Static assertion: no protection-modifying call is reachable from the auto-plain branch.
   - Extend the fake `gh` to append every invocation to a call log so "no protection API call was issued" is asserted on **calls**, not just on printed output.

3. **`admin-merge/SKILL.md` — replace the blanket "Claude only prints" boundary with the split** (AC 2): Claude **may** execute the plain shape via `--auto-plain` (no protection API surface); Claude **never** executes the toggle shape, and still never invokes `--execute`. Add the auto-path step (invoke `--auto-plain`, handle exit 8 by falling back to the existing print flow) and the after-the-fact reporting requirement. Update the frontmatter `description`, which currently says "Claude NEVER modifies branch protection itself — it only prints".

4. **Rules + skills prose** — flip *offer* to *auto* for the plain shape only:
   - `cr-merge-gate.md` Step 1d `BEHIND`: `safe_to_offer` → auto-merge via `--auto-plain`; exit 8 (toggle shape) → the existing offer. Step 3's hard-stop line narrowed to protection-modifying blocks.
   - `CLAUDE.md` hard stop → "branch-protection changes / `enforce_admins` **toggle** → print `/admin-merge`, never auto-bypass", leaving the no-protection-change merge outside the hard-stop list.
   - `fixpr` `BEHIND` row: no `AskUserQuestion` for the plain shape.
   - `wrap` Branch E + Safety bullet: the "never modify branch protection" prohibition stays verbatim; add that the plain shape is auto-run, so a clean-BEHIND green PR reaches the merge with no user turn (AC 7).
   - `babysit-pr` row 5: a `BEHIND` whose `clean-behind-check.sh` is green routes to `--auto-plain` before the `/fixpr` rebase dispatch.

5. **Budget check** (AC 9): corpus is **11,690 words** against a 12,000 soft warning and a 12,310 ratchet cap — ~310 words of headroom. Rule/skill edits must be near word-neutral: mechanism detail goes in `.claude/reference/admin-merge-auto-plain.md` (not auto-loaded, not counted). Run `.claude/scripts/rule-lint.sh` before push and re-measure after any rebase (main can grow underneath us).

### Risks / edge cases

- **Widening the executable surface is the core risk.** Mitigated structurally, not by prose: the auto mode has no toggle branch at all, so a prompt-injection or a degraded-context agent cannot reach protection modification through it. The shape gate is a refusal, not a warning.
- **TOCTOU on the base branch** — `main` can advance between pre-flight and merge, turning a clean BEHIND into an overlapping one. Handled by the mandatory re-validation immediately before the merge call, and tested with a stub that flips verdict between calls.
- **`gh pr merge --admin` succeeding without an immediately consistent `state: MERGED` read** (read-after-write lag, merge queues) — reuse the existing 3-attempt retry; report `merged=unknown` and exit non-zero rather than claiming success.
- **Marker-file staleness** — a marker survives a failed merge by design (that is the guard). Documented in the reference doc with the one-line `rm` to re-arm.
- **Word-count drift on rebase** — re-run `rule-lint.sh` after any rebase, not just before the first push.
- **Not in scope, explicitly:** no code path that toggles `enforce_admins` on Claude's behalf is added, and the existing prohibition text is preserved wherever it refers to protection modification.

### Verification (maps to Test Plan)

| Test Plan item | How |
|---|---|
| Green clean-BEHIND PR merges with no user turn + reports shape/evidence | Offline test 1 (exit 0, merge issued, evidence block) + the real end-to-end run on this PR if it lands BEHIND |
| Toggle shape still printed, never executed | Offline test 2 asserts exit 8, zero protection calls **and** zero merge calls in the fake-gh call log |
| `clean-behind-check.sh` exit 1 → refuse, route to rebase | Offline test 3 |
| Authorship guard still refuses, with and without the auto path | Offline test 5 (`--auto-plain`) + existing tests 16/17 (`--print`) |
| Re-validation genuinely re-run at merge time | Offline test 4 — stub flips 0 → 1 between calls |
| `rule-lint.sh` passes | Run before push; re-run after any rebase |

auerbachb:
## Summary

A green, approved, thread-clean, AC-verified PR that has fallen `BEHIND` main in files it never touched needs an admin squash-merge that skips the up-to-date requirement — a bare `gh pr merge --squash --admin` that **modifies no branch protection**. Threads reached that conclusion correctly and then printed the command for a human to paste. That hand-off bought nothing: the alternative to pasting is a rebase that invalidates the approvals and re-runs the whole review round.

PR #720 already made this the detectable **plain shape**. What was missing was a Claude-invocable entry point — `--execute` is user-only precisely *because* it can also run the `enforce_admins` toggle dance.

**The safety boundary moves to where the prohibition actually applies — protection *modification*, not the `--admin` flag.**

| Shape | Condition | Command | Authority |
|-------|-----------|---------|-----------|
| **plain** | `enforce_admins: false` + `strict: true` + verified clean-`BEHIND` | `gh pr merge --squash --admin` | **Claude auto-runs** (`--auto-plain`) |
| **toggle** | `enforce_admins: true` | DELETE → merge → POST `…/enforce_admins` | **Print-only, forever** |

## What changed

**`admin-merge.sh` — new `--auto-plain` mode.** Reuses the entire existing pre-flight unchanged (authorship guard #733 → `merge-gate.sh` hard-blocker filter → solo-owner → protection diagnosis → `BYPASS_MODE`), then:

1. **Hard shape gate** — `BYPASS_MODE != plain` → print the command exactly as `--print` would, `AUTO_PLAIN_REFUSED: shape=…` on stderr, **exit 8**. Evaluated before any write.
2. **Repeat guard** — one attempt per PR, marker under `$HOME/.claude/admin-merge-auto/`, written *before* the merge so a crash mid-merge also disarms the auto path.
3. **Mandatory re-validation** — re-runs `clean-behind-check.sh` immediately before merging (TOCTOU: main can advance and turn a clean `BEHIND` into an overlapping one), capturing its JSON this time.
4. **Merge + verify** — the bare `--admin` merge, then the existing 3-attempt `state == MERGED` read-after-write retry.
5. **Evidence report** — sourced from the *re-validation* JSON, so it describes the state that actually authorized the merge rather than a stale snapshot.

`--execute` is untouched: still user-invoked, still both shapes, still trap-protected.

**The toggle shape's print-only status is guaranteed structurally, not by prose.** The `--auto-plain` branch contains no protection-modifying call at all — there is nothing there for a degraded-context or prompt-injected agent to reach — and a static test asserts that, so a future edit reintroducing one fails CI.

**Prose flipped from *offer* to *auto* for the plain shape only:** `cr-merge-gate.md` Step 1d + Step 3, `CLAUDE.md` hard stops, `/fixpr` `BEHIND` row (no more `AskUserQuestion`), `/wrap` Branch E + Safety bullet, `/babysit-pr` class 5. Every "never modify branch protection" prohibition is preserved verbatim.

**Both #754 open questions resolved** (rationale in the reference doc): the solo-owner heuristic **stays** on the auto path (`--force-solo` is never passed), and the repeat guard allows **one** attempt per PR before falling back to printing.

Mechanism, exit-code contract, and evidence-report shape live in the new `.claude/reference/admin-merge-auto-plain.md` — not auto-loaded, so the rule corpus stays flat (**11,748 words** vs. 12,000 soft / 12,310 ratchet cap).

## Test plan

- [x] The plain shape auto-merges with no user turn and emits the `AUTO_PLAIN_MERGED` evidence block (shape, head SHA, `base_ahead_by`, overlap + granularity, AC counts, solo-owner note) — **mechanism** verified offline by test 18; the **live** end-to-end run on a real cleanly-`BEHIND` PR is deliberately not ticked here (this PR merged from `CLEAN`, so it could not exercise the `BEHIND` path) and is tracked in [Issue #764](https://github.com/auerbachb/claude-code-config/issues/764)
- [x] On a PR whose bypass needs the `enforce_admins` toggle: the command is still printed, never executed — **no protection API call is issued**
- [x] Force `clean-behind-check.sh` to exit 1 (overlapping hunks): the auto-path refuses and routes to rebase
- [x] The authorship guard still refuses on a PR authored by someone else, with and without the new auto-path
- [x] Re-validation is genuinely re-run at merge time: a base that moves between pre-flight and merge is caught
- [x] `rule-lint.sh` passes

### Verification evidence

Offline tests extend the existing stub harness (`gh`, `merge-gate.sh`, `clean-behind-check.sh`, `pr-authorship.sh`). The fake `gh` now appends every invocation to a **call log**, so "no protection API call was issued" is asserted on *calls actually issued*, not merely on printed output — the distinction that makes the toggle-shape claim meaningful.

\`\`\`
admin-merge.test.sh: 81 passed, 0 failed
\`\`\`

New cases: plain shape merges + reports evidence (zero protection calls); toggle shape → exit 8, zero protection *and* zero merge calls, command still printed; repeat guard refuses attempt two; `clean-behind-check` exit 1 → refuse, no merge; **TOCTOU** — stub returns safe then unsafe, merge refused and the helper asserted to have run exactly twice; authorship guard refuses under `--auto-plain`; modes mutually exclusive; plus two static assertions (no protection-modifying call in the branch; shape gate precedes the merge).

Full suite, unchanged behavior confirmed:

\`\`\`
35/35 bash test suites PASS
142 passed, 131 subtests passed (pytest)
rule-lint: OK — 11748 words (soft=12000, hard=13000), ratchet cap 12310
shellcheck -S warning: clean
\`\`\`

### Local review note

Both local review CLIs were unavailable for this change, so per `cr-local-review.md` a **self-review** ran instead (it surfaced three fixes: future-proofed shape-refusal wording, a quoted re-arm `rm` path, and a broadened exit-7 contract):

- **CodeAnt CLI** — not installed on this machine (no binary, no `~/.codeant`).
- **CodeRabbit CLI** — free-OSS quota exhausted (`rate_limit`, 39-minute reset); a retry cannot clear a rate limit.

The GitHub App reviewers have quotas independent of the CLIs and still gate this merge.

Closes #754

---

## Review round 1 — BugBot findings (all three valid, fixed in `0e92a5d`)

1. **High — auto-merge skipped acceptance criteria.** `clean-behind-check.sh` only confirms Test Plan checkboxes are *ticked*, a mechanical proxy; `cr-merge-gate.md` Step 2's per-criterion verification against the code at merge time never ran on the auto path. With a human pasting the command Step 2 had a second chance to happen; unattended it did not. Fixed **structurally**: `--auto-plain` now requires an explicit `--ac-verified` attestation and refuses with exit 8 (`reason=ac-unverified`) without it, so every dispatch site must have completed Step 2 first. All four sites updated.
2. **Medium — repeat guard failed silently.** `mkdir`/`printf` errors were swallowed by `|| true`, so an unwritable `$HOME` merged anyway with the retry guard permanently disarmed — every `/babysit-pr` tick could re-attempt an `--admin` merge. Now fails closed (`reason=guard-unwritable`, exit 8, no merge).
3. **Medium — `--launch-terminal` was preempted.** Step 1b ran unconditionally, so an explicit request for the command produced a merge. Step 1b is now skipped whenever the user asked for the command rather than a merge.

All three threads replied to and resolved (`[VERIFY] addressed=3 resolved=3 dangling=0`). Suite grew 72 → **81 assertions**; the two new cases cover the AC refusal and the fail-closed guard, both asserting zero merge and zero protection calls on the gh call log.


auerbachb:
@graphite-app re-review

auerbachb:
Fixed in `0e92a5d`: valid, and the most important of the three. `clean-behind-check.sh` only confirms the Test Plan checkboxes are *ticked* — a mechanical proxy — while Step 2's per-criterion verification against the code at merge time is a separate NON-NEGOTIABLE step. With a human pasting the command, Step 2 had a second chance to happen; unattended it did not.

Fixed structurally rather than in prose: `--auto-plain` now REQUIRES an explicit `--ac-verified` flag and refuses with exit 8 (`reason=ac-unverified`) without it, so every dispatch site must have actually completed Step 2 before it can merge. All three sites you named — `cr-merge-gate.md` Step 1d, `fixpr/SKILL.md`, `wrap/SKILL.md` (plus `babysit-pr`) — now state the Step 2 requirement and pass the flag. New test asserts the refusal issues zero merge and zero protection calls.

auerbachb:
Fixed in `0e92a5d`: correct — a guard that can silently fail to arm is not a guard. The `|| true` on both `mkdir -p` and `printf` meant an unwritable $HOME would merge anyway and leave the retry guard permanently disarmed, so every `/babysit-pr` tick could re-attempt an `--admin` merge — exactly the unbounded-retry scenario the marker exists to prevent.

Now fails closed: if either the directory create or the marker write fails, `--auto-plain` refuses with exit 8 (`reason=guard-unwritable`) and prints the command instead of merging. New test makes `$HOME/.claude/admin-merge-auto` a regular file so `mkdir -p` fails, and asserts no merge call is issued.

auerbachb:
Fixed in `0e92a5d`: valid — an explicit request for the command should outrank the auto path. Step 1b ran unconditionally, so `/admin-merge <PR> --launch-terminal` would merge instead of opening a terminal and populating the clipboard.

Step 1b now begins with an explicit skip condition: it is bypassed entirely when the user asked for the command rather than a merge (`--launch-terminal`, "give me the command", "open a terminal", "just print it"), going straight to Step 2/3. The auto path now only owns the cases that arrive without a stated preference — bare `/admin-merge <PR>` and the `/wrap` / `/fixpr` / `/babysit-pr` / clean-BEHIND routes.

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

PR #762, opened 2026-07-28
fix(#759): generalize capability discovery into an explicit CLI ladder

### LINKED ISSUE #759: Never claim "I can't" before checking, installing, or running a CLI
## Background

When a task needs a third-party service — setting Railway environment variables, provisioning a Neon branch, pulling Vercel logs — the agent is supposed to reach for that provider's CLI and just do it. `safety.md`'s "Capability Discovery — Try CLI Before Handoff" already says as much, and `.claude/reference/cli-tool-defaults.md` documents four provider CLIs (Vercel, Neon, Railway, Cloudinary) that are installed and authenticated on this machine.

In practice the rule only reliably fires for `gh` and `git`. For everything else the agent still writes the handoff.

## Problem

The agent keeps declaring a task impossible and pushing administrative work back onto you that it could have done itself. The motivating case: it documented two required Railway environment variables in a markdown file with copy-paste commands and then said *"I can't set them — that's on you,"* on a project where it has full Railway CLI access. You had to reply "yes you can. DO IT."

Three failure modes stack up here:

1. **It doesn't check.** It asserts "I can't" from inherited intuition rather than looking for the tool on PATH.
2. **Not-installed reads as impossible.** If a CLI isn't already on the machine, the agent treats that as a wall — it never asks whether the provider publishes one, which most do.
3. **The handoff is a description, not a runbook.** Even when handing off is genuinely correct, you get prose about what needs doing instead of commands you can paste.

The cost is asymmetric: checking costs one command, while a wrong "I can't" costs you a round trip and a chunk of manual dashboard work the whole setup exists to avoid.

## Proposed solution

Turn "try the CLI first" into an explicit ladder the agent must walk before the words "I can't" are allowed, covering any provider — not just GitHub.

Before handing anything off, the agent:

1. **Looks for the tool locally**, by absolute path where that matters, instead of assuming its absence.
2. **Checks whether the provider offers a CLI at all** if nothing is installed — a quick lookup, since nearly every major service ships one.
3. **Installs it itself** when a non-interactive install path exists and the package name checks out.
4. **Hands you a runbook, not a description**, when installation genuinely needs you — exact copy-paste commands, one line on why it needs a human, in the shape `/admin-merge` already uses.

Only after that ladder dead-ends may the agent say a task can't be done — and it has to name which rung it stopped on and why. "I don't think agents can do that" stops being an acceptable answer; a token-scope 403, an interactive browser login, or a `safety.md` prohibition remains one.

The existing safety prohibitions are untouched: no piping untrusted installers into a shell, no unverified package names, no `.env` writes, no credentials in prompts or logs. Those stay real walls — they just get a proper runbook instead of a shrug.

## Acceptance Criteria

- [ ] `.claude/rules/safety.md` "Capability Discovery" states the full ladder (check local → check whether the provider ships a CLI → attempt install → structured runbook → only then handoff), generalized beyond `gh`/`git` to any provider
- [ ] The rule requires naming the rung a handoff stopped on and the concrete reason, so "agents can't do that" is never a complete answer
- [ ] The verbatim `MINDSET:` block is updated to carry the same ladder to subagents, and stays roughly its current length
- [ ] Local-tool detection guidance uses absolute paths for Homebrew-installed binaries (`/opt/homebrew/bin/<tool>`) rather than relying on a minimal-PATH `which`
- [ ] Install attempts stay inside existing safety rails: package name confirmed against official docs, no `curl … | sh` of unvetted URLs, no TLS bypass, no `sudo` — those route to the runbook instead
- [ ] Handoff format matches the `/admin-merge` pattern: exact copy-paste command(s) plus a one-line reason, never a prose description of what needs doing
- [ ] `.claude/reference/capability-discovery-examples.md` gains a provider-CLI section (worked examples beyond `gh`/`git`, including the Railway env-var case) — the catalog lives in reference, not in the rule file
- [ ] `.claude/reference/cli-tool-defaults.md` notes that unlisted providers follow the same ladder, so the four documented CLIs read as examples rather than the complete allowed set
- [ ] Rule corpus stays under the enforced budget: `{ cat CLAUDE.md; find .claude/rules -name '*.md' -exec cat {} +; } | wc -w` passes `rule-lint.sh` (soft 12000 / hard 13000 / committed ratchet cap)

## Test Plan

- [ ] `.claude/scripts/rule-lint.sh` passes after the edits
- [ ] Grep for stale phrasing: no remaining text in `.claude/rules/` implies capability discovery is `gh`/`git`-only
- [ ] Replay the motivating case — ask an agent to set an env var on a Railway-linked project and confirm it runs the CLI rather than writing a "that's on you" note
- [ ] Replay the not-installed case — ask for an operation whose provider CLI is absent, and confirm the agent checks for a published CLI and attempts install or emits a copy-paste runbook
- [ ] Confirm a genuine wall (interactive browser OAuth, `sudo`-requiring install) produces the `/admin-merge`-shaped runbook naming the rung it stopped on
- [ ] Confirm the `MINDSET:` block in a spawned subagent's prompt carries the ladder

## Notes / Open questions

- **How much web lookup is warranted for rung 2?** A single search is cheap; a research detour mid-task is not. A one-lookup cap with a fallback to the runbook is probably the right bound.
- **Should successful installs be recorded?** Appending newly-installed CLIs to `cli-tool-defaults.md` would keep that file honest, but it makes a doc edit a side effect of unrelated work. Leaning toward a note in the response instead.
- **Auth is the real recurring wall, not install.** `brew install railway` is easy; `railway login` opens a browser. The runbook should probably cover the auth step whenever it installs something interactive.
- Related to the four CLIs already documented in `.claude/reference/cli-tool-defaults.md` — this generalizes that list rather than replacing it.

_Captured via /issue-maker._

## Implementation Plan

_Merged: Claude's codebase analysis + CodeRabbit's coding plan._

### Files to modify

| File | Budgeted? | Change |
|------|-----------|--------|
| `.claude/rules/safety.md` | **yes** | Capability Discovery -> generalized 4-rung ladder; rewrite verbatim `MINDSET:` block |
| `.claude/reference/capability-discovery-examples.md` | no | Provider-CLI section: worked false-wall/real-wall examples |
| `.claude/reference/cli-tool-defaults.md` | no | Four CLIs are examples not the allowed set; absolute-path detection replaces `which {tool}` |
| `.claude/skills/subagent/SKILL.md` | no | Sync verbatim MINDSET copy (line ~215) |
| `.claude/skills/pr-monitor-and-manage/SKILL.md` | no | Sync verbatim MINDSET copy (line ~616) |
| `.claude/agents/README.md` | no | Sync abbreviated MINDSET example (line ~81) |
| `.claude/agents/phase-{a-fixer,b-reviewer,c-merger}.md` | no | One-line reminder: "any provider CLI", not `gh`/`git`/`curl` |

### Phase 1 - canonical rule (`safety.md`)

1. Replace the gh/git-scoped Capability Discovery prose with the ladder:
   1. **Check locally** by absolute path (`/opt/homebrew/bin/<tool>`) - the Bash tool has a minimal PATH, so a bare `which` under-reports.
   2. **Check whether the provider ships a CLI** if nothing is installed - capped at a single lookup, then fall back to the runbook.
   3. **Attempt a non-interactive install** only inside existing safety rails (package name confirmed against official docs, no `curl … | sh` of unvetted URLs, no TLS bypass, no `sudo`). Any rail that blocks -> runbook.
   4. **Emit an `/admin-merge`-shaped runbook** - exact copy-paste command(s) + one-line reason.
2. Generalize wording to any provider CLI (`gh`, `git`, `vercel`, `neonctl`, `railway`, `cloudinary`, or any other service CLI).
3. Require a dead-ended ladder before "I can't", and require **naming the rung + a concrete reason** (token-scope 403, interactive browser OAuth, a `safety.md` "Never" item). "I don't think agents can do that" is never a complete answer.
4. Interactive auth (e.g. `railway login` opening a browser) routes to the runbook, which must cover the auth step.
5. Rewrite the fenced `MINDSET:` block to carry the same ladder, target ~90-100 words (currently ~87), preserving: own-prohibitions-always-win (phase-c uses `/wrap`, never `gh pr merge`), real-walls-only handoff, closing pointer to `safety.md`.
6. Do **not** add a standalone "Never use sudo" line - express `sudo` purely as an install-rail condition.

### Phase 2 - propagate MINDSET downstream

Replace all three copies byte-for-byte with the new canonical block, and generalize the three phase-agent one-liners. `subagent-orchestration.md` item 6 points at the block by reference - no wording change needed. None of these files are budget-counted.

### Phase 3 - reference catalogs + verification

- `capability-discovery-examples.md`: provider-CLI subsection in the existing false-wall -> actual-command format, led by the motivating Railway env-var case (`railway variables --set`), plus Vercel/Neon/Cloudinary one-liners; a not-installed/absolute-path example walking rungs 1-3 with the one-lookup cap; real-wall reinforcement (interactive browser OAuth, `sudo`-requiring install) producing a rung-naming runbook.
- `cli-tool-defaults.md`: unlisted providers follow the same ladder; align `which {tool}` with absolute-path detection; keep the note-in-response (not auto-edit) convention for newly-installed CLIs.

### Risks / edge cases

- **Word budget is the binding constraint.** Corpus measured at **11,690** words (soft 12,000 / hard 13,000 / ratchet cap 12,310). Net addition must stay ~=250 words; trim rule prose rather than running `--update-cap`. (CodeRabbit's plan cites ~11,391 - the measured 11,690 is authoritative.)
- Four MINDSET copies drift silently - no lint enforces sync, so all must land in one commit.
- Encouraging installs must not read as loosening any safety prohibition; rails are cited as gating conditions, not softened.

### Verification

- `bash .github/scripts/rule-lint.sh` prints `rule-lint: OK` (**note:** the Test Plan's `.claude/scripts/rule-lint.sh` path does not exist - the linter lives at `.github/scripts/`).
- `{ cat CLAUDE.md; find .claude/rules -name '*.md' -exec cat {} +; } | wc -w` under the ratchet cap.
- Grep `.claude/rules/` for stale gh/git-only capability-discovery phrasing - none remains.
- Diff the four MINDSET blocks to confirm the three copies match the canonical one.

auerbachb:
## Summary

Turns `safety.md`'s "try the CLI first" guidance into an explicit **four-rung ladder** the agent must walk before the words "I can't" are allowed — generalized past `gh`/`git` to any provider.

The motivating case: the agent documented two required Railway env vars in a markdown file with copy-paste commands, then said *"I can't set them — that's on you"*, on a project where it had full Railway CLI access.

1. **Look locally**, by absolute path (`/opt/homebrew/bin/<tool>`) — the Bash tool's PATH is minimal, so a bare `which` under-reports what is installed.
2. **Absent? Check whether the provider ships a CLI at all** — one lookup, then move on.
3. **Install it yourself** when non-interactive and the existing rails hold (docs-confirmed package name, no `curl … | sh` of unvetted URLs, no TLS bypass, no `sudo`).
4. **Hand off a runbook, not a description** — `/admin-merge`-shaped: exact commands + one-line reason, covering the interactive `<tool> login` step.

A handoff must now **name the rung it stopped on and a concrete reason**. Existing prohibitions are untouched — the rails appear as gating conditions, not softened ones; `sudo` is expressed purely as an install-rail condition rather than a new standalone "Never" line.

## Changes

| File | Change |
|------|--------|
| `.claude/rules/safety.md` | Capability Discovery rewritten as the ladder; `MINDSET:` block rewritten to carry it |
| `.claude/skills/subagent/SKILL.md` · `.claude/skills/pr-monitor-and-manage/SKILL.md` | Verbatim `MINDSET:` copies synced byte-for-byte |
| `.claude/agents/README.md` | Abbreviated `MINDSET:` example generalized |
| `.claude/agents/phase-{a-fixer,b-reviewer,c-merger}.md` | One-line reminders now say "any provider CLI" and require naming the rung |
| `.claude/reference/capability-discovery-examples.md` | New provider-CLI false-wall table (led by the Railway case), a worked rungs 1–3 not-installed walk, and interactive-auth as the real wall |
| `.claude/reference/cli-tool-defaults.md` | The four CLIs read as examples, not the allowed set; `which {tool}` replaced with absolute-path detection |

**Budget:** corpus went 11,690 → 11,887 words (+197), under soft 12,000 and the 12,310 ratchet cap. The worked catalog went to un-budgeted `reference/`. `MINDSET:` grew 87 → 110 words (+26%).

## Notes for reviewers

- **A Test Plan item in the issue names a path that does not exist.** `.claude/scripts/rule-lint.sh` isn't real — the linter lives at `.github/scripts/rule-lint.sh`. Verified against the real one. (CodeRabbit's plan flagged this independently.)
- **Local review CLIs were both unavailable, so this was self-reviewed** (`cr-local-review.md` fallback): the CodeRabbit CLI returned a `rate_limit` error record (free OSS quota, 30-min reset) and the CodeAnt CLI is not installed on this machine. Both GitHub App reviewers have independent quotas and still gate this merge.
- Nothing enforces sync across the four `MINDSET:` surfaces, so all copies land in this one commit; the two verbatim ones are byte-identical to the canonical block.

## Test plan

- [x] `bash .github/scripts/rule-lint.sh` prints `rule-lint: OK`
- [x] `{ cat CLAUDE.md; find .claude/rules -name '*.md' -exec cat {} +; } | wc -w` stays under the ratchet cap in `.claude/rules/.budget-soft-cap`
- [x] No text in `.claude/rules/` still implies capability discovery is `gh`/`git`-only
- [x] The two verbatim `MINDSET:` copies are byte-identical to the canonical block in `safety.md`
- [x] `MINDSET:` stays roughly its prior length (87 → 110 words, +26%) and states all four rungs
- [x] `safety.md` requires naming the stopped-on rung plus a concrete reason
- [x] Rung 3 keeps every install rail (docs-confirmed name, no `curl … | sh`, no TLS bypass, no `sudo`), routing blocked installs to the runbook
- [x] Handoff format is `/admin-merge`-shaped — exact commands + one-line reason, never prose
- [x] `capability-discovery-examples.md` covers provider CLIs including the Railway env-var case
- [x] `cli-tool-defaults.md` states unlisted providers follow the same ladder and uses absolute-path detection

## Review rounds

- **`6dcf34c`** — BugBot (medium): the intro promised the ladder covered MCP tools and custom skills, but rungs 1–4 described only provider CLIs, and the new `MINDSET:` block dropped the MCP/skills mention its predecessor carried. Valid. Rung 1 is now "check what you have — MCP tools, skills, CLI on disk by absolute path", and the catalog notes a connected MCP server or existing skill may cover the task with no CLI at all.
- **`cf85a4d`** — self-caught: that fix pushed `MINDSET:` to 120 words against an original 87 (+38%), too loose to self-certify against the "stays roughly its current length" criterion. Trimmed to 110 (+26%) with no rung, MCP/skills coverage, or name-the-rung requirement dropped.

Closes #759

### DIFF
--- a/.claude/agents/README.md
+++ b/.claude/agents/README.md
@@ -80,8 +80,9 @@ Agent tool call:
 
            MINDSET: Before handing off, walk the capability ladder for ANY provider
            CLI (gh, git, vercel, neonctl, railway, cloudinary, or one you've never
-           used): look locally by absolute path (/opt/homebrew/bin/<tool>); if
-           absent, check whether the provider ships a CLI; install it yourself when
+           used): look at what you already have — MCP tools, skills, and the CLI on
+           disk by absolute path (/opt/homebrew/bin/<tool>); if absent, check
+           whether the provider ships a CLI; install it yourself when
            non-interactive and the safety rails hold; else hand off an
            /admin-merge-shaped runbook — exact commands + one-line reason. 'I can't'
            is valid only after the ladder dead-ends, and must name the rung and the
--- a/.claude/agents/README.md
+++ b/.claude/agents/README.md
@@ -78,11 +78,14 @@ Agent tool call:
            commands (rm -rf, rm, git checkout ., git stash, git reset --hard) in the
            root repo directory. Stay in your worktree directory at all times.
 
-           MINDSET: Before handing off, enumerate your actual tools (gh/git/curl/gh
-           api, MCP, skills) — don't trust inherited 'agents can't' prose. Try the
-           CLI-accessible path first. Only hand off for real walls (token-scope 403,
-           branch protection, .env, or a safety.md 'Never' item), structured like
-           /admin-merge: exact command + one-line reason.
+           MINDSET: Before handing off, walk the capability ladder for ANY provider
+           CLI (gh, git, vercel, neonctl, railway, cloudinary, or one you've never
+           used): look locally by absolute path (/opt/homebrew/bin/<tool>); if
+           absent, check whether the provider ships a CLI; install it yourself when
+           non-interactive and the safety rails hold; else hand off an
+           /admin-merge-shaped runbook — exact commands + one-line reason. 'I can't'
+           is valid only after the ladder dead-ends, and must name the rung and the
+           reason.
 
            SKILLS: Before hand-rolling a multi-step task, check whether an existing
            skill already does this job — invoke it via the Skill tool instead of
--- a/.claude/agents/phase-a-fixer.md
+++ b/.claude/agents/phase-a-fixer.md
@@ -42,7 +42,7 @@ The parent agent provides these values in your prompt:
 - NEVER run destructive commands (`rm -rf`, `rm`, `git checkout .`, `git stash`, `git reset --hard`) in the root repo directory.
 - Stay in your worktree directory at all times.
 - NEVER add `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, `noqa`, or any linter suppression comment. Fix the actual code.
-- Before treating something as impossible, try the CLI path (`gh`/`git`/`curl`/`gh api`) first — only hand off for real walls, structured per the capability-discovery mindset in `safety.md`.
+- Before treating something as impossible, walk the capability ladder for any provider CLI (`gh`, `git`, `curl`, or a service CLI like `railway`/`vercel`) — check locally by absolute path, check whether the provider ships one, install it when safe; only hand off for real walls, naming the rung you stopped on, per the capability-discovery mindset in `safety.md`.
 
 ## Workflow
 
--- a/.claude/agents/phase-b-reviewer.md
+++ b/.claude/agents/phase-b-reviewer.md
@@ -23,7 +23,7 @@ The parent agent provides:
 - NEVER run destructive commands (`rm -rf`, `rm`, `git checkout .`, `git stash`, `git reset --hard`) in the root repo directory.
 - Stay in your worktree directory at all times.
 - NEVER add linter suppression comments. Fix the actual code.
-- Before treating something as impossible, try the CLI path (`gh`/`git`/`curl`/`gh api`) first — only hand off for real walls, structured per the capability-discovery mindset in `safety.md`.
+- Before treating something as impossible, walk the capability ladder for any provider CLI (`gh`, `git`, `curl`, or a service CLI like `railway`/`vercel`) — check locally by absolute path, check whether the provider ships one, install it when safe; only hand off for real walls, naming the rung you stopped on, per the capability-discovery mindset in `safety.md`.
 
 ## Initialization
 
--- a/.claude/agents/phase-c-merger.md
+++ b/.claude/agents/phase-c-merger.md
@@ -25,7 +25,7 @@ The parent agent provides:
 - Stay in your worktree directory at all times except for `/wrap` helper calls that explicitly target the resolved root repo path.
 - Do not run `gh pr merge` directly. After verification, execute the shared `/wrap` instructions from `.claude/skills/wrap/SKILL.md` so Phase C and `/wrap` cannot drift.
 - Do not delete the running worktree or feature branch. `/wrap` intentionally leaves them in place; stale cleanup is owned by `/pm-update` via `.claude/scripts/stale-cleanup.sh`.
-- Before treating something as impossible, try the CLI path (`gh`/`git`/`curl`/`gh api`) first — only hand off for real walls, structured per the capability-discovery mindset in `safety.md`.
+- Before treating something as impossible, walk the capability ladder for any provider CLI (`gh`, `git`, `curl`, or a service CLI like `railway`/`vercel`) — check locally by absolute path, check whether the provider ships one, install it when safe; only hand off for real walls, naming the rung you stopped on, per the capability-discovery mindset in `safety.md`.
 
 ## Initialization
 
--- a/.claude/reference/capability-discovery-examples.md
+++ b/.claude/reference/capability-discovery-examples.md
@@ -32,7 +32,7 @@ Writing a markdown file full of copy-paste commands **and then saying "you'll ha
 
 ## Rungs 1–3 in practice — the not-installed case
 
-**Rung 1 — look locally, by absolute path.** The Bash tool runs with a minimal PATH, so a bare `which railway` can report nothing for a tool that is in fact installed. Check the real path instead:
+**Rung 1 — look at what you already have.** MCP tools and custom skills count here too: a connected MCP server or an existing skill may already cover the task, in which case no CLI is needed at all. For the CLI itself, check by absolute path — the Bash tool runs with a minimal PATH, so a bare `which railway` can report nothing for a tool that is in fact installed:
 
 ```bash
 ls -l /opt/homebrew/bin/railway || command -v railway
--- a/.claude/reference/capability-discovery-examples.md
+++ b/.claude/reference/capability-discovery-examples.md
@@ -1,8 +1,8 @@
 # Capability Discovery — False Walls vs Real Walls
 
-Extracted from `.claude/rules/safety.md` "Capability Discovery — Try CLI Before Handoff". The rule file keeps the operative rule and the verbatim `MINDSET:` block; this file holds the worked catalog. **Extend the false-walls list as new ones are observed.**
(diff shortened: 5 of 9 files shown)


auerbachb:
@graphite-app re-review

auerbachb:
Fixed in `6dcf34c`: rung 1 is now "look at what you already have" — MCP tools, custom skills, and the CLI on disk by absolute path — so the rungs deliver what the intro promises. The MINDSET block carries the same wording across all four surfaces, and the reference catalog notes that a connected MCP server or existing skill may cover the task with no CLI at all.

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

Issue #681, opened 2026-07-21
CI: hook-scripts.yml test-step list is a merge-conflict hotspot


auerbachb:
## Problem
`.github/workflows/hook-scripts.yml` lists each script/hook test as a sequential `- name:/run:` step, and every PR that adds a test appends its step at the **same anchor** (after the last existing step). When two PRs are in flight, the second to rebase hits a git conflict there — even though the additions are independent and both should simply coexist.

## Impact
Observed repeatedly this cycle: PRs #657, #658, #660, #661, #672 all added a test step and collided on this file. For PR #676 it forced **3 rebase cycles**. Because the collision makes the branch `CONFLICTING` (not cleanly `BEHIND`), the admin-merge-for-BEHIND path from #631/#658 does **not** apply — a human/agent must hand-resolve each time.

## Proposed solutions (pick one)
- Drive the `hook-tests` job from a **discovered list** (glob `.claude/scripts/tests/*.test.sh` + `.claude/hooks/tests/*.test.sh`) in a matrix/loop, so adding a test needs no YAML edit at all.
- Or keep the explicit list but **sort it deterministically** (e.g. by issue number) so additions distribute rather than always landing at the tail.

## Acceptance Criteria
- [ ] Two PRs that each add a new script test can both rebase onto each other's merge without a `hook-scripts.yml` conflict (or the file no longer needs per-test edits at all).
- [ ] All existing tests still run in CI.

## Notes
Follow-up surfaced by /wrap session sweep on PR #676 (Closes #657). Related: #631 (admin-merge for cleanly-BEHIND PRs) addresses the symptom for BEHIND PRs but not this CONFLICTING root cause.

_Filed via /wrap._

## Implementation Plan

Merges Claude's analysis + CodeRabbit's coding plan. Chosen approach: **runtime test discovery** (the root-cause fix — adding a test needs no YAML edit, so the shared tail-anchor conflict disappears entirely).

**Design decisions**
- **Single-job glob loop, NOT `strategy.matrix`.** A matrix expands per-entry check names (`hook-tests (…)`), which would silently break any branch-protection rule referencing the literal `hook-tests` / `hook-tests-py39` contexts. The loop preserves both job ids / check names. (Verified: today only `rule-lint` is a required check, but preserving the names also keeps the merge-gate scripts' check-name handling stable.)
[...]

PR #766, opened 2026-07-28
feat(#755): detect multi-PR file churn and auto-file refactor-candidate issues

### LINKED ISSUE #755: Detect multi-PR file churn and auto-file refactor-candidate issues
## Background

Some files become magnets: every feature seems to pass through them, and any long-lived branch that touches them pays a conflict tax on every sync. In a recent app-repo session, five separate PRs merged into the same form component in one sitting while a sixth in-flight PR — a ~350-line re-indent of that same file — re-resolved conflicts against it three times and rebased a fourth. The structural fix for that file did get filed as a follow-up, but only because a human happened to notice the pattern in the wrap-up summary.

## Problem

Nothing in the workflow watches for this. The signal — the same file appearing across many merged PRs in a short window, plus repeated conflict re-resolutions inside one session — is already sitting in merge history and session records, but nobody reads it. Churn hotspots only surface when someone eyeballs a wrap report, so a file that genuinely wants to be several smaller files keeps burning sessions' worth of conflict work before anyone flags it. A prior fleet-wide episode of the same species (#671) was likewise only visible because a wrap thread wrote it up by hand.

## Proposed solution

Make the wrap-up flow read the signal automatically. After a merge, scan recent merge history for files that keep appearing across distinct PRs — weighting actual conflict re-resolutions more heavily — and when a file crosses the churn threshold, automatically open a refactor-candidate issue for it with the evidence attached: which PRs, how many conflict rounds, over what window. Dedup so the same hotspot is never filed twice; if an open hotspot issue already exists for the file, append the new evidence as a comment instead. Hotspots then surface themselves as backlog items while the pain is fresh, instead of relying on someone noticing.

## Acceptance Criteria

- [ ] New script `.claude/scripts/churn-hotspots.sh [--since <date|ref>] [--threshold N] [--repo <owner/repo>]` lists files touched by >=N distinct merged PRs in the window (defaults: N=3, 14 days), with the PR numbers as evidence; exit 0 = hits found, 1 = clean
- [ ] Conflict re-resolutions observed during the session (e.g. recorded by `/fixpr` or the merge-conflict skill in session state) count toward the threshold with a documented weight
- [ ] The `/wrap` follow-ups phase runs the script post-merge and auto-files one refactor-candidate issue per new hotspot, reported in-thread with number/title/rationale/link (satisfies the autonomous-filing contract in `issue-planning.md`)
- [ ] Dedup before filing: an existing open hotspot issue for the same file gets an evidence comment appended, never a duplicate issue
- [ ] Works from `gh` data alone in a repo with no session-state history (fresh clone)
- [ ] Script has an offline bash unit test that runs in the hook-scripts CI job

## Test Plan

- [ ] Fixture history where 3 merged PRs touch the same file inside the window → script reports the file + the 3 PR numbers, exit 0
- [ ] Below-threshold history → exit 1 and `/wrap` files nothing
- [ ] Re-run with the hotspot issue already open → comment path exercised, no duplicate issue
- [ ] Offline test with a stubbed `gh` passes in CI

## Notes / Open questions

- Threshold tuning: is PR-count alone enough, or should at least one recorded conflict event be required before filing (fewer, higher-signal issues)?
- Should the auto-filed issue suggest *how* to split the file, or stay observational at first? (Lean observational — the CR plan that lands on the filed issue can propose the split.)
- Motivating incident: ItemForm.tsx in the app repo — five PRs (204, 205, 197, 211, 212 in that repo) merged into it during one session around an in-flight ~350-line re-indent; the app-side structural fix is already filed there (issue 224 in that repo). This issue is the workflow-side generalization so the *next* hotspot flags itself.

## Related Issues

- Related to #754 (removes the cost of clean-BEHIND churn once it happens; this issue reduces why churn recurs at all)
- Related to #671 (prior rebase-treadmill incident — same signal, hand-noticed)

_Captured via /issue-maker._

## Implementation Plan

_Merged: Claude's codebase analysis + CodeRabbit's coding plan. Divergences from CR are called out inline with rationale._

### Files

| File | Action |
|------|--------|
| `.claude/scripts/churn-hotspots.sh` | **new** — read-only detector |
| `.claude/scripts/tests/churn-hotspots.test.sh` | **new** — offline test (auto-discovered) |
| `.claude/skills/wrap/SKILL.md` | edit — new Step 3.10a in Phase 3 Part B |
| `.claude/scripts/README.md` | edit — scripts table row + tests table row |
| `.claude/reference/autofile-dedup.md` | edit — register the exact-match hotspot dedup path |
| `.claude/reference/churn-hotspots.md` | **new** — mechanism, weight rationale, tuning knobs |

No change to `CLAUDE.md` or `.claude/rules/*.md`, so the rule-corpus word budget is untouched. No change to `session-state-schema.json` (reuses the existing `babysit.conflict_streak` field) and none to `issue-planning.md` (autonomous filing is already permitted) — both considered and intentionally left alone.

### 1. `.claude/scripts/churn-hotspots.sh` (read-only detector)

```
churn-hotspots.sh [--since <date|ref>] [--threshold N] [--repo <owner/repo>]
                  [--conflict-weight N] [--exclude <glob,glob>]
                  [--no-default-excludes] [--source auto|git|gh] [--json]
churn-hotspots.sh --help
```

House style per `backlog-staleness.sh` / `issue-dedup.sh`: usage-log append, `print_help`/`err()`, `command -v` dependency checks, manual `while/case` parsing accepting both `--flag value` and `--flag=value`, bash-3.2-safe `REPO_ARGS` guard, documented `# EXIT CODES:` block.

**Enumeration — `--source auto` (default).**
1. **git path (primary, adopted from CR Design Choice 1).** `git log --since=<window> --no-merges --name-only` on the current checkout. Every PR here squash-merges, so each commit subject carries GitHub's appended `(#N)` marker. **Take the _last_ `(#N)` in the subject** — subjects like `fix(#749): … (#750)` contain the issue number first and the PR number last. Build `file → {distinct PR numbers}`. Zero API calls; works in a fresh clone.
2. **gh path (fallback).** Used when the git path yields no PR-marked commits in the window (merge-commit or rebase-merge repos) or when `--repo` names a repo other than the current checkout: `gh pr list --state merged --json number,title,mergedAt,url` filtered to the window, then `gh api repos/{owner}/{repo}/pulls/{N}/files` per PR, bounded by a `CHECK_LIMIT`-style cap (log what was dropped — never truncate silently). This path is what satisfies AC "works from `gh` data alone".

`--since` accepts a date (`YYYY-MM-DD`, used as-is) or a git ref/SHA (resolved via `git log -1 --format=%cI <ref>`). Default: 14 days via `gh-window.sh --days 14`.

**Conflict weighting (AC 2) — adopted from CR Design Choice 2.** For each in-window PR, read `.repos["<owner>/<repo>"].prs["<N>"].babysit.conflict_streak` through `session-state.sh --get` (scope from `--repo` or cwd). That is the only conflict signal persisted today — `/fixpr` and `/merge-conflict` keep their conflict data on transient stdout, so no new writer is added here. Missing file, missing scope, or missing key all read as `0`, which is what makes the fresh-clone AC hold. A forward-compatible generic `.prs["<N>"].conflict_rounds` is also read when present; the larger of the two is used, never the sum, so a babysit-driven `/fixpr` round is not double-counted.

**Score and threshold.**

```
score = distinct_pr_count + (conflict_weight × conflict_rounds)     # conflict_weight default 2
hotspot  ⇔  score >= threshold  AND  distinct_pr_count >= MIN_PRS   # MIN_PRS = 2
```

With zero conflict rounds `score == distinct_pr_count`, so the plain AC-1 reading ("files touched by >= N distinct merged PRs") holds exactly. **`MIN_PRS = 2` is a divergence from CR** and answers the ticket's own open question: without it a single PR with 3 conflict rounds scores 7 and would be filed as "multi-PR churn", which it is not. Conflict rounds accelerate a genuinely multi-PR file across the line; they can never manufacture a hotspot on their own.

**Default exclusions.** Generated files that churn by design and are never refactor candidates: `package-lock.json`, `yarn.lock`, `pnpm-lock.yaml`, `Cargo.lock`, `poetry.lock`, `go.sum`, `CHANGELOG.md`. Extend with `--exclude`, disable with `--no-default-excludes`. Deliberately universal only — repo-specific by-design churn (e.g. this repo's `.claude/scripts/README.md`, which every new script edits) is left to `--exclude` rather than baked in.

**Existing-issue lookup (feeds AC 4).** For each hotspot, find any already-open hotspot issue for that exact path. `gh issue list --state open --search "Refactor hotspot in:title" --json number,title,body --limit 200`, then **match `title == "Refactor hotspot: <path>"` exactly, client-side** (or an exact `<!-- churn-hotspot: <path> -->` body marker). **Divergence from CR:** CR searches `'"<path>" in:title'` and trusts the result; GitHub tokenizes paths, so that can match a sibling file (`src/ItemForm.tsx` vs `src/OtherForm.tsx`). Searching broadly and comparing exactly is deterministic and offline-testable. The 200-issue cap is logged when hit.

**Output.** Default: TSV rows `path  pr_count  conflict_rounds  score  pr_numbers  existing_issue`. `--json`: one object per hotspot with `file`, `pr_count`, `pr_numbers`, `conflict_rounds`, `conflict_prs`, `score`, `first_merged_at`, `last_merged_at`, `existing_hotspot_issue`, wrapped in a envelope carrying `repo`, `since`, `threshold`, `conflict_weight`, `min_prs`, `source`, `scanned_pr_count`, `excluded`, `truncated`. Valid (possibly empty) output on every exit path.

**Exit codes.** `0` ≥1 hotspot, `1` clean, `2` usage, `3` env error (missing dep / not a git repo), `4` gh/API error. (**Divergence from CR**, which proposed `3` for both env and gh errors — the repo-wide convention is `3` not-found/env, `4` gh error.)

The script **never writes** — no issue creation, no comments, no state mutation. That matches every other detector here (`backlog-staleness.sh`, `clean-behind-check.sh`) and keeps the filing decision inside `/wrap`, where the reporting contract already lives.

### 2. `.claude/scripts/tests/churn-hotspots.test.sh` (AC 6)

Hand-rolled harness matching `backlog-staleness.test.sh`: `PASS`/`FAIL` counters, `mktemp -d` + `trap cleanup EXIT`, `export HOME=<tmp>`, summary line, non-zero exit on any failure. No CI edit — `run-hook-tests.sh` globs it. Builds a real temp git repo with squash-style subjects carrying `(#N)` markers and stubs `gh` on `PATH` (`case` on args, catch-all `exit 1`).

Cases:
1. 3 distinct PRs touch one file inside the window → file + all 3 PR numbers reported, exit `0`.
2. Below threshold → exit `1`, empty hotspot list.
3. Subject carrying both an issue and a PR marker → the **last** `(#N)` wins.
4. Conflict weight: seeded `babysit.conflict_streak` pushes a 2-PR file over a threshold of 3.
5. `MIN_PRS` floor: 1 PR + 3 conflict rounds → still clean, exit `1`.
6. No session-state file at all (fresh clone) → runs, conflict rounds `0`, exit unchanged.
7. `gh` stub returns no matching issue → `existing_hotspot_issue` null (file path).
8. `gh` stub returns an open `Refactor hotspot: <path>` issue → `existing_hotspot_issue` populated (comment path).
9. Near-miss title (`Refactor hotspot: src/Other.tsx`) → **not** matched (exact-match guard).
10. Default exclusions drop a lockfile that would otherwise score above threshold.
11. `--source gh` with a stubbed `gh` → gh-only enumeration works (the "gh data alone" AC).

Stub authoring note per prior lessons: match on a **single stable token**, never on `--json` field order; avoid `${VAR:-default}` inside the stub heredoc.

### 3. `/wrap` wiring — new Step 3.10a (AC 3, AC 4)

Insert **Step 3.10a: Category 5a — Churn hotspots (issue #755)** into Phase 3 Part B, immediately after Step 3.10 (PM hygiene). **Divergence from CR:** CR proposed a new numbered category with "renumbering subsequent steps"; a lettered sub-step (the file already uses `Step 1.1f`) achieves the same placement with none of the renumber diff — fitting for a ticket about reducing churn.

It runs post-merge and after Step 2.5's root-main sync, so `main` history is fresh:

```bash
CHURN_SH=$(resolve_script churn-hotspots.sh || true)   # Step 3.5 helper
CHURN_JSON='{}'
if [ -n "$CHURN_SH" ]; then
  CHURN_JSON=$("$CHURN_SH" --json 2>/dev/null) || CHURN_RC=$?
fi
```

Exit `1` (clean) or a missing/failed script files nothing and adds no report line. For each hotspot:

- **`existing_hotspot_issue` present** → `gh issue comment` an evidence update (PR numbers, conflict rounds, window) and record `Appended to #{N} instead of filing — churn hotspot <path>` for the Step 4.3 suppressed-duplicates line. Never a second issue.
- **absent** → `gh issue create --title "Refactor hotspot: <path>"` with an **observational** body (per the ticket's lean): provenance line, the evidence table, the `<!-- churn-hotspot: <path> -->` marker, and the `_Filed via /wrap._` footer. Body stays observational — the CR plan that lands on the filed issue proposes the split. Use the same `gh issue create` guard as Step 3.3 (validate the parsed number; on failure record and continue, never abort Phase 3).

Each filed issue appends `{number, title, keywords, rationale: "churn hotspot: <path>"}` to `WRAP_FILED_ISSUES`, its number to `DEDUP_EXCLUDE` and to `SWEEP_FILED`, so it flows into the Step 4.3 report (number + title + one-line rationale + clickable link — satisfying the `issue-planning.md` autonomous-filing contract) and into the Step 3.13 `wrap_sweep` idempotency record.

The step text states explicitly that this category's authoritative re-find is the script's exact-title lookup, **not** the fuzzy `dedup_search()` helper.

**No `hotspot` label. Divergence from CR:** `gh issue create --label hotspot` fails outright when the label does not exist, and `/wrap`'s own Step 3.3 contract requires filing to be "zero-config and requires no repo label to exist". The title convention plus the body marker give exact identity without a repo-setup dependency.

### 4. Docs

- `.claude/scripts/README.md` — scripts-table row (signature, behavior, JSON fields, conflict-weight caveat, "Called by `/wrap` Step 3.10a", exit codes) placed after `backlog-health.sh`; matching `## tests/` row.
- `.claude/reference/autofile-dedup.md` — register the exact-match hotspot path as a narrower complement to the fuzzy `issue-dedup.sh` flow; note the new `/wrap` category in the filer inventory.
- `.claude/reference/churn-hotspots.md` — mechanism, the weight formula and why `MIN_PRS = 2`, tuning knobs, the `git`-vs-`gh` source decision, and the marker/title contract. Reference docs are not auto-loaded, so this costs no rule budget.

### Risks / edge cases

- **Subjects with two `(#N)` markers** — take the last (test 3).
- **Merge-commit repos** — `--name-only` prints nothing for merge commits, so the git path finds no data and `auto` falls through to gh.
- **By-design churn** (`README.md`, index files) — mitigated by default exclusions plus `--exclude`; the `MIN_PRS` floor and threshold keep volume low. Worst case is one closable issue, and every filing is announced.
- **Renames** — a renamed file reads as two paths; `git log --follow` is single-path only, so renames are out of scope and documented as such.
- **Silent truncation** — both the 200-issue lookup cap and the gh-path PR cap log what was dropped.
- **`set -e` + `$(cmd)`** — capture with `RC=0; VAR=$(cmd) || RC=$?` (prior lesson).

### Verification

Each AC maps to a test or an inspectable artifact: AC 1 → tests 1–3 + `--help`; AC 2 → tests 4–5 + the reference doc's documented weight; AC 3 → Step 3.10a text + the Step 4.3 report path; AC 4 → tests 7–9 + the comment branch; AC 5 → tests 6 and 11; AC 6 → the suite running green under `bash .github/scripts/run-hook-tests.sh` and in the `hook-scripts` CI job.

auerbachb:
Closes #755

## What this does

Files that keep reappearing across merged PRs charge every long-lived branch a conflict tax. The signal — the same path across many distinct merged PRs in a short window — already sits in merge history, but nothing read it, so hotspots only surfaced when someone noticed the pattern in a wrap-up summary.

This adds a detector that reads it mechanically, and wires `/wrap` to file a refactor-candidate issue for the worst offender after a merge, while the pain is fresh.

Validation that the signal is real: run against this repo's last 14 days, the top hotspot is `.github/workflows/hook-scripts.yml` at **24 distinct PRs** — precisely the file issue #681 rewrote *because* it had become a recurring merge-conflict hotspot. This would have flagged it automatically.

## Scoring

```
score = distinct_pr_count + (conflict_weight × conflict_rounds)   # weight default 2
hotspot  ⇔  score >= threshold (default 3)  AND  distinct_pr_count >= 2
```

With no conflict data the score *is* the PR count, so the plain AC-1 reading holds exactly. The `>= 2` floor answers the open question the issue raised: a conflict event is **not** required to file, but conflicts also cannot substitute for multi-PR churn — one PR that re-resolved conflicts three times would otherwise score 7 and be reported as "multi-PR churn" when a single PR touched the file. That is a different problem (issue #754's territory) with a different fix.

Conflict rounds come from `babysit.conflict_streak`, maxed with a forward-compatible `conflict_rounds` (never summed, so a babysit-driven `/fixpr` round is not double-counted), and read as `0` when absent.

## Volume control — the thing worth reviewing

The detector is sensitive by design, and it found **63 files above the default threshold** in one 14-day window here (33 at threshold 5, 9 at threshold 10). Auto-filing all of those would bury the backlog the feature exists to inform.

So `/wrap` Step 3.10a files **at most one** new hotspot issue per run — the highest-scoring one with no existing issue — and always reports how many candidates the cap held back. Because each filed issue is then found by the lookup, successive wraps work down the list highest-signal-first. The calibration table is in the reference doc so the tuning advice is grounded in measurement rather than invented.

An existing hotspot issue gets an evidence comment instead of a duplicate — but **only when the just-merged PR is one of that hotspot's PRs**. Without that guard every later wrap would append another comment restating the same history.

## Divergences from the CodeRabbit plan

CR's plan was adopted nearly wholesale (git-log enumeration over API fan-out, read-only script, `Refactor hotspot: <path>` title convention). Four deliberate departures:

1. **No `hotspot` label.** `gh issue create --label` fails outright when the label doesn't exist, and `/wrap`'s own contract requires filing to be "zero-config and requires no repo label to exist". The title convention plus a `<!-- churn-hotspot: <path> -->` body marker give exact identity with no repo-setup dependency.
2. **Exact client-side matching**, not `gh issue list --search '"<path>" in:title'`. GitHub tokenizes paths, so a server-side path query matches sibling files — `src/Form.tsx` would hit an issue about `src/OtherForm.tsx`. Covered by test 9a.
3. **`min_prs >= 2` floor** (above) — CR's plan had no floor.
4. **Lettered `Step 3.10a`** instead of a new numbered category "renumbering subsequent steps". Same placement, none of the renumber diff — which seemed apt for a churn-reduction change.

## Local review status

Both local CLIs were unavailable, so this went in on a **self-review** per `cr-local-review.md`'s both-CLIs-down fallback. Noting it explicitly since self-review does not satisfy the merge gate — the GitHub App reviewers do, and their quotas are independent of the CLIs.

- **CodeAnt CLI** — not installed on this machine (`codeant: command not found`).
- **CodeRabbit CLI** — `rate_limit`: "You've used all free OSS reviews for now", 18-minute reset. Per the rule this is a *failed run*, not a clean pass, and a failed CLI is not retried more than once.

The self-review did catch a real bug that is fixed here: without `git -c core.quotePath=false`, a non-ASCII path is emitted as `"src/caf\303\251.ts"` — mangled *and* wrapped in quotes that would corrupt the file key the issue lookup matches on. Regression test 3d locks it in.

Two further bugs were found and fixed while building the test suite, both worth knowing about:

- Under **bash 3.2** (macOS), `${line#$'\x01'}` silently fails to strip even where the matching `case` pattern succeeds. The control byte leaked into every timestamp while PR numbers parsed fine. The parser now uses plain ASCII sentinels and relies on no ANSI-C quoting anywhere.
- `grep -c` on an empty file prints `0` **and** exits 1, so a `|| echo 0` fallback emitted `"0\n0"` and aborted the downstream `jq --argjson` with no output at all. Counters now use `wc -l` plus an `as_number` coercion before any jq call.

## Test plan

- [x] Fixture history where 3 merged PRs touch the same file inside the window → script reports the file + the 3 PR numbers, exit 0
- [x] Below-threshold history → exit 1 and `/wrap` files nothing
- [x] Re-run with the hotspot issue already open → comment path exercised, no duplicate issue
- [x] Offline test with a stubbed `gh` passes in CI
- [x] Conflict rounds count toward the threshold with a documented weight, and cannot alone create a hotspot
- [x] Works with no session-state history at all (fresh clone), and via `gh` data alone
- [x] A near-miss hotspot title (sibling file) does not suppress a real finding
- [x] Full suite green: `bash .github/scripts/run-hook-tests.sh` and the Python discovery job


auerbachb:
@graphite-app re-review

auerbachb:
Confirmed and fixed in d8f84d6.

You're right that this was a duplicate-filing risk. A lookup that hits ISSUE_LOOKUP_CAP is an *incomplete* lookup, so "no match" stops proving "no issue" — and since callers gate filing on existing_lookup_failed rather than truncated, they were free to file a duplicate for a hotspot whose issue sat beyond the cap. The cap branch now sets existing_lookup_failed as well as truncated, and the stderr warning says it is not safe to file.

Regression test 9b/9c in churn-hotspots.test.sh stubs a 200-issue result and asserts both flags.

auerbachb:
Confirmed and fixed in d8f84d6.

The mismatch was real and slightly worse than described: --since is day-granular (ET-anchored via gh-window.sh, or a git ref resolved to that commit's date) while mergedAt is a UTC instant, so the gh path's window disagreed with what `git log --since` gives the git path for the same input. On the ref path it was outright unsound — a resolved ref carries a local offset (2026-03-01T18:00:00-05:00) that string-compares incorrectly against UTC Z timestamps.

Both paths now compare calendar date to calendar date, inclusive at the start boundary. The dead SINCE_ISO variable is gone.

Test 11i covers the ref case specifically; I verified it fails against the pre-fix comparison and passes after.

auerbachb:
Confirmed and fixed in d8f84d6 — this was the best catch of the three.

The step selected "highest-scoring hotspot whose existing_hotspot_issue is null" and *then* branched on whether that hotspot had an existing issue, so the append-evidence branch was unreachable by construction. Evidence would never have reached an existing hotspot issue.

The two sets are now chosen independently, because they have genuinely different cardinality rules:
- Comment set: every hotspot that already has an issue AND whose pr_numbers include the merging PR. Uncapped — it adds no tickets and is naturally bounded, since a merging PR touches few files and contributes to each at most once.
- File set: at most one, the highest-scoring hotspot with no existing issue.

The cap exists to stop a first run on an active repo from filing dozens of tickets at once (63 candidates above threshold here in one 14-day window); that reasoning never applied to comments, which is what the conflated selection accidentally suppressed.

auerbachb:
@graphite-app re-review

PR #15, opened 2026-03-19
Add /standup skill for daily standup summaries

auerbachb:
## Summary

- Adds `.claude/skills/standup/SKILL.md` — a user-invocable `/standup` skill
- Queries GitHub for closed issues, merged PRs, and open PRs since a given time
- Generates a business-focused standup summary (<150 words) with scale metrics
- Uses dynamic TZ-aware ISO 8601 timestamps (handles EST/EDT correctly)
- Manual trigger only (`disable-model-invocation: true`)

Also installed as a global skill at `~/.claude/skills/standup/` so it works from any repo.

Closes #14

## Test plan

- [x] Skill file exists at `.claude/skills/standup/SKILL.md`
- [x] Skill is user-invocable via `/standup` with optional time argument
- [x] Skill has `disable-model-invocation: true` (manual trigger only)
- [x] Instructions specify GitHub CLI commands for data gathering
- [x] Output format includes accomplishments, in-progress work, and scale metrics
- [x] ISO 8601 timestamp uses dynamic TZ offset (not hardcoded `-05:00`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a standalone "standup" skill that generates concise daily standup reports summarizing closed issues, merged PRs (with additions/deletions), and open PRs authored by you.
  * Accepts an optional "since-time" (eg. “yesterday at noon ET”) and converts natural-language times to an ISO 8601 cutoff with correct UTC offset.
  * Determines relevant repositories from your context; produces first-person, structured reports ("Since [time]", "In progress", "Scale"), ≤150 words, aggregated items, and optional "~lines changed".

PR #807, opened 2026-07-30
feat(#803): hard 20-minute ceiling on chat silence while background work runs

### LINKED ISSUE #803: Hard 20-minute ceiling on chat silence while background work runs
## Background

When a thread hands work to a subagent or kicks something off in the background, the thread itself goes quiet — the parent's job becomes watching, and watching mostly means waiting. Today that wait is invisible. You look at the chat and the last thing anyone said was 24 minutes ago, with no way to tell whether the subagent is grinding away, wedged, or dead.

We already claim this doesn't happen. CLAUDE.md's third non-negotiable behavior says a thread never goes more than 5 minutes without a status message, and there is real machinery behind it: a pair of hooks that nag the agent once it's been quiet too long, plus a macOS watchdog that raises a desktop notification after 10 minutes of stall.

## Problem

The 5-minute promise silently stops applying in exactly the situation where it matters most.

The in-context nag only fires **after a tool call**. When the parent spawns a subagent and ends its turn to wait for the completion notification, no tool calls happen — so nothing checks, nothing warns, and the thread can sit silent for as long as the subagent takes. The desktop watchdog does notice the stall, but all it can do is pop an OS notification; it has no way to make the thread say anything. Either way the transcript shows an unexplained hole, and the person watching has to ask "status?" to find out anything at all.

Underneath both symptoms is one cause: **printing a message requires a turn, and nothing in the spawn path guarantees a turn will ever happen.** We are strict about this for user-facing polls — never promise "I'll check back in N minutes" without a real timer behind it — but monitoring background work was never classified as polling, so spawning a subagent arms nothing.

## Proposed solution

Treat "this thread has background work in flight" as a state that mechanically guarantees a turn boundary, so silence has a hard ceiling no matter what the model remembers to do.

The moment a thread starts something it will wait on — a subagent, a long-running process, a watcher — it arms a timer. If nothing else has produced a chat message by the time the ceiling arrives, the timer fires and one gets printed, even when the only honest content is "still running, nothing new." The person watching always knows the difference between working and wedged, and never has to prompt for it.

Two design calls worth making explicitly:

- **5 minutes stays the only cadence anyone states.** The ceiling is a backstop that should never read as a target. Publishing two numbers invites the looser one to become the norm — a rule that says "no longer than 20 minutes" quietly authorizes 19 minutes of silence, which is worse than what we have written today.
- **Scope is any in-flight background work,** not just subagents — long-running commands, watchers, and scheduled loops all produce the same hole.

## Acceptance Criteria

- [ ] Starting background work a thread will wait on (subagent spawn, background process, watcher) arms a wakeup timer as part of that same step — not as a separate thing the model has to remember
- [ ] The armed timer guarantees a chat message within a hard ceiling of **20 minutes** from the last visible message, whichever background work is running
- [ ] The ceiling holds in the failing case from this report: parent spawns a subagent, ends its turn, and waits on the completion notification with zero intervening tool calls
- [ ] The ceiling-triggered message is a normal timestamped heartbeat (`monitor-mode.md` "User Heartbeat" one-line format), including when there is nothing new to report
- [ ] `20` appears only in enforcement (script/hook/state), never as a stated cadence in `CLAUDE.md` or any rule file — the 5-minute cadence in CLAUDE.md #3 stays the single published number
- [ ] `scheduling-reliability.md` classifies "thread has background work in flight" as a polling context, so its existing pre-exit checklist (next tick scheduled / heartbeat sent / state recorded) applies
- [ ] `subagent-orchestration.md` spawn checklist and `monitor-mode.md` monitor loop both reflect the arm-a-timer step
- [ ] Failing to arm the timer fails closed and is visible — a silent no-op re-creates the current bug (`feedback_guard_must_fail_closed.md`)
- [ ] Rule corpus stays under the enforced word budget; net additions offset elsewhere if needed

## Test Plan

- [ ] Spawn a subagent that runs >20 minutes, end the turn, and confirm a heartbeat is printed before the 20-minute mark with no user prompting
- [ ] Same scenario with a background `Bash` process instead of a subagent — confirm identical behavior
- [ ] Confirm a thread producing normal 5-minute heartbeats never triggers the ceiling path (no duplicate/spurious messages)
- [ ] Confirm the ceiling message survives context compaction — a thread that compacts mid-wait still reports within the ceiling
- [ ] `grep -rn '20' CLAUDE.md .claude/rules/` shows no stated 20-minute cadence
- [ ] `.claude/scripts/rule-lint.sh` passes; corpus word count within the committed ratchet cap
- [ ] Existing hook tests still pass (`.claude/hooks/tests/silence-detector-time-dedupe.test.sh`)

## Notes / Open questions

- **Why 20 and not 10 or 15.** 20 is ~4x the 5-minute cadence — far enough above the target that it only fires on genuine failure, never on normal jitter. A backstop that trips during healthy operation gets tuned out. The existing `silence-watchdog.sh` uses 10 minutes for its desktop notification; whether to align the two or keep the in-chat ceiling deliberately looser is open.
- **Which primitive arms the timer.** `ScheduleWakeup` (single tick, re-armed each turn) and `/loop` (runtime owns cadence) both create turn boundaries; `CronCreate` is the durable option. `scheduling-reliability.md` already forbids hand-rolled one-shot chains for recurring polls, which argues for `/loop` — but a subagent wait is closer to "one long wait with a deadline" than a recurring poll, so this needs a call.
- **Should the macOS watchdog be upgraded too?** Today it can only notify the OS. Giving it a path to inject into the session would cover the case where the timer itself was never armed — a belt-and-braces layer, but it may need harness support we don't have. Possibly a follow-up rather than part of this change.
- **Interaction with stable-state backoff.** `scheduling-reliability.md` widens polling intervals to 5m/15m on stable state and deletes the job at streak ≥9. The ceiling must survive backoff — a widened or deleted poll job must not be able to push silence past 20 minutes.
- **Related closed work:** issue #773 (silence-detector time-injection dedupe) and issue #715 (babysitting PRs by default) both touched this machinery without closing this hole.

## Related Files

- `CLAUDE.md` — non-negotiable behavior #3 (5-minute heartbeat)
- `.claude/rules/monitor-mode.md` — monitor loop, User Heartbeat format
- `.claude/rules/scheduling-reliability.md` — polling primitives, pre-exit checklist, backoff
- `.claude/rules/subagent-orchestration.md` — spawn checklist
- `.claude/hooks/silence-detector.sh` — PostToolUse nag (can't fire with no tool calls)
- `.claude/hooks/silence-detector-ack.sh` — Stop hook heartbeat touch
- `.claude/scripts/silence-watchdog.sh` — launchd watchdog, desktop notification only

_Captured via /issue-maker._

## Implementation Plan

_Source: Claude's analysis only — CodeRabbit returned the issue-enrichment boilerplate (unchecked "Create Plan" checkbox), not a substantive plan._

### Key finding that shapes the design

The measurement this issue needs **already exists**: `/tmp/claude-heartbeat-$SESSION_ID`, touched by the `silence-detector-ack.sh` Stop hook on every visible response. Its mtime *is* "time of last visible message."

What is missing is a **turn-independent observer of that file that can force a turn**:

- `silence-detector.sh` (PostToolUse) reads it correctly but only runs after a tool call — never in the failing case.
- `silence-watchdog.sh` (launchd) reads it out-of-turn, but (a) can only raise an OS notification, and (b) at `silence-watchdog.sh:63` it **`continue`s when `/tmp/claude-active-$SID` is absent** — and that marker is removed by the Stop hook. It therefore deliberately skips the exact "ended its turn and is waiting" state this issue is about.

So the fix is not new measurement — it is one out-of-turn observer plus a fail-closed gate that makes arming it non-optional.

### Design call: `Monitor`, not `ScheduleWakeup` / `CronCreate`

Resolves the issue's "Which primitive arms the timer" open question:

- **`ScheduleWakeup`** is the `/loop` dynamic-mode scheduler. It is unavailable in a plain coding thread — exactly where the hole is. Rejected.
- **`CronCreate`** fires unconditionally on wall-clock cadence, so a healthy thread already emitting 5-minute heartbeats would get spurious ceiling messages (fails Test Plan item 3). Its jitter (up to 10% of period) also eats into the ceiling, and `durable: true` is a documented no-op — it is session-only anyway. Rejected.
- **`Monitor`** runs an out-of-turn shell process whose stdout lines become chat events. It is (a) turn-independent → satisfies AC3, (b) **conditional** — the loop prints nothing unless the heartbeat file is genuinely stale, so a healthy thread never sees a ceiling message, (c) `persistent: true` covers the whole session, so the timer is armed **once per session** rather than once per spawn, and (d) stopped only by `TaskStop` — `CronUpdate`/`CronDelete` cannot reach it, which answers the issue's "must survive stable-state backoff" open question structurally rather than in prose.

Arm-once-per-session also sidesteps having to detect *completion* of background work: the watch stays correct whatever the cause of silence.

### Files to create

| Path | Role |
|------|------|
| `.claude/scripts/bgwork-ceiling.sh` | **Sole owner of the number.** `CEILING_MINUTES=20` (env `CLAUDE_BGWORK_CEILING_S`). Modes: `--note-started <kind>`, `--record-armed`, `--check` (exit 1 = bg work in flight + unarmed), `--tick` (breach line or silence — this is what the Monitor loop runs), `--arm-command` (prints the exact `Monitor` command), `--status` (JSON), `--clear`. |
| `.claude/hooks/bgwork-ceiling-arm.sh` | PostToolUse (all tools). Marks in-flight on `Agent`, `Bash` with `run_in_background: true`, and `Workflow`; marks armed when the ceiling `Monitor` is seen (sentinel match). Injects `additionalContext` with the exact arm call while in-flight-and-unarmed. |
| `.claude/hooks/bgwork-ceiling-guard.sh` | Stop hook. **Fail-closed:** `--check` failure → `{"decision":"block","reason":…}` so the turn cannot end silently. Bounded to 2 consecutive blocks; past the bound it stops blocking but writes a loud stderr line + a breach record that the PostToolUse hook resurfaces — degraded, never silent (`feedback_guard_must_fail_closed.md`). |
| `.claude/scripts/tests/bgwork-ceiling.test.sh` | Script contract: ceiling arithmetic, `--check` exit codes, `--tick` silent-vs-breach, unwritable-state handling. |
| `.claude/hooks/tests/bgwork-ceiling-hooks.test.sh` | Hook contract: in-flight detection per tool shape, arm detection, Stop-hook block JSON, block-bound degradation. |
| `.claude/reference/bgwork-ceiling.md` | Mechanism, primitive-selection rationale, backoff interaction, failure modes. Not auto-loaded — keeps the rule corpus flat. |

### Files to modify

| Path | Change |
|------|--------|
| `global-settings.json` | Register both hooks (PostToolUse all-tools; Stop). |
| `.claude/rules/scheduling-reliability.md` | Add "thread has background work in flight" as a polling context so the existing pre-exit checklist applies (AC6); decision-tree row; state that backoff must not reach the ceiling watch. |
| `.claude/rules/subagent-orchestration.md` | Arm-the-ceiling step in the spawn checklist (AC7). |
| `.claude/rules/monitor-mode.md` | Arm/verify step in the per-cycle monitor loop (AC7). |
| `.claude/hooks/README.md` | Document both hooks (free — not auto-loaded). |
| `.claude/reference/diagrams/hook-lifecycle.md` | Add both hooks to the sequence. |
| `.claude/reference/scheduling-failure-modes.md` | Pattern 6 — "Background work with no armed observer". |
| `.claude/reference/README.md` | Index the new reference doc. |

`CLAUDE.md` is deliberately **not** edited: AC5 keeps the 5-minute cadence as the only published number, and AC7 does not ask for it.

### Implementation steps

1. Write `bgwork-ceiling.sh` with the full `--help` contract and documented exit codes, matching the house style of `greptile-budget.sh`. Ceiling constant + derived poll interval live here and nowhere else.
2. Write `bgwork-ceiling-arm.sh` (PostToolUse) — parse `tool_name` / `tool_input.run_in_background`; sentinel-match the ceiling watch so arming the watch is not itself counted as new background work.
3. Write `bgwork-ceiling-guard.sh` (Stop) — block JSON, `stop_hook_active` awareness, bounded consecutive-block counter, loud degradation.
4. Register both in `global-settings.json`.
5. Add both test suites (auto-discovered by `.github/scripts/run-hook-tests.sh` — **no workflow edit**, issue #681).
6. Rule edits (2–4 sentences each, no number stated).
7. Reference docs + hooks README + lifecycle diagram + failure-mode pattern.
8. Budget pass: `.github/scripts/rule-lint.sh` must pass against the committed ratchet cap; trim to offset additions.

### Risks / edge cases

- **Budget is the binding constraint.** Corpus is 12,166 words against a committed ratchet cap of 12,310 — **144 words of headroom**. Rule additions must be surgical and offset; all prose detail goes to `.claude/reference/`.
- **Stop-hook block loops.** An unbounded block would hang the thread. Bounded at 2 consecutive blocks, `stop_hook_active`-aware, degrading loudly rather than silently.
- **Double-counting the arm call.** The ceiling `Monitor` is itself background work; the sentinel match must classify it as *arming*, never as new in-flight work, or the guard never clears.
- **bash 3.2 / BSD userland.** macOS is the primary host — `stat -f %m` vs `stat -c %Y`, no `${var#$'\x01'}` control-byte tricks (`feedback_bash32_control_byte_strip.md`).
- **Unwritable `/tmp` marker.** The silence detector fails *open* there by design; this guard must fail *closed* — surface, do not swallow.
- **Stale path in the issue's Test Plan.** `rule-lint.sh` lives at `.github/scripts/rule-lint.sh`, not `.claude/scripts/`. Verification uses the real path.

### Verification

| AC | How it is verified |
|----|--------------------|
| Arms as part of the same step | PostToolUse hook injects the arm instruction on the background-work call itself; hook test asserts the injected context. |
| Hard ceiling from last visible message | Script test drives `--tick` against a back-dated heartbeat file and asserts silent-below / breach-at-or-above. |
| Holds with zero intervening tool calls | The watch is an OS process reading `/tmp`; test asserts `--tick` breaches with no hook invocation in between. |
| Message is a normal one-line timestamped heartbeat | Breach line instructs the `monitor-mode.md` "User Heartbeat" format; asserted in the script test. |
| `20` only in enforcement | `grep -rn` over `CLAUDE.md` + `.claude/rules/` for a stated 20-minute cadence returns nothing. |
| `scheduling-reliability.md` classifies bg work as polling | Rule diff + `rule-lint.sh`. |
| Spawn checklist + monitor loop updated | Rule diffs. |
| Fails closed and visible | Hook test asserts the Stop-hook `decision: block` payload and the loud degradation past the bound. |
| Corpus under budget | `.github/scripts/rule-lint.sh` exits 0. |

auerbachb:
Closes #803

## What this fixes

`CLAUDE.md` #3 promises a user-visible message at least every 5 minutes. That promise silently stopped applying in the one situation where it matters most: a thread spawns a subagent, ends its turn, and waits. The in-context nag is a **PostToolUse** hook, and a thread waiting on a completion notification makes no tool calls — so nothing checked and nothing warned. The launchd watchdog saw the stall but could only raise an OS notification, and it `continue`s when `/tmp/claude-active-<id>` is absent — a marker the Stop hook removes — so it **deliberately skipped the ended-its-turn-and-waiting state**.

One cause underneath both: printing a message requires a turn, and nothing in the spawn path guaranteed a turn would ever happen.

## Approach

Measurement is not re-implemented. `/tmp/claude-heartbeat-<session-id>` already means "time of last user-visible message" and stays the single source of truth. What was missing is a **turn-independent observer of it that can force a turn**, plus a gate that makes arming that observer non-optional.

- **`.claude/scripts/bgwork-ceiling.sh`** — sole owner of the ceiling duration and the derived trip point / poll interval. Full contract in `--help`.
- **`.claude/hooks/bgwork-ceiling-arm.sh`** (PostToolUse) — records background work (`Agent`, `Workflow`, `Monitor`, backgrounded `Bash`) and injects the exact arming call. **Advisory.**
- **`.claude/hooks/bgwork-ceiling-guard.sh`** (Stop) — returns `decision: block` when background work is in flight and the ceiling is unarmed, so the turn cannot end silently. **This is the enforcement.**

### Which primitive arms it — the issue's open question, resolved

| Primitive | Verdict |
|-----------|---------|
| `ScheduleWakeup` | **Rejected** — it is the `/loop` dynamic-mode scheduler, unavailable in a plain coding thread, which is exactly where the hole is. |
| `CronCreate` | **Rejected** — fires on wall-clock cadence whether or not the thread is silent, so a healthy thread would get spurious ceiling messages; jitter eats into the ceiling; `durable: true` is a documented no-op. |
| `Monitor` | **Chosen** — an out-of-turn process whose stdout lines become chat events. |

`Monitor` is the only option that is turn-independent *and* conditional (silent unless the thread really goes quiet) *and* armed once per session via `persistent: true` *and* out of reach of stable-state backoff — backoff widens and deletes **cron** jobs, and a `Monitor` is stopped only by `TaskStop`, so the issue's "must survive backoff" requirement is met structurally rather than by a prose carve-out.

Arming once per session also sidesteps detecting *completion* of background work: the watch stays correct whatever the cause of silence.

### Why the number is not published

A backstop must never read as a target — "no longer than 20 minutes" quietly authorises 19 minutes of silence. The constant lives in `bgwork-ceiling.sh`; every rule file says "the ceiling" or names the script. `CLAUDE.md` #3's 5-minute cadence stays the only published number.

### Fail-closed, bounded, never silent

Per `feedback_guard_must_fail_closed.md`, the default is to **block**, not warn. Blocking is bounded at 2 consecutive turns (an unbounded block would hang the thread); past the bound the guard stands down **loudly** — stderr, `~/.claude/logs/bgwork-ceiling.log`, and an unguarded marker the arm hook resurfaces on every later tool call. The advisory rate-limit deliberately does **not** apply once unguarded: throttling the only remaining signal would be the same silent degradation this change removes.

## Test plan

- [x] Spawning background work and ending the turn produces a heartbeat inside the ceiling, unprompted — the end-to-end case is replayed in `bgwork-ceiling-hooks.test.sh` #15 (spawn → blocked → armed → turn ends → watch breaks the silence with **zero** intervening hook invocations). The timing guarantee is asserted as the invariant `trip + poll <= ceiling` rather than by waiting out a real 20 minutes, so it holds for any retuning.
- [x] Identical behaviour for a background `Bash` process instead of a subagent — `bgwork-ceiling-hooks.test.sh` #2 covers `Agent` / `Workflow` / `Monitor` / `Bash(run_in_background)`; #3 confirms a foreground `Bash` is *not* treated as background work.
- [x] A thread producing normal 5-minute heartbeats never triggers the ceiling path — `bgwork-ceiling.test.sh` #9 (silent on a fresh heartbeat and below the trip point) and #11 (one breach per stretch of silence, not one per poll).
- [x] The ceiling survives context compaction — verified structurally rather than by forcing a compaction: neither half lives in context (the watch is an OS process, the state is `/tmp` markers), and the suites exercise that state across independent process invocations. Called out under "Known limits" in the reference doc.
- [x] No stated 20-minute cadence in `CLAUDE.md` or any rule file — `grep -rnE '\b20[- ]?(min|minute)' CLAUDE.md .claude/rules/` returns nothing, and the literal appears in no rule file.
- [x] `rule-lint.sh` passes and the corpus is within the committed ratchet cap — 11,171 words against a cap of 11,749, and now under the 12,000 soft budget as well, so this adds no warning (note: the script lives at `.github/scripts/rule-lint.sh`, not the `.claude/scripts/` path the issue cites).
- [x] Existing hook tests still pass — `silence-detector-time-dedupe.test.sh` green, and the full auto-discovered run is 47/47 suites, 0 failures.
- [x] The two new suites are picked up with no workflow edit (issue #681 auto-discovery) and `shellcheck -S warning` is clean on all three new scripts.

## Review notes

- **CodeAnt CLI was not run** — it is not installed on this host and its auth is browser-only (rung 4 of the `safety.md` capability ladder), so per `cr-local-review.md` it is dropped for this session and noted here. The CodeAnt **GitHub App** is unaffected and satisfies the merge gate on its own.
- **CodeRabbit CLI: one verified-successful pass, then a stall.** Round 1 completed normally with 6 findings (triaged below). Round 2, re-run after staging so it would also cover the new scripts, sat in `status: reviewing` for 16+ minutes with no findings and no error — well past the 2-minute hang threshold in `cr-local-review.md` — so it was stopped and the CLI dropped for this session. The new files therefore have **not** had a local CLI pass; the CodeRabbit and CodeAnt **GitHub Apps** review them on this PR and are what gates the merge.
- Of round 1's 6 findings: 3 accepted (ceiling described as a backstop rather than a replacement poll; re-arm recovery in the pre-exit checklist; the bounded stand-down stated explicitly in the failure-mode doc). 3 declined with reasoning — moving `session-start-sync.sh` to a `SessionStart` segment (it really is registered under `PostToolUse`); requiring arm-*before*-spawn (the design is spawn-then-arm within the same step, and there is nothing to arm against beforehand); and collapsing the `monitor-mode.md` / `subagent-orchestration.md` one-liners into references (AC7 requires the step to appear in both).

## Rebased onto PR #804

PR #804 (rule-corpus compression, 12,166 → 10,999 words + ratchet reset) landed mid-flight and touched the same three rule files. Rebased with both conflicts hand-merged rather than resolved ours/theirs (`feedback_merge_conflict_trace_both_sides.md`): main's compressed wording is kept everywhere, with the ceiling additions layered on top. Two trims I had made *purely* to fit the old budget — the "Why the Never above" line in `scheduling-reliability.md` and the `File-Write Status Updates` text in `monitor-mode.md` — are **reverted**, since the compression removed the pressure that justified them. The diff is now the feature and nothing else.

## Follow-up spotted, not fixed here

`scheduling-reliability.md` still describes `CronCreate` as the primitive for "cross-session durability", but the tool's own contract now states `durable` has no effect and jobs are session-only. Out of scope for this change; worth its own issue.

### DIFF
--- a/.claude/hooks/README.md
+++ b/.claude/hooks/README.md
@@ -111,6 +111,24 @@ Replace `/absolute/path/to/claude-code-config` with the actual path to your clon
 
 - All hook scripts must be executable: `chmod +x .claude/hooks/silence-detector*.sh`
 
+## bgwork-ceiling-arm.sh + bgwork-ceiling-guard.sh
+
+Puts a hard ceiling on chat silence while background work runs (issue #803). The pair above enforces the 5-minute heartbeat *within* a turn; this pair covers what happens *after* one ends — a thread that spawns a subagent, ends its turn, and waits makes no tool calls, so no PostToolUse hook can ever fire for it.
+
+**How it works:** the measurement is not re-implemented. `/tmp/claude-heartbeat-$SESSION_ID` already means "last user-visible message"; what these hooks add is an observer of it that can force a turn, plus a gate making that observer non-optional.
+
+- **`bgwork-ceiling-arm.sh`** (PostToolUse, all tools): records background work — an `Agent` spawn, a `Workflow`, a `Monitor`, or a `Bash` call with `run_in_background: true` — and, while the ceiling is unarmed, injects the exact arming call into context. Arming is recognised by the `--tick` sentinel in the command text, so the ceiling watch is never miscounted as new background work. **Advisory only.**
+- **`bgwork-ceiling-guard.sh`** (Stop): returns `decision: block` when background work is in flight and the ceiling is unarmed, so the turn cannot end silently. **This is the enforcement.** Blocking is bounded at 2 consecutive turns (`CLAUDE_BGWORK_MAX_BLOCKS`); past the bound it stands down loudly — stderr, `~/.claude/logs/bgwork-ceiling.log`, and an unguarded marker the arm hook resurfaces on every later tool call.
+
+The armed watch is a `Monitor` running `bgwork-ceiling.sh --tick` on a loop. It prints nothing unless the heartbeat file is genuinely stale, so a thread sending normal heartbeats never sees a message from it, and `persistent: true` means the ceiling is armed **once per session** rather than once per spawn.
+
+All state lives in `/tmp` markers (overridable as a set via `CLAUDE_BGWORK_MARKER_DIR`, which is what the test suites use) and in `.claude/scripts/bgwork-ceiling.sh`, which owns the ceiling duration — no rule file states it. Rationale, the `Monitor`-vs-`ScheduleWakeup`-vs-`CronCreate` call, and the derived-timing invariant: `.claude/reference/bgwork-ceiling.md`.
+
+### Prerequisites
+
+- `jq` must be installed
+- `.claude/scripts/bgwork-ceiling.sh` must be present and executable — both hooks exit 0 silently without it, so a checkout missing it degrades rather than breaks
+
 ## skill-usage-tracker.sh + skill-command-tracker.sh
 
 Records every skill invocation to `~/.claude/skill-usage.log` (and the aggregate `~/.claude/skill-usage.csv`), which `.claude/scripts/skill-usage-report.sh` rolls up.
--- a/.claude/hooks/bgwork-ceiling-arm.sh
+++ b/.claude/hooks/bgwork-ceiling-arm.sh
@@ -1,0 +1,135 @@
+#!/bin/bash
+# Background-work ceiling — arming half (PostToolUse hook, all tools).
+#
+# Two jobs, both driven off the tool call that just ran:
+#
+#   1. Notice that the thread started background work it will wait on — a
+#      subagent spawn, a backgrounded Bash command, a Workflow, or a Monitor —
+#      and record it via bgwork-ceiling.sh --note-started.
+#   2. While background work is in flight and the ceiling watch is NOT armed,
+#      inject the exact arming call into the agent's context, so arming happens
+#      as part of the same step rather than as a separate thing to remember.
+#
+# Arming itself is recognised by the `--tick` sentinel in the command text, so
+# the ceiling watch is never miscounted as new background work (which would
+# leave the guard permanently unsatisfiable).
+#
+# The companion Stop hook (bgwork-ceiling-guard.sh) is what makes this
+# non-optional: this hook only advises, it cannot stop a turn from ending.
+
+STDIN_JSON=$(cat)
+
+emit_context() {
+  jq -n --arg message "$1" '{
+    hookSpecificOutput: {
+      hookEventName: "PostToolUse",
+      additionalContext: $message
+    }
+  }'
+}
+
+json_field() {
+  printf '%s' "$STDIN_JSON" | jq -r "$1 // empty" 2>/dev/null
+}
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+CEILING_SH="${SCRIPT_DIR%/hooks}/scripts/bgwork-ceiling.sh"
+[[ -x "$CEILING_SH" ]] || exit 0
+
+SESSION_ID=$(json_field '.session_id')
+SESSION_ID="${SESSION_ID:-${CLAUDE_SESSION_ID:-default}}"
+
+TOOL_NAME=$(json_field '.tool_name')
+COMMAND=$(json_field '.tool_input.command')
+BACKGROUNDED=$(json_field '.tool_input.run_in_background')
+
+# --- 1. Is this call arming the ceiling watch? -------------------------------
+# Matched on the command text rather than the tool name, so arming still
+# registers if the watch is started some other way (e.g. a backgrounded Bash).
+#
+# The match is against the WHOLE generated command, not just a `--tick`
+# substring: in this repo plenty of commands merely mention the sentinel (a
+# grep, a docs edit, a test run), and treating one of those as "armed" would
+# satisfy the Stop hook with no watch actually running — a guard defeated by
+# talking about it. Regenerating the expected command here also keeps detection
+# in step with --arm-command automatically if its shape ever changes.
+if [[ "$COMMAND" == *bgwork-ceiling.sh* ]]; then
+  EXPECTED_ARM=$("$CEILING_SH" --arm-command --session "$SESSION_ID" 2>/dev/null)
+  if [[ -n "$EXPECTED_ARM" && "$COMMAND" == *"$EXPECTED_ARM"* ]]; then
+    "$CEILING_SH" --record-armed --session "$SESSION_ID" || true
+    echo '{}'
+    exit 0
+  fi
+fi
+
+# --- 2. Did this call start background work the thread will wait on? --------
+started=0
+case "$TOOL_NAME" in
+  Agent|Workflow|Monitor)
+    started=1
+    ;;
+  Bash)
+    [[ "$BACKGROUNDED" == "true" ]] && started=1
+    ;;
+esac
+
+if (( started == 1 )); then
+  "$CEILING_SH" --note-started "$TOOL_NAME" --session "$SESSION_ID" || true
+fi
+
+# --- 3. Advise arming while in flight and unarmed ----------------------------
+if "$CEILING_SH" --check --session "$SESSION_ID"; then
+  echo '{}'
+  exit 0
+fi
+
+ARM_COMMAND=$("$CEILING_SH" --arm-command --session "$SESSION_ID" 2>/dev/null)
+if [[ -z "$ARM_COMMAND" ]]; then
+  echo '{}'
+  exit 0
+fi
+
+MARKER_DIR="${CLAUDE_BGWORK_MARKER_DIR:-/tmp}"
+SAFE_ID="${SESSION_ID//[^[:alnum:]_.-]/_}"
+SAFE_ID="${SAFE_ID:-default}"
+
+UNGUARDED_NOTE=""
+if [[ -f "${MARKER_DIR}/claude-bgceiling-unguarded-${SAFE_ID}" ]]; then
+  UNGUARDED_NOTE="This thread is currently UNGUARDED — the Stop hook already blocked twice and stood down. Arm it now. "
+fi
+
+# The call that STARTED the work always gets the advisory — that is the step
+# arming belongs in. Every later call while still unarmed is a reminder, and
+# reminders are rate-limited so a burst of tool calls cannot replay this whole
+# paragraph into context on each one. The Stop hook, not repetition, is what
+# makes arming actually happen.
+#
+# The rate limit deliberately does NOT apply once the guard has stood down: at
+# that point the thread really is unguarded, and quietly throttling the only
+# remaining signal would be the same silent degradation this whole change
+# exists to remove (feedback_guard_must_fail_closed.md).
+ADVISED_FILE="${MARKER_DIR}/claude-bgceiling-advised-${SAFE_ID}"
+COOLDOWN_S="${CLAUDE_BGWORK_ADVISE_COOLDOWN_S:-60}"
+[[ "$COOLDOWN_S" =~ ^[0-9]+$ ]] || COOLDOWN_S=60
+
+if (( started == 0 )) && [[ -z "$UNGUARDED_NOTE" ]] && [[ -f "$ADVISED_FILE" ]]; then
+  if [[ "$(uname -s)" == "Darwin" ]]; then
+    last_advice=$(stat -f %m "$ADVISED_FILE" 2>/dev/null)
+  else
+    last_advice=$(stat -c %Y "$ADVISED_FILE" 2>/dev/null)
+  fi
+  if [[ -n "$last_advice" ]]; then
+    advice_age=$(( $(date +%s) - last_advice ))
+    # A negative age means a future-dated marker (clock rollback) — treat it as
+    # stale and re-advise rather than suppressing indefinitely.
+    if (( advice_age >= 0 && advice_age < COOLDOWN_S )); then
+      echo '{}'
+      exit 0
+    fi
+  fi
+fi
+touch "$ADVISED_FILE" 2>/dev/null || true
+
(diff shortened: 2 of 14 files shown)


auerbachb:
@graphite-app re-review

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

PR #817, opened 2026-07-30
fix(#810): fire the capability ladder on deferral, not on "impossible"

### LINKED ISSUE #810: Capability ladder doesn't fire when work is deferred as out of scope
## Background

When a merged change needs a configuration step before it actually does anything — generate keys, set env vars, flip a per-service flag — the expectation is that the agent just does it. That is what the capability ladder in `.claude/rules/safety.md` exists for (issue #759): before saying "I can't," look for the CLI, install it if that's non-interactive, and only then hand back a runbook. The rule loads globally through the `~/.claude` symlinks, so it is in context in every repo, not just this one.

## Problem

The ladder only catches one of the two ways an agent walks away from work.

It fires on **capability** claims — *"I can't do that."* It does not fire on **scope** claims — *"that's not a session task,"* *"that's a deployment step,"* *"the runbook is in `docs/…`."* In the motivating case the agent noticed the merge changed no production behavior, identified the three env vars still unset, wrote a correct runbook, and handed the whole thing over. It never claimed the work was impossible — it treated the work as somebody else's. The ladder was loaded and simply never triggered.

Two things make that easy to fall into:

- **The last rung reads like permission.** Rung 4 is *"hand off a runbook, not a description."* Nothing in the loaded rule says you may only *reach* rung 4 by failing rungs 1–3, so writing a good runbook looks like compliance rather than the fallback it is meant to be. The sentence that would have caught this — *"If you can write the command, you can run it"* — lives in `.claude/reference/capability-discovery-examples.md`, which `CLAUDE.md` deliberately excludes from auto-loading. The guardrail loads; its counterweight does not.
- **Secrets work looks pre-refused.** The deferred task was generating and provisioning signing keys. `safety.md` is emphatic — correctly — about never committing or logging credentials, and that is easy to read as "don't touch keys at all" rather than "set them through the provider CLI with the value on stdin, never echoed." A rule read as a blanket prohibition gives the scope deferral cover it should not have.

The net effect is the failure mode the whole setup exists to prevent: a security fix merged, CI green, session reported complete — and production still on the old path, waiting on three variables the agent had the access to set.

## Proposed solution

Make the ladder fire on the *deferral*, not on the word "impossible."

Any time the agent is about to leave work undone — it can't, it's out of scope, it's a deployment step, or it's just pointing at a runbook — that is the trigger. The question it must answer first is "could I do this with the tools I already have?", and only a genuine dead-end lets it hand the work back.

Rung 4 should stop reading as a destination in its own right. Reaching it means having actually walked rungs 1–3 and saying which one failed and why — the ladder's existing "name the rung" discipline, applied to the runbook case and not only to the "I can't" case.

The credential boundary should also be stated plainly where it is actually read, so it can't be borrowed as an excuse: generating a key and provisioning it through a provider CLI is on the allowed side of the line. Committing, echoing, or logging the value is what isn't.

## Acceptance Criteria

- [ ] The ladder's entry condition in `.claude/rules/safety.md` covers scope-shaped deferrals, not just impossibility claims — it names the shapes explicitly ("not a session task", "that's a deployment step", "runbook is in …")
- [ ] Rung 4 states its own precondition: reachable only after rungs 1–3 have been walked, and the handoff names the rung that failed
- [ ] The `MINDSET:` block in `safety.md` carries both changes — it is the only form subagents receive (`subagent-orchestration.md` item 6)
- [ ] `safety.md` disambiguates the credential rule: provisioning a generated secret via a provider CLI (value on stdin, never echoed) is permitted; the prohibition is on committing, pasting, or logging values
- [ ] Rule corpus stays under the ratchet cap — currently **11,171** words against `.claude/rules/.budget-soft-cap` = **11,749**; prefer net-neutral by trimming as you add
- [ ] `.claude/reference/capability-discovery-examples.md` gains the scope-deferral shape to its false-walls catalog, with this case as the worked example

## Test Plan

- [ ] `{ cat CLAUDE.md; find .claude/rules -name '*.md' -exec cat {} +; } | wc -w` reports under 11,749
- [ ] `.github/scripts/rule-lint.sh` exits 0
- [ ] `.github/scripts/verbatim-block-lint.sh` exits 0 — the `MINDSET:` copy in `.claude/skills/subagent/SKILL.md` matches the canonical block in `safety.md`
- [ ] Read-back check against the motivating transcript: given *"merging changed no behaviour — three env vars are absent in production, runbook in `docs/security/S2S-IDENTITY.md`"*, the revised text unambiguously requires walking the ladder before that sentence may be written
- [ ] Read-back check on the credential line: an agent reading only `safety.md` can tell that `railway variables --set` with a generated key is allowed and that echoing the key into a PR body is not

## Notes / Open questions

- Parent issue #759 built the ladder for the capability framing; this is the scope framing that slipped past it. Worth considering whether a third framing exists ("I'll leave that to you in case you want to review it first") that also needs naming.
- Deliberately filed narrow on the ladder. The adjacent idea — a post-merge dormancy gate in `/wrap` that flags "this merged but changes no behavior until X is configured" — is a separate surface, and in the motivating case the *detection* already worked; only the action was missing. Say the word and it gets its own issue.
- The corpus is word-budget constrained, so the fix is most likely a rewrite of the ladder's opening line and rung 4 rather than added paragraphs.

## Related Issues

- Related to #759

_Captured via /issue-maker._

## Implementation Plan

_Merged: Claude's analysis + CodeRabbit's plan. CR's Phases 1–2 adopted as-is; two deltas resolved below._

### Files to modify

| File | Budgeted? | Change |
|------|-----------|--------|
| `.claude/rules/safety.md` | **Yes** (corpus) | Ladder entry condition, rung 3→4 gate relocation, rung 4 precondition, credential disambiguation, `MINDSET:` block |
| `.claude/skills/subagent/SKILL.md` | No | Re-sync verbatim `MINDSET:` copy |
| `.claude/skills/pr-monitor-and-manage/SKILL.md` | No | Re-sync verbatim `MINDSET:` copy |
| `.claude/reference/capability-discovery-examples.md` | No | Scope-deferral entry in the **false-walls** catalog + worked example |
| `.claude/agents/{phase-a-fixer,phase-b-reviewer,phase-c-merger}.md` | No | Reframe the one-line ladder reminder off "impossible" |
| `.claude/agents/README.md` | No | Reframe the paraphrased `MINDSET:` in the example spawn prompt |

### Steps

1. **`safety.md` — ladder entry condition (AC1).** Replace `Before declaring any task impossible, walk this ladder…` so the trigger is the *deferral*, not the word "impossible". Name the shapes explicitly: `"I can't"`, `"not a session task"`, `"that's a deployment step"`, `"the runbook is in docs/…"`, and the review-deferral framing (`"I'll leave that to you in case you want to review it first"`). Governing question stated up front: *could I do this with the tools I already have?*
2. **`safety.md` — rung 4 precondition (AC2).** Rung 4 becomes reachable only after rungs 1–3 were actually walked and failed, and the handoff must name the failing rung and why. Relocate the gate currently buried in rung 3 ("Any rail that blocks … drops to rung 4") down to rung 4 so it reads as one condition. Carry the counterweight sentence — *if you can write the command, you can run it* — into the loaded rule, since `capability-discovery-examples.md` is deliberately not auto-loaded.
3. **`safety.md` — credential disambiguation (AC4).** In `## Secrets & Credentials`, add a bolded clarifier: generating a secret and provisioning it through a provider CLI (`railway variables --set`, `vercel env add`, value on stdin, never echoed) is ordinary work; committing, pasting, echoing, or logging the value is what's prohibited. Existing absolute NEVERs stay intact — this narrows the reading, it does not weaken the ban.
4. **`safety.md` — `MINDSET:` block (AC3).** Carry the deferral trigger + rung-4 precondition into the fenced block, plus a one-clause credential note (subagents receive only `SAFETY:` + `MINDSET:`, so the prose clarifier never reaches them). Drop the now-redundant trailing "valid only after the ladder dead-ends" sentence — rung 4 states it.
5. **Re-sync the two verbatim copies.** `.claude/skills/subagent/SKILL.md` and `.claude/skills/pr-monitor-and-manage/SKILL.md` must be byte-identical; `verbatim-block-lint.sh` diffs both.
6. **`capability-discovery-examples.md` (AC6).** Add the scope-deferral shape to the **false-walls** catalog with a shape→reality table (including the review-deferral framing), the motivating case as the worked example, and a "secrets are not pre-refused" note. Extend `## Anti-pattern` to name the scope framing alongside the capability framing.
7. **Agent-definition paraphrases.** `phase-a-fixer.md`, `phase-b-reviewer.md`, `phase-c-merger.md` all open with "Before treating something as impossible, walk the capability ladder…" — the exact framing this issue exists to fix. `agents/README.md`'s example spawn prompt carries a paraphrased `MINDSET:` with the same opening. Reframe all four onto the deferral trigger.
8. **Budget + lints.** Re-measure the corpus, then run `rule-lint.sh`, `verbatim-block-lint.sh`, `chip-model-guard-lint.sh`, `merge-authority-lint.sh`.

### Word budget

Baseline **11,171** against cap **11,749** (578 headroom). Only `safety.md` is budgeted. Additions are funded by trimming within the same section: rung 2's "a research detour mid-task is not warranted", rung 1's PATH clause, rung 3's relocated gate sentence, and the closing sentence made redundant by the new rung 4. Target: net increase under ~60 words. Re-measure after any rebase (main can grow underneath — prior incident: word-count drift after rebase).

### Deltas from CodeRabbit's plan

- **CR Phase 3 placement — rejected.** CR would append the worked example inside `## Real walls (handoff is correct, but structured)`. AC6 says *false-walls catalog*, and a scope deferral is by definition a **false** wall — the work was doable the whole time. Filing it under "Real walls" would teach the opposite lesson. Using the false-walls catalog instead.
- **Agent definitions — CR missed them.** Three agent definitions and `agents/README.md` restate the ladder with the "impossible"-only trigger. These load as system context for every subagent, so leaving them would preserve the bug on the exact surface `MINDSET:` exists to protect. Added as step 7.
- **CR Design Choice 1 — accepted.** The third framing ("I'll leave that to you in case you want to review it first") goes inline in the scope-shape list rather than deferring to a follow-up issue.

### Risks / edge cases

- **Verbatim drift** is the main CI risk: three copies of `MINDSET:` (canonical + 2 skills) must match byte-for-byte, including line wrapping and the trailing `Full rules:` line.
- **Semantic deletion while trimming** — a prior incident had compression silently drop an inline qualifier that all four lints missed. Every trim here must be verified as redundant-with-an-adjacent-sentence, not just shorter.
- **Weakening a safety rule** — the credential clarifier must read as a narrowing of interpretation, never as a new permission. The absolute NEVERs stay verbatim.
- **Rebase drift** on the word count if main lands other corpus PRs first.

### Verification

- AC1/AC2/AC4: read-back against the motivating sentence ("merging changed no behaviour — three env vars are absent in production, runbook in `docs/security/S2S-IDENTITY.md`") and the credential case (`railway variables --set` allowed / echoing into a PR body not).
- AC3: `verbatim-block-lint.sh` exit 0 across both copy surfaces.
- AC5: `{ cat CLAUDE.md; find .claude/rules -name '*.md' -exec cat {} +; } | wc -w` under 11,749; `rule-lint.sh` exit 0.
- AC6: grep the reference for the scope-deferral shapes and the worked example.

auerbachb:
## Summary

The capability ladder in `.claude/rules/safety.md` fired on **capability** claims ("I can't") but not on **scope** claims ("that's a deployment step", "the runbook is in `docs/…`"). An agent that reassigns work rather than declaring it impossible never triggered it — the guardrail was loaded and simply stayed silent.

This makes the **deferral** the trigger, not the word "impossible", and closes the two things that made the gap easy to fall into:

- **Rung 4 read like permission.** "Hand off a runbook" was stated as a peer of rungs 1–3, so writing a good runbook looked like compliance rather than the fallback it is. It now states its own precondition — reachable only after rungs 1–3 were walked and *actually failed*, naming the rung that stopped you. The counterweight sentence that would have caught the motivating case, *"if you can write the command, you can run it"*, moves into the loaded rule; it previously lived only in `capability-discovery-examples.md`, which `CLAUDE.md` deliberately excludes from auto-loading.
- **Secrets work looked pre-refused.** `safety.md`'s credential prohibitions read at a glance as "don't touch keys at all", which gave scope deferral cover it shouldn't have. A new **Provisioning is not committing** clarifier draws the line where it actually falls: setting a generated secret through a provider CLI is ordinary work; committing, pasting, echoing, or logging the value is not.

Closes #810

## Changes

| File | Change |
|------|--------|
| `.claude/rules/safety.md` | Ladder entry condition names the deferral shapes; rung 4 gains its precondition; credential clarifier; `MINDSET:` block carries all of it |
| `.claude/skills/subagent/SKILL.md` | Verbatim `MINDSET:` re-sync |
| `.claude/skills/pr-monitor-and-manage/SKILL.md` | Verbatim `MINDSET:` re-sync |
| `.claude/reference/capability-discovery-examples.md` | Scope-deferral shapes added to the false-walls catalog + the motivating case as a worked example |
| `.claude/agents/{phase-a-fixer,phase-b-reviewer,phase-c-merger}.md` | Reframed off "before treating something as impossible" |
| `.claude/agents/README.md` | Paraphrased `MINDSET:` in the example spawn prompt reframed to match |

**Scope note — the four agent definitions were not in the issue's AC list.** They are included because all four restated the ladder with the "impossible"-only trigger, and they load as system context for every subagent. Leaving them would have preserved the exact bug on the surface `MINDSET:` exists to protect.

## Word budget

`11,171 → 11,356` against the `.budget-soft-cap` of **11,749** (393 headroom). Additions were partly funded by trimming within the same section — rung 2's "a research detour mid-task is not warranted", rung 1's PATH clause, rung 3's relocated rung-4 gate, and the closing sentence made redundant by the new rung 4. Net positive rather than net-neutral because the change adds a genuinely new trigger class, a precondition, and a credential boundary; no existing qualifier was dropped to fund it.

## Divergences from CodeRabbit's issue plan

- **Worked-example placement.** CR proposed appending it inside `## Real walls (handoff is correct, but structured)`. AC6 says *false-walls catalog*, and a scope deferral is a false wall by definition — the work was doable the whole time. Filing it under "Real walls" would teach the opposite lesson, so it went to the false-walls catalog.
- **Agent definitions.** Not in CR's plan; added for the reason above.
- **CR Design Choice 1 — accepted.** The third framing from the issue's open questions ("I'll leave that to you in case you want to review it first") is named inline rather than deferred to a follow-up issue, and the reference file explains why courtesy is the hardest of the shapes to catch.

## Local review

- **CodeRabbit CLI:** two passes. First pass returned 4 findings; 3 were applied and 1 declined:
  - *Applied* — the `Always:` summary said secrets were "never-committed or -pasted" while the body and `MINDSET:` say "echoed, committed, pasted, or logged"; the summary now matches.
  - *Applied* — the reference's "reversible and inspectable" test could be read as licence to mutate any production environment. It now requires the target to be **already established by the task**; standing up a new environment, or writing to one the task never named, is an explicit rung-4 handoff.
  - *Applied* — `phase-c-merger.md`: the strengthened "before leaving any work undone" wording conflicted with Phase C's no-fix contract. The line now states that it never overrides `OUTCOME: blocked`.
  - *Declined* — replacing `/opt/homebrew/bin/<tool>` with `command -v` / `type -P` in `agents/README.md`. This would reverse a deliberate, documented decision: the Bash tool runs with a minimal PATH, so bare `which`/`command -v` under-reports installed tools. `safety.md` rung 1, the `MINDSET:` block, `cli-tool-defaults.md`, and `capability-discovery-examples.md` (which shows `ls -l /opt/homebrew/bin/railway || command -v railway` — absolute path first, `command -v` as fallback) all encode this. It also reproduced live in this session: `command -v codeant` found nothing useful, and `coderabbit` was only located by an explicit path sweep to `~/.local/bin`.
- **CodeAnt CLI:** dropped for this session — not installed. Ladder walked per `safety.md`: rung 1 absent from every install path; rung 2 confirms it ships one (`npm install -g codeant-cli`); rung 3 blocked because auth is `codeant login` browser OAuth with no `~/.codeant/config.json` present, which no agent can drive. **Stopped at rung 3 (interactive auth).** The CodeAnt GitHub App is unaffected and still gates the merge.

## Incidental fix — `verbatim-block-lint.test.sh`

The lint's own negative test injected drift with `sed 's/capability ladder/DRIFTED ladder/'`. This rewrite removes that literal from the `MINDSET:` block, so the sed became a no-op, no drift was injected, and the test that is supposed to prove the lint *catches* drift passed for the wrong reason. Retargeted to `ANY provider`, plus a uniqueness guard that fails loudly if a future reword breaks the assumption again. Verified the guard fires: substituting the old token back reproduces the failure with an explicit message instead of a silent pass.

## Test plan

- [x] `{ cat CLAUDE.md; find .claude/rules -name '*.md' -exec cat {} +; } | wc -w` reports under 11,749
- [x] `.github/scripts/rule-lint.sh` exits 0
- [x] `.github/scripts/verbatim-block-lint.sh` exits 0 — the `MINDSET:` copies in `.claude/skills/subagent/SKILL.md` and `.claude/skills/pr-monitor-and-manage/SKILL.md` match the canonical block in `safety.md`
- [x] Read-back check against the motivating transcript: given *"merging changed no behaviour — three env vars are absent in production, runbook in `docs/security/S2S-IDENTITY.md`"*, the revised text unambiguously requires walking the ladder before that sentence may be written
- [x] Read-back check on the credential line: an agent reading only `safety.md` can tell that `railway variables --set` with a generated key is allowed and that echoing the key into a PR body is not
- [x] Ladder entry condition names the scope-shaped deferrals explicitly (AC1)
- [x] Rung 4 states its own precondition and requires naming the failed rung (AC2)
- [x] `.claude/reference/capability-discovery-examples.md` carries the scope-deferral shape in its false-walls catalog with the motivating case worked through (AC6)

### DIFF
--- a/.claude/agents/README.md
+++ b/.claude/agents/README.md
@@ -91,15 +91,20 @@ Agent tool call:
            commands (rm -rf, rm, git checkout ., git stash, git reset --hard) in the
            root repo directory. Stay in your worktree directory at all times.
 
-           MINDSET: Before handing off, walk the capability ladder for ANY provider
-           CLI (gh, git, vercel, neonctl, railway, cloudinary, or one you've never
-           used): look at what you already have — MCP tools, skills, and the CLI on
-           disk by absolute path (/opt/homebrew/bin/<tool>); if absent, check
-           whether the provider ships a CLI; install it yourself when
+           MINDSET: The trigger is the DEFERRAL, not the word 'impossible' —
+           'I can't', 'not a session task', 'that's a deployment step', 'runbook is
+           in docs/…' all fire this ladder. Walk it for ANY provider CLI (gh, git,
+           vercel, neonctl, railway, cloudinary, or one you've never used) before
+           writing any of them: look at what you already have — MCP tools, skills,
+           and the CLI on disk by absolute path (/opt/homebrew/bin/<tool>); if
+           absent, check whether the provider ships a CLI; install it yourself when
            non-interactive and the safety rails hold; else hand off an
-           /admin-merge-shaped runbook — exact commands + one-line reason. 'I can't'
-           is valid only after the ladder dead-ends, and must name the rung and the
-           reason.
+           /admin-merge-shaped runbook — reachable only after the first three rungs
+           were walked and failed. It must name the rung that stopped you and the
+           reason, and give the exact commands, including the interactive auth step
+           when that is the wall. If you can write the command, you can run it.
+           Provisioning a generated secret via a provider CLI is allowed — the
+           value just must never be echoed, committed, pasted, or logged.
 
            SKILLS: Before hand-rolling a multi-step task, check whether an existing
            skill already does this job — invoke it via the Skill tool instead of
--- a/.claude/agents/phase-a-fixer.md
+++ b/.claude/agents/phase-a-fixer.md
@@ -42,7 +42,7 @@ The parent agent provides these values in your prompt:
 - NEVER run destructive commands (`rm -rf`, `rm`, `git checkout .`, `git stash`, `git reset --hard`) in the root repo directory.
 - Stay in your worktree directory at all times.
 - NEVER add `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, `noqa`, or any linter suppression comment. Fix the actual code.
-- Before treating something as impossible, walk the capability ladder for any provider CLI (`gh`, `git`, `curl`, or a service CLI like `railway`/`vercel`) — check locally by absolute path, check whether the provider ships one, install it when safe; only hand off for real walls, naming the rung you stopped on, per the capability-discovery mindset in `safety.md`.
+- Before leaving any work undone — whether you'd frame it as impossible, out of scope, a deployment step, or a runbook for someone else — walk the capability ladder for any provider CLI (`gh`, `git`, `curl`, or a service CLI like `railway`/`vercel`): check locally by absolute path, check whether the provider ships one, install it when safe. Handing off is rung 4, reachable only after rungs 1–3 actually failed: name the rung that stopped you and why, and give the exact commands, including the interactive auth step when that is the wall — per the capability-discovery mindset in `safety.md`.
 
 ## Workflow
 
--- a/.claude/agents/phase-b-reviewer.md
+++ b/.claude/agents/phase-b-reviewer.md
@@ -23,7 +23,7 @@ The parent agent provides:
 - NEVER run destructive commands (`rm -rf`, `rm`, `git checkout .`, `git stash`, `git reset --hard`) in the root repo directory.
 - Stay in your worktree directory at all times.
 - NEVER add linter suppression comments. Fix the actual code.
-- Before treating something as impossible, walk the capability ladder for any provider CLI (`gh`, `git`, `curl`, or a service CLI like `railway`/`vercel`) — check locally by absolute path, check whether the provider ships one, install it when safe; only hand off for real walls, naming the rung you stopped on, per the capability-discovery mindset in `safety.md`.
+- Before leaving any work undone — whether you'd frame it as impossible, out of scope, a deployment step, or a runbook for someone else — walk the capability ladder for any provider CLI (`gh`, `git`, `curl`, or a service CLI like `railway`/`vercel`): check locally by absolute path, check whether the provider ships one, install it when safe. Handing off is rung 4, reachable only after rungs 1–3 actually failed: name the rung that stopped you and why, and give the exact commands, including the interactive auth step when that is the wall — per the capability-discovery mindset in `safety.md`.
 
 ## Initialization
 
--- a/.claude/agents/phase-c-merger.md
+++ b/.claude/agents/phase-c-merger.md
@@ -25,7 +25,7 @@ The parent agent provides:
 - Stay in your worktree directory at all times except for `/wrap` helper calls that explicitly target the resolved root repo path.
 - Do not run `gh pr merge` directly. After verification, execute the shared `/wrap` instructions from `.claude/skills/wrap/SKILL.md` so Phase C and `/wrap` cannot drift.
 - Do not delete the running worktree or feature branch. `/wrap` intentionally leaves them in place; stale cleanup is owned by `/pm-update` via `.claude/scripts/stale-cleanup.sh`.
-- Before treating something as impossible, walk the capability ladder for any provider CLI (`gh`, `git`, `curl`, or a service CLI like `railway`/`vercel`) — check locally by absolute path, check whether the provider ships one, install it when safe; only hand off for real walls, naming the rung you stopped on, per the capability-discovery mindset in `safety.md`.
+- Before leaving any work undone — whether you'd frame it as impossible, out of scope, a deployment step, or a runbook for someone else — walk the capability ladder for any provider CLI (`gh`, `git`, `curl`, or a service CLI like `railway`/`vercel`): check locally by absolute path, check whether the provider ships one, install it when safe. Handing off is rung 4, reachable only after rungs 1–3 actually failed: name the rung that stopped you and why, and give the exact commands, including the interactive auth step when that is the wall — per the capability-discovery mindset in `safety.md`. This never overrides your no-fix contract: a code issue found during AC verification is still `OUTCOME: blocked`, not yours to fix.
 
 ## Initialization
 
--- a/.claude/reference/capability-discovery-examples.md
+++ b/.claude/reference/capability-discovery-examples.md
@@ -1,6 +1,6 @@
 # Capability Discovery — False Walls vs Real Walls
 
-Extracted from `.claude/rules/safety.md` "Capability Discovery — Try the CLI Before Handoff". The rule file keeps the operative ladder and the verbatim `MINDSET:` block; this file holds the worked catalog. **Extend the false-walls list as new ones are observed.**
+Extracted from `.claude/rules/safety.md` "Capability Discovery — Try the CLI Before Deferring". The rule file keeps the operative ladder and the verbatim `MINDSET:` block; this file holds the worked catalog. **Extend the false-walls list as new ones are observed.**
 
 ## Common false walls — GitHub (usually doable)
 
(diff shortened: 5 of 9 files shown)


auerbachb:
@graphite-app re-review

auerbachb:
Fixed in `0e22364`: the guard's `grep -o` now fails closed to `occurrences=0` instead of aborting the run.

You're right that this was self-defeating — under `set -euo pipefail` a missing token made `grep` exit 1, the pipeline fail, and the script abort *before* the guard could report the very condition it exists to catch. Verified both directions: with an absent token the guard now prints `occurs 0x` and the remaining cases still run; with the real token the suite passes 4/4.

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

PR #820, opened 2026-07-30
fix(#794): reconcile stable-state backoff ladder across rule, skill, and hook

### LINKED ISSUE #794: Backoff tier contradiction: scheduling-reliability.md says 5m at streak>=3, babysit-pr says 15m
## Background

Hit live while running `/babysit-pr 775 --tick` (PR #775). At `digest_streak == 3` the watcher had to widen its cadence, and two rules gave different answers.

## Problem

Two contracts disagree about the stable-state backoff ladder:

- **`.claude/rules/scheduling-reliability.md`** — "Widen at streak ≥3→5m, ≥6→15m; `CronDelete` at ≥9". The `polling-backoff-warn.sh` hook enforces this, emitting `Call CronUpdate to widen the interval to 5m` at streak 3.
- **`.claude/skills/babysit-pr/SKILL.md` (T5)** — `>= 3` → `max(15m, 3 × base)`, i.e. **15m** at streak 3 for the default 5m base.

For the default `--cadence 5m`, the rule's "widen to 5m" is a **no-op** — it tells you to widen to the interval you are already on. The skill's 15m is the only instruction that actually widens.

I resolved it live by taking 15m (satisfies the hook's floor and matches the skill), but an agent following the hook literally would widen to 5m, change nothing, and re-trip the same warning on the next tick — a loop that looks like backoff without being it.

This is exactly the class of defect `/harness-audit` was built to catch — two sources describing the same behavior differently — except here both sources are ours, so the audit's harness-vs-us lens does not see it.

## Proposed solution

Pick one ladder and make both sources agree. Options:

1. **Skill wins** — change `scheduling-reliability.md` (and the hook's message) to `≥3 → max(15m, 3×base)`, `≥6 → wider`, `≥9 → stop`.
2. **Rule wins** — change `babysit-pr` T5 to the 5m/15m ladder, accepting that a 5m-base watcher does not widen until streak 6.

Option 1 looks right: the rule's absolute minutes assume a 1m base and silently no-op at the documented 5m default, whereas the skill's base-relative formula is correct at every base.

Whichever is chosen, `polling-backoff-warn.sh`'s emitted message must match, since that string is what an agent acts on.

## Acceptance Criteria

- [ ] `scheduling-reliability.md` and `babysit-pr/SKILL.md` state the same tier ladder
- [ ] `polling-backoff-warn.sh`'s warning text names the same interval the rules do
- [ ] The ladder is a no-op for no supported base cadence (widening always widens)
- [ ] A note records which source was authoritative and why

## Test Plan

- [ ] Simulate `digest_streak = 3` with base 5m; confirm rule, skill, and hook all say the same interval
- [ ] Same at base 1m and base 20m — confirm the widened value is always strictly greater than the base
- [ ] `bash .github/scripts/rule-lint.sh` passes
- [ ] `bash .github/scripts/run-hook-tests.sh` green

## Notes / Open questions

- The hook also reads `last_cron_action.type` / `.interval`; a related sharp edge is that those field names are not documented anywhere near the rule that requires them (I wrote `action`/`to_minutes` first and the hook kept firing, correctly). Worth documenting the expected schema alongside the ladder.
- Word budget: `scheduling-reliability.md` edits land in the auto-loaded corpus, currently pinned near its soft cap — see issue #790.

_Filed via /wrap._

auerbachb:
## Summary

- `scheduling-reliability.md` said "widen to 5m at streak≥3, 15m at streak≥6" while `babysit-pr/SKILL.md` T5 used `max(15m, 3×base)` — contradicting each other
- At the default 5m base, "widen to 5m" is a no-op: the hook kept emitting every tick forever (the flooding bug)
- This PR makes the base-relative formula from the skill authoritative across all three surfaces

## Changes

- **`scheduling-reliability.md`** — Stable-State Backoff paragraph: `≥3→5m, ≥6→15m` replaced with `≥3 to max(15m, 3×base)`; added note that the hook reads `babysit.cadence_base_minutes` (default 5)
- **`polling-backoff-warn.sh`** — reads `babysit.cadence_base_minutes` (default 5), computes `WIDE_MIN=max(15,3×base)`, drops the ≥6 tier entirely (2-tier: widen or stop)
- **`scheduling-failure-modes.md`** — Pattern 5 Fix text updated; added authoritative-source decision note (issue #794)
- **`state-file-contracts.md`** — new section documenting the 3 field groups the hook reads (`last_cron_action.type/interval/at`, `babysit.cadence_base_minutes`, `babysit.cadence_effective_minutes`)
- **`session-state-schema.json`** — added `_last_cron_action_inner_keys` and `_babysit_backoff_fields` comment keys to `_field_types.pr_nested`; corrected PR-619 worked example (`last_cron_action.interval` 5m→15m, added `babysit` object)
- **`hooks/tests/polling-backoff-warn.test.sh`** (new) — 14 test cases covering all key scenarios

## Local review state

Both CLIs (CodeRabbit, CodeAnt) not installed in this environment. Self-review performed; no issues found. Coverage state: `none` (self-review only). GitHub CR/BugBot will gate the merge per the normal review flow.

Closes #794

## Test plan

- [x] `polling-backoff-warn.sh` reads `babysit.cadence_base_minutes` and computes `WIDE_MIN=max(15, 3×base)`
- [x] At streak=3 with default base 5m, hook emits `CronUpdate 15m` (not 5m — the old no-op)
- [x] At streak=3 with base=1m, floor applies: hook emits `CronUpdate 15m`
- [x] At streak=3 with base=20m, `3×20=60 > 15`: hook emits `CronUpdate 60m`
- [x] No mid-tier: streak=6 base=5m still emits 15m (≥6 tier removed)
- [x] Already-applied guard respects WIDE_MIN: `last_cron_action = {type:update, interval:15m}` → no re-emit at base=5m; `{interval:5m}` (old stale) → still emits 15m
- [x] `scheduling-reliability.md` Stable-State Backoff paragraph matches the formula in the hook and the skill
- [x] All 14 new tests in `polling-backoff-warn.test.sh` pass; full hook test suite passes

### DIFF
--- a/.claude/hooks/polling-backoff-warn.sh
+++ b/.claude/hooks/polling-backoff-warn.sh
@@ -74,10 +74,18 @@ blocker=$("$state_helper" --get "${pr_jq}.blocker // empty" 2>/dev/null || true)
 blocker_kind=$("$state_helper" --get "${pr_jq}.blocker_kind // empty" 2>/dev/null || true)
 last_cron_type=$("$state_helper" --get "${pr_jq}.last_cron_action.type // empty" 2>/dev/null || true)
 last_cron_interval=$("$state_helper" --get "${pr_jq}.last_cron_action.interval // empty" 2>/dev/null || true)
+base_min=$("$state_helper" --get "${pr_jq}.babysit.cadence_base_minutes // 5" 2>/dev/null || printf '5')
 
 if ! [[ "$streak" =~ ^[0-9]+$ ]]; then
   streak=0
 fi
+if ! [[ "$base_min" =~ ^[0-9]+$ ]]; then
+  base_min=5
+fi
+
+# Widened cadence: max(15m, 3 × base) — always strictly greater than the default 5m base.
+WIDE_MIN=$(( base_min * 3 ))
+(( WIDE_MIN < 15 )) && WIDE_MIN=15
 
 user_blocker=0
 if [[ "$blocker_kind" == "user_input" ]]; then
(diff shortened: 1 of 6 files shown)


auerbachb:
@cursor review

Issue #499, opened 2026-07-01
Roll back `/quota` skill + Stop-hook — local estimation is unreliable


auerbachb:
## Background

The `/quota` skill shipped in PR #484 (closed #458) with the goal of surfacing daily Anthropic API spend and projecting end-of-day burn against a configurable cap. The design assumed we could reliably estimate quota by capturing per-response `usage` blocks via a PostToolUse hook, aggregating by day, and comparing against a user-configured cap.

In practice this hasn't worked, and the failure mode is not just the misconfigured $3.33/day default (which #495 was meant to fix). The whole approach is misaligned with how Anthropic actually surfaces quota to users.

## Problem

Three concrete failure modes, in order of severity:

1. **Agents are gating real decisions on the fake data.** A recent transcript shows an agent declining to fan out a wave of coding agents "because quota is critical" — the wave would have been the right move, but the skill's warnings entered the agent's decision loop and blocked it. This is the load-bearing harm: the skill is no longer observational, it's influencing behavior on data that isn't authoritative.
2. **Stop-hook noise across threads for days.** Once tripped, the `[quota] today $X/$Y — critical` line fires on every turn across every thread. Even after the daily default gets fixed, the underlying mechanism (per-turn Stop-hook line) is intrusive by design — it prevents the skill from being ignored even when the user knows the warning is wrong.
3. **The measurement model can't match Anthropic's enforcement model.** Anthropic doesn't expose a machine-readable plan-tier quota endpoint. Real enforcement is monthly + rate-limits + surfaced only in the app UI (overage mode, brick-wall, per-token billing prompts). None of those signals reach any Claude Code hook we can subscribe to. Even a "correct" local estimate is comparing measured spend against a cap we're guessing at.

The in-app UI is the source of truth. It hasn't fired once during the days the local skill has been screaming.

Fix in #495 (switch to monthly $1000 + daily $100 warn) reduces the false-positive rate but doesn't address (1) or (3). The skill will still emit noise on heavy days and agents will still gate decisions on it.

## Proposed solution

Roll back PR #484 in full:

1. **Delete `.claude/skills/quota/`** — remove the skill directory and its symlink at `~/.claude/skills/quota`
2. **Remove the PostToolUse hook** — unregister `quota-usage-hook.sh` from `global-settings.json`; delete the hook script if it's not used elsewhere
[...]

PR #829, opened 2026-07-30
feat(#824): record usage-limit turn deaths; document why pre-emptive wind-down is blocked upstream

### LINKED ISSUE #824: Detect an approaching usage limit, write a handoff, and shut down
## Background

Long working sessions end abruptly. You are mid-flight — a couple of pipelines running, a PR in review, a plan half-executed — and the account hits its usage window. Whatever the thread knew about where things stood goes with it. Picking back up after the limit resets means reconstructing state from GitHub and memory, which is slow and lossy, and it is exactly the moment when a written handoff would have been worth the most.

The repo already has the machinery for this: `~/.claude/handoffs/` handoff files, `session-state.json`, and `/pm-handoff` for context turnover. What is missing is the **trigger** — nothing today notices "we are close to the wall" and spends a little of the remaining budget writing things down before the wall arrives.

## Problem

A session that runs out of quota mid-task stops in the worst possible state: no handoff written, in-flight work unrecorded, and no statement of what the next session should do first. The existing handoff tooling is all user-invoked or phase-triggered — it fires when a subagent finishes or when the user asks, never when the *account* is about to run dry.

There is also a real constraint in the way: this repo's safety rules deliberately forbid gating agent behavior on a locally-computed spend estimate. `safety.md` ("Anthropic Quota & Spend Authority") says the in-app usage UI is authoritative, that local estimation is unreliable and must not pause, downgrade, defer, or refuse work, and that local quota tracking must not be re-implemented **without an upstream Anthropic signal**. That rule exists for good reason — a wrong local estimate that shuts down a healthy session is worse than no estimate at all. So the honest first question is not "how do we build this" but "**does the runtime give us a trustworthy signal to build it on?**"

## Proposed solution

Make graceful shutdown a real behavior — but hang it on an upstream signal, never on our own arithmetic.

1. **Find out what the runtime actually exposes.** Determine whether Claude Code surfaces any first-party indication of approaching a usage limit that a hook, skill, or the session itself can read — a warning event, a rate-limit response, a usage field, anything upstream. Write down what exists and what does not.
2. **If a trustworthy signal exists, act on it.** On the signal, the session stops taking on *new* work, lets in-flight work reach a safe stopping point where it can, writes a handoff capturing what is running and what to do first on resume, tells the user plainly that it is winding down and why, and stops — rather than dying mid-sentence.
3. **If no trustworthy signal exists, say so and stop there.** The deliverable becomes a written finding plus, at most, a *user-invoked* wind-down command that does the same handoff work on demand. No inferred local budget, no heuristic shutdown — that path stays closed exactly as `safety.md` requires.

The user-visible outcome in the good case: instead of a session that vanishes, you get a short message saying it is wrapping up because the window is nearly spent, a handoff you can paste into a fresh thread when tokens return, and no orphaned in-flight work you have to rediscover.

## Acceptance Criteria

- [ ] A written finding (issue comment or `.claude/reference/` doc) stating exactly what upstream limit/usage signal Claude Code does or does not expose to a hook, skill, or session, with evidence — this gates everything below
- [ ] Any shutdown behavior is triggered **only** by that upstream signal; no locally-derived token/spend estimate ever pauses, downgrades, defers, or refuses work (`safety.md` "Anthropic Quota & Spend Authority" preserved verbatim in force)
- [ ] If a signal exists: on trigger, the session **stops accepting new work** (no new pipelines/subagents launched) while allowing in-flight work to reach a safe stopping point
- [ ] If a signal exists: a handoff is written before shutdown capturing in-flight PRs/issues/phases, watermarks, the current blocker if any, and an explicit "do this first on resume" — reusing `handoff-state.sh` / `session-state.sh` (no inline writes, per `handoff-files.md`)
- [ ] If a signal exists: the user gets one plain message naming the trigger, what was written, and where — never a silent stop
- [ ] If **no** trustworthy signal exists: the issue closes on the written finding plus (optionally) a user-invoked wind-down command; **no** heuristic or estimated-budget trigger is shipped, and `safety.md` is left unchanged
- [ ] Whatever ships is reconciled with `safety.md` in the same PR — either it demonstrably complies, or the rule is amended with explicit rationale; a silent contradiction between the rule and the behavior is a blocking failure
- [ ] Rule-corpus word budget respected (`rule-lint.sh`) for any `CLAUDE.md` / `.claude/rules/` edits

## Test Plan

- [ ] Document the signal investigation with concrete evidence (what was checked: hook payloads, API responses, runtime env, transcript metadata) — not an assertion
- [ ] If a signal exists: simulate the approaching-limit condition → verify no new pipelines launch, a handoff file is written and validates against `.claude/reference/handoff-file-schema.json`, and the user message names the trigger
- [ ] Verify the handoff is written **before** the session stops, and that a subsequent fresh session can reconstruct in-flight state from it alone
- [ ] Verify a normal session with plenty of budget is unaffected — no early wind-down, no behavior change
- [ ] Verify no code path computes a local spend estimate and uses it to gate a decision (grep the change set; confirm against `safety.md`)
- [ ] `rule-lint.sh` passes on any edited corpus files

## Notes / Open questions

- **This may correctly end as a "no" issue.** If the runtime exposes nothing trustworthy, the right outcome is a documented finding that we deliberately did not build it — that is a successful close, not a failure. Filing it makes the answer durable instead of re-litigated every few weeks.
- Open question: what counts as a "safe stopping point" for in-flight work? A PR at `merge_ready` with Phase C pending is different from a Phase A mid-edit. The wind-down probably needs a per-phase notion of where it is safe to freeze.
- Open question: should wind-down abort in-flight subagents or let them finish? Letting them finish costs tokens the session may not have; aborting risks losing their work. Leaning toward: stop launching new ones, let running ones write their handoffs, then stop.
- Related but distinct: #710 (spend/thread-type telemetry) would give *measurement*, not a limit signal; it is not a prerequisite and does not satisfy the upstream-signal requirement on its own. #773 (closed) reduced token *consumption* — this issue is about behavior at the boundary, not about using less.
- Deliberately kept separate from #823 (keeping the pipeline full continuously), per the user's request — the two interact (continuous work reaches the limit faster) but should not be designed together.

## Related Issues

- Related to #710
- Related to #823

## Related Files

- `.claude/rules/safety.md`
- `.claude/scripts/handoff-state.sh`
- `.claude/skills/pm-handoff/SKILL.md`

_Captured via /issue-maker._

## Implementation Plan

Merged from Claude's investigation + CodeRabbit's plan. **CR's plan was substantially declined** — see "Divergence from the CR plan" below; its premise was right but its mechanism does not execute in this environment.

### Finding (the gate) — no trustworthy approaching-limit signal exists

Audited Claude Code **2.1.219** by `strings`-extracting the shipped binary plus a live runtime probe:

| Surface | Carries `rate_limits`? | Usable as a trigger? |
|---|---|---|
| **statusLine** stdin | **Yes** — `five_hour`/`seven_day` `used_percentage` + `resets_at` | **No** — TUI-render-path only; stdout goes to a React state var, never to model context |
| **All 32 hook events** | **No** — shared payload base is `{session_id, transcript_path, cwd, prompt_id, permission_mode, agent_id, agent_type, effort}` | No |
| **`StopFailure`** | Post-hoc only (`error: "rate_limit"`) | Only as a **recorder** — fires after the wall; "fire-and-forget, output and exit codes ignored" |
| **`/usage`** | Yes (authoritative) | No — in-app UI, no machine-readable export |

Decisive extra evidence: this desktop-app session runs headless (`claude --output-format stream-json --input-format stream-json`), so **no status line renders at all**. A registered statusLine probe that dumped its raw stdin produced **0 invocations across ~6 minutes and dozens of tool calls**.

**Verdict:** the signal exists in the runtime but reaches **none** of hook, skill, or session. The issue's own no-signal branch applies.

### Divergence from the CR plan

CR proposed a statusLine → marker-file → PostToolUse relay plus a `safety.md` amendment. Declined because:

1. The observer half **never runs** in the desktop app — the relay would be silently inert exactly where the user works. Worst failure mode for a safety-adjacent feature: looks shipped, does nothing.
2. It would make correctness depend invisibly on client choice (terminal TUI vs. desktop app).
3. Amending a deliberately strict safety rule to accommodate a mechanism that cannot fire weakens the rule for zero gain. `safety.md` is left **unchanged**.

### Files

| File | Change |
|---|---|
| `.claude/reference/usage-limit-signal-audit-2026-07.md` | **New** — the gating finding: evidence, what NOT to build, re-check triggers |
| `.claude/hooks/usage-limit-record.sh` | **New** — `StopFailure`/`rate_limit` recorder |
| `.claude/hooks/tests/usage-limit-record.test.sh` | **New** — 10 assertions |
| `global-settings.json` | Register `StopFailure` (first use of this event here), matcher `rate_limit`, timeout 5 |
| `setup-skills-worktree.sh` | `HOOKS_MANIFEST` entry so fresh installs register it |
| `.claude/hooks/README.md` | Document the hook and why it is a recorder, not a wind-down |
| `.claude/rules/safety.md` | **Deliberately unchanged** |

### What shipped

A single post-hoc recorder. On — and only on — the runtime reporting `error: "rate_limit"`, it appends `{recorded_at, session_id, cwd, transcript_path, truncated last_assistant_message, resume_hint}` to `~/.claude/usage-limit-events.jsonl` (rotated at 256 KiB) and atomically rewrites `~/.claude/usage-limit-last.json`.

- **Upstream-triggered only** — sole input is the runtime's own error classification; computes nothing.
- **Gates no decision** — never pauses, downgrades, defers, or refuses work, so `safety.md` holds as written.
- **Recorder, not shutdown** — the value is that the next session finds a breadcrumb instead of silence.

This does not deliver the pre-emptive wind-down the issue hoped for. That is blocked upstream, and the finding doc records the four conditions that would unblock it.

### Why not estimate locally

Issue #499 already rolled back a `/quota` tracker (PR #484) precisely because agents gated real decisions on the estimate — one declined a correct wave "because quota is critical." That rollback is *why* `safety.md` §"Anthropic Quota & Spend Authority" exists. Its inert residue (`~/.claude/quota-usage.log`, last written 2026-07-01) is left alone as user data.

### Verification

- 10-assertion test; both injected regressions (removed `rate_limit` guard; drifted timeout) confirmed **failing** before restore
- All 8 hook suites and all **50** CI-discovered bash suites pass
- `rule-lint: OK`; corpus 11,458 words (no rule files touched)

auerbachb:
## Summary

Issue #824 asked for a session that notices an approaching usage limit, writes a handoff, and shuts down gracefully — explicitly **gated** on first proving a trustworthy upstream signal exists to hang it on, because `safety.md` forbids acting on a locally-computed estimate.

**The finding is the deliverable: no such signal reaches a hook, skill, or session.** So this PR ships the issue's own no-signal branch — the written finding, no heuristic trigger, and `.claude/rules/safety.md` **unchanged** — plus the one genuinely upstream-triggered thing the runtime does permit: a post-hoc recorder.

## The investigation

Audited **Claude Code 2.1.219** (SHA-256 `49b1845c…`) by `strings`-extracting the shipped binary plus a live runtime probe.

| Surface | Carries `rate_limits`? | Usable as a trigger? |
|---|---|---|
| **statusLine** stdin | **Yes** — `five_hour`/`seven_day` `used_percentage` + `resets_at` | **No** — runs only on the Ink/React TUI render path; stdout lands in a React state var, never in model context |
| **All 32 hook events** | **No** — shared payload base is `{session_id, transcript_path, cwd, prompt_id, permission_mode, agent_id, agent_type, effort}` | No |
| **`StopFailure`** | Post-hoc only (`error: "rate_limit"`) | Only as a **recorder** — fires after the turn died; "fire-and-forget, output and exit codes ignored" |
| **`/usage`** | Yes (authoritative) | No — in-app UI, no machine-readable export |

Decisive: this desktop-app session runs headless (`claude --output-format stream-json --input-format stream-json`), so **no status line renders at all**. A registered statusLine probe that dumped its raw stdin produced **0 invocations across ~6 minutes** of heavy tool use.

## Why CodeRabbit's plan was declined

CR planned a statusLine → marker-file → PostToolUse relay plus a `safety.md` amendment. Its premise (a real `rate_limits` signal exists) is correct; the mechanism is not:

1. The observer half **never executes** in the desktop app — the relay would be silently inert exactly where the user works. Worst failure mode for a safety-adjacent feature: looks shipped, does nothing.
2. It would make correctness depend invisibly on client choice (terminal TUI vs. desktop app).
3. Amending a deliberately strict safety rule to accommodate a mechanism that cannot fire weakens the rule for zero gain.

Local estimation stays closed on its own merits: #499 rolled back the `/quota` tracker because agents gated real decisions on it — one declined a correct wave "because quota is critical." That rollback is *why* the safety rule exists.

## What ships

`.claude/hooks/usage-limit-record.sh`, registered on **`StopFailure`** with matcher **`rate_limit`** (first use of that event in this repo). On the runtime's own error classification it appends `{recorded_at, session_id, cwd, transcript_path, truncated last_assistant_message, resume_hint}` to `~/.claude/usage-limit-events.jsonl` (rotated at 256 KiB) and atomically rewrites `~/.claude/usage-limit-last.json`.

- **Upstream-triggered only** — sole input is the runtime's error classification; computes nothing.
- **Gates no decision** — never pauses, downgrades, defers, or refuses work, so `safety.md` holds as written.
- **Recorder, not shutdown** — the next session finds a breadcrumb instead of silence.

It does **not** deliver the pre-emptive wind-down the issue hoped for. That is blocked upstream; the finding doc records the four conditions that would unblock it.

## Review notes

**Local CLI coverage: `none`** — both CLIs were unavailable, so this change set had a **self-review** as its final gate.

- **CodeAnt:** 403 + `meta.capped: true` on both the initial run and the one sanctioned retry (undocumented daily cap). Dropped per `cr-local-review.md`; deliberately **not** re-authenticated, since the documented cause is a cap, not auth.
- **CodeRabbit:** the first run was verified-successful and returned **8 findings — all 8 verified as valid and fixed** (below). The confirming second pass hit the free-OSS tier cap (`rate_limit`, `waitTime: 44 minutes`), which is a failed run, not a clean pass.

Per `feedback_review_clis_down_app_independent.md`, CLI unavailability does not block merge — the GitHub App reviewers have independent quotas and still gate it.

**Findings fixed (all 8):**
- *major* — state files were world-readable despite embedding conversation content → `umask 077` + defensive `chmod 600`
- *major* — no cross-process lock. Concurrency is the **expected** case here (one account hitting its wall fails every active session at once) → bounded, portable `mkdir` mutex (macOS ships no `flock`), publish-only-on-successful-append
- *major* — split registration assertions could accept matcher on one entry and timeout on another → consolidated into one conjunction
- *minor ×4 / trivial ×1* — truncation test could pass without the record ever being written; batch tests compared line counts instead of content; events-log fields unasserted; "Three facts" introduced four; audit doc lacked reproduction details

Self-review additionally caught a latent `trap` quoting bug (`$LOCK_DIR` expanded at definition time), now fixed and covered by a quoted-path test.

## Verification

- 12-assertion test suite; injected regressions confirmed **failing** before restore (removed `rate_limit` guard; drifted timeout; split registration)
- Concurrency: 25 simultaneous hooks → 25 records, 0 interleaved lines, 25 distinct sessions, no leftover lock
- Permissions asserted `600`; quoted-path (`it's a dir`) handled
- All **50** CI-discovered bash suites pass; `rule-lint: OK`; `shellcheck` clean
- Corpus 11,458 words — no rule files touched

## Test plan

- [x] Signal investigation documented with concrete evidence (binary strings, hook payload base, live probe), not assertion
- [x] No shutdown/wind-down behavior ships on a heuristic; the only trigger is the runtime's `error: "rate_limit"`
- [x] No code path computes a local spend estimate to gate a decision (asserted by test 8, comments stripped)
- [x] `.claude/rules/safety.md` is unchanged
- [x] A normal session with plenty of budget is unaffected — non-`rate_limit` errors and malformed payloads mutate neither state file
- [x] Handoff/record is written durably and a fresh session can find it (`usage-limit-last.json` + append-only log)
- [x] Registration is consistent across `global-settings.json` and `HOOKS_MANIFEST` (matcher + command + timeout on one entry)
- [x] `rule-lint.sh` passes and the corpus budget is respected
- [x] All CI-discovered test suites pass

Closes #824

### DIFF
--- a/.claude/hooks/README.md
+++ b/.claude/hooks/README.md
@@ -143,3 +143,27 @@ Both write through `lib/skill-usage-recorder.sh`, which owns the storage layout 
 
 - `python3` (both hooks) and `jq` (PostToolUse tracker only)
 - Registered automatically via `global-settings.json` — see **Hook Auto-Registration** above
+
+## usage-limit-record.sh
+
+Records a durable breadcrumb when a turn dies because the account hit an Anthropic usage limit, so the next session finds a pointer instead of silence (issue #824).
+
+Registered on **`StopFailure`** with matcher **`rate_limit`** — the first use of that event in this repo. The runtime fires `StopFailure` instead of `Stop` when an API error ends a turn, and matches on the payload's `error` field.
+
+**This is a recorder, not a wind-down.** `StopFailure` is documented by the runtime as *"Fire-and-forget — hook output and exit codes are ignored"*, and it fires only after the turn has already failed. The hook therefore cannot warn the session, inject context, or gate any decision — it writes to disk and exits 0.
+
+Claude Code exposes **no** approaching-limit signal to any hook, skill, or session; the only surface carrying `rate_limits` is the status line, which is display-only and never executes in a headless desktop-app session. Full evidence: `.claude/reference/usage-limit-signal-audit-2026-07.md`.
+
+**Safety.** Its sole input is the runtime's own `error` classification. It estimates nothing and gates nothing, so `.claude/rules/safety.md` §"Anthropic Quota & Spend Authority" is satisfied as written and was not amended.
+
+**Outputs**
+
+- `~/.claude/usage-limit-events.jsonl` — append-only history, rotated to `.1` past 256 KiB
+- `~/.claude/usage-limit-last.json` — the most recent event, written atomically; read this first when resuming
+
+Each record carries `recorded_at`, `session_id`, `cwd`, `transcript_path`, a truncated `last_assistant_message`, and a `resume_hint`.
+
+### Prerequisites
+
+- `jq` must be installed — the hook exits 0 silently without it
+- Override the output directory with `CLAUDE_USAGE_LIMIT_DIR` (used by the tests)
--- a/.claude/hooks/tests/usage-limit-record.test.sh
+++ b/.claude/hooks/tests/usage-limit-record.test.sh
@@ -134,9 +134,19 @@ grep -q 'Verdict' "$AUDIT_DOC" || fail "finding doc does not state a verdict"
 
 # --- 11. State files are owner-only ---
 # The record embeds conversation content and absolute paths.
+# Switch on uname, matching the hook's own file_size() helper: GNU `stat -f`
+# means "filesystem status" and SUCCEEDS with unrelated output, so a
+# `stat -f ... || stat -c ...` fallback never reaches the GNU form on Linux.
+file_mode() {
+  if [[ "$(uname -s)" == "Darwin" ]]; then
+    stat -f '%Lp' "$1" 2>/dev/null
+  else
+    stat -c '%a' "$1" 2>/dev/null
+  fi
+}
 for f in "$EVENTS" "$LAST"; do
-  MODE=$(stat -f '%Lp' "$f" 2>/dev/null || stat -c '%a' "$f" 2>/dev/null)
-  [[ "$MODE" == "600" ]] || fail "$(basename "$f") is mode $MODE, expected 600 (owner-only)"
+  MODE=$(file_mode "$f")
+  [[ "$MODE" == "600" ]] || fail "$(basename "$f") is mode '$MODE', expected 600 (owner-only)"
 done
 
 # --- 12. Concurrent invocations lose no records ---
@@ -164,5 +174,35 @@ UNIQ=$(jq -r '.session_id' "$CONC_LOG" | sort -u | wc -l | tr -d ' ')
 
 [[ -d "$CONC_DIR/.usage-limit-record.lock" ]] && fail "lock directory was left behind"
 true
+
+# --- 13. Non-string fields are truncated too ---
+# A structured error_details object must not bypass the breadcrumb limit.
+STRUCT_DIR="$TMP_DIR/struct"
+mkdir -p "$STRUCT_DIR"
+BIG_VAL=$(printf 'y%.0s' $(seq 1 4000))
+printf '%s' "{\"hook_event_name\":\"StopFailure\",\"error\":\"rate_limit\",\"error_details\":{\"blob\":\"$BIG_VAL\"},\"last_assistant_message\":{\"nested\":\"$BIG_VAL\"}}" \
+  | CLAUDE_USAGE_LIMIT_DIR="$STRUCT_DIR" bash "$HOOK"
+
+STRUCT_LAST="$STRUCT_DIR/usage-limit-last.json"
+[[ -f "$STRUCT_LAST" ]] || fail "structured payload produced no record"
+jq -e . "$STRUCT_LAST" >/dev/null || fail "structured payload produced invalid JSON"
+ED_LEN=$(jq -r '.error_details | length' "$STRUCT_LAST")
+LAM_LEN=$(jq -r '.last_assistant_message | length' "$STRUCT_LAST")
+[[ "$ED_LEN" -le 500 ]] || fail "object error_details not truncated (len=$ED_LEN)"
+[[ "$LAM_LEN" -le 1000 ]] || fail "object last_assistant_message not truncated (len=$LAM_LEN)"
+
+# --- 14. Rotation fires at the cap and preserves the archive ---
+ROT_DIR="$TMP_DIR/rotate"
+mkdir -p "$ROT_DIR"
+ROT_LOG="$ROT_DIR/usage-limit-events.jsonl"
+# Seed just past the 256 KiB cap so the next append must rotate.
+head -c 262200 /dev/zero | tr '\0' 'a' >"$ROT_LOG"
+printf '\n' >>"$ROT_LOG"
+printf '%s' '{"hook_event_name":"StopFailure","error":"rate_limit","session_id":"rot","last_assistant_message":"em—dashes—are—multibyte"}' \
+  | CLAUDE_USAGE_LIMIT_DIR="$ROT_DIR" bash "$HOOK"
+
+[[ -f "$ROT_LOG.1" ]] || fail "rotation did not create the .1 archive"
+[[ "$(wc -l <"$ROT_LOG" | tr -d ' ')" == "1" ]] || fail "post-rotation log should hold exactly the new record"
+jq -e '.session_id == "rot"' <"$ROT_LOG" >/dev/null || fail "post-rotation log lost the new record"
 
 echo "PASS: usage-limit-record.sh"
--- a/.claude/hooks/tests/usage-limit-record.test.sh
+++ b/.claude/hooks/tests/usage-limit-record.test.sh
@@ -1,0 +1,168 @@
+#!/usr/bin/env bash
+# Tests for usage-limit-record.sh — the StopFailure recorder (issue #824).
+#
+# Covers the behavior the issue's Test Plan requires: the recorder fires only
+# on the upstream rate-limit signal, a healthy session is completely
+# unaffected, and no code path computes a local spend estimate.
+set -euo pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+HOOK="$REPO_ROOT/.claude/hooks/usage-limit-record.sh"
+SETTINGS="$REPO_ROOT/global-settings.json"
+SETUP_SCRIPT="$REPO_ROOT/setup-skills-worktree.sh"
+AUDIT_DOC="$REPO_ROOT/.claude/reference/usage-limit-signal-audit-2026-07.md"
+SAFETY_RULE="$REPO_ROOT/.claude/rules/safety.md"
+
+fail() { echo "FAIL: $*" >&2; exit 1; }
+
+command -v jq >/dev/null 2>&1 || { echo "SKIP: jq not installed"; exit 0; }
+
+TMP_DIR="$(mktemp -d)"
+cleanup() { rm -rf "$TMP_DIR"; }
+trap cleanup EXIT
+
+run_hook() {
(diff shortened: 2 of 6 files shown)


auerbachb:
@graphite-app re-review

auerbachb:
Fixed in `aa82087`: rotation now measures the exact bytes appended via `printf '%s\n' "$RECORD" | wc -c` instead of `${#RECORD}`. Correct — that was a character count, so multibyte UTF-8 in the captured assistant message (em-dashes are routine here) under-measured the append and deferred rotation past the 256 KiB cap. The newline is now included because the printf emits exactly what gets appended. Added a regression test that seeds the log past the cap and asserts the .1 archive is created.

auerbachb:
Fixed in `aa82087`: non-string values are now serialized with `tojson` before slicing, so an object or array `error_details` is truncated to the same 500/1000-char breadcrumb limits instead of being embedded whole. Added a regression test that feeds object-valued `error_details` and `last_assistant_message` and asserts both are truncated.

auerbachb:
Fixed in `aa82087`: rotation now runs only while the lock is held. The prior 'proceed anyway' fallback was reasoned about the append, not the rotation — and the two differ in consequence: an unlocked append risks at worst an interleave, while an unlocked rm+mv can replace the archive with a freshly-emptied log and lose history outright. The unlocked path still appends and simply defers rotation to a later locked run.

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

Issue #838, opened 2026-07-30
Stale chips in the task list still carry the pre-#837 model guard


auerbachb:
## Background

[Issue #837](https://github.com/auerbachb/claude-code-config/issues/837) made the MODEL GUARD compare model *families* instead of exact strings, so a `**Model:**` line reading `Opus <N>.<N>` no longer hard-stops a thread that is already running Opus.

## Problem

That fix reaches payloads emitted from here on — it does not reach the ones already offered.

A chip copies the preamble into its `prompt` **at spawn time**. The text is frozen there. Editing `.claude/reference/chip-launching.md` rewrites the template, not the payloads already sitting in the task list, and a thread launched from an old chip reads the old wording and never consults the file. The same is true of any paste block someone saved in a note.

So #837's own framing — that fixing the comparison rule "fixes every past and future stale payload at once" — holds for the future half only. #837 stops the class from growing; it does not drain the existing backlog. Old chips remain a coin flip: the pre-#837 wording never defined what a mismatch *is*, so a thread reading it may stop or may proceed.

This was flagged during #837's local review rather than at capture time, and #837 shipped with the limitation documented in `chip-model-guard-decision.md` rather than silently.

## Proposed solution

Decide what to do with the backlog. Roughly, in increasing cost:

1. **Nothing.** Old chips age out as they are clicked or dismissed. The failure is a single wasted round-trip, and the user can reply "continue anyway". Cheapest, and defensible if the backlog is small.
2. **Dismiss the stale ones.** Withdraw chips whose payloads predate #837 so they cannot produce a false stop, and let the issue be re-offered fresh by whichever emitter surfaces it next.
3. **Re-emit.** Spawn a replacement chip per stale issue, then dismiss the old one — the "Re-planned" trigger already defined in `chip-launching.md` "Stale-chip hygiene". Most work; also the only option that preserves the offer.

#837 rejected (3) as its *primary* fix and was right to — a one-time sweep does not fix a recurring problem, which is why the reading rule was changed first. With the recurring half now closed, a one-time sweep is exactly the right shape for the remaining one-time backlog, if it is big enough to be worth it.

First step either way is a count: how many live chips predate #837, and are they concentrated in a few `/pm` or `/issue-maker` batches?

## Acceptance Criteria

[...]

Issue #802, opened 2026-07-30
chip-model-guard-lint: first-line placement check accepts generic wording without a **Model:** reference


auerbachb:
## Background

`chip-model-guard-lint.sh` enforces that every canonical chip emitter documents the `**Model:**` line as the *first* line of the chip `prompt`. It does that with one broad alternation:

```
first line|first prompt|MUST open|open the chip|MUST open with|first content|base block
```

## Problem

Any of those phrases satisfies the check no matter what they are talking about. A skill that says "MUST open with the task description" passes; so does one whose only match is an unrelated sentence containing "first line". The check confirms a *phrase* is present, not that the skill actually states the `**Model:**` line comes first — which is the thing it exists to protect.

This surfaced during PR #799 in a concrete way: rewording `/pm`'s contract paragraph from "as the **first line** of the `prompt`" to "the `**Model:**` line **first**" dropped the match and turned the lint red, even though the new wording said the same thing more clearly. The check is simultaneously too loose (unrelated phrases pass) and too brittle (a correct rewording fails).

The sibling checks have already been tightened this way. The Fable pre-click warning moved from a bare `Fable` match to `parent thread is on Fable` in #791, precisely because a bare model mention is not evidence the warning is documented.

## Proposed solution

Require the placement phrase and an explicit `**Model:**` reference to co-occur, the way the pre-click check now requires its phrase in context. A single pattern that matches a placement claim *about the model line* — rather than either half independently — would do it.

Whatever pattern is chosen, it has to be checked against all six emitters as written before landing: the point is to reject genuinely-malformed skills, not to force one blessed phrasing on authors who are already correct.

## Acceptance Criteria

- [ ] The first-line placement check fails when a skill documents a first-position claim that is not about the `**Model:**` line
- [ ] All six canonical emitters (`/pm`, `/prompt`, `/start-issue`, `/issue-maker`, `/wave`, `/harness-audit`) still pass unchanged, or are reworded in the same PR with the change explained
- [ ] A mutation fixture in `chip-model-guard-lint.test.sh` covers the new failure mode, and is verified to fail when the tightened check is reverted
- [ ] Fixtures go through the existing `mutate` helper so a no-op mutation is loud rather than silently vacuous

## Test Plan

- [ ] `.github/scripts/chip-model-guard-lint.sh` exits 0 on the tree
- [ ] It exits 1 against a scratch copy where an emitter's placement wording is decoupled from `**Model:**`
[...]

Issue #791, opened 2026-07-30
Versionless model names + effort recommendations on every surface


auerbachb:
## Background

Whenever we hand work off — a click-to-launch chip, a copy/paste prompt, or a subagent spawn — we tell the receiving thread which model to run on. That part works. Chips and paste blocks both carry a `**Model:**` line plus a MODEL GUARD that stops the thread on a mismatch, and it's enforced in CI by `chip-model-guard-lint.sh` across all five canonical emitters (`/pm`, `/prompt`, `/start-issue`, `/issue-maker`, `/wave`) plus ad-hoc offers via `chip-spawn.md`. Subagents get it structurally instead of by prose — the Agent tool takes a real `model` parameter, and every spawn site is required to set it explicitly.

Two things are inconsistent, though. First, we name models with version numbers baked in — "Opus 5", "Sonnet 5", "Fable 5", "Haiku 4.5" — which means every model release turns into a documentation sweep, and right now the corpus is already carrying four generations of stale names side by side. Second, the effort level — the other half of the picker the user actually sets before clicking a chip — is almost never mentioned. Only `/prompt` recommends one. The four other emitters recommend a model and stay silent on effort, so in practice you never see an effort suggestion at the moment you're about to launch something.

## Problem

**Version numbers are churn we chose, and they've already drifted.** The corpus contains `Opus 4`, `Opus 4.7`, `Opus 4.8`, `Opus 5`, `Sonnet 4`, `Sonnet 4.6`, `Sonnet 5`, `Haiku 4.5`, and `Fable 5` — 178 versioned mentions across ~11 rule/skill/agent files (50 in `/prompt` alone, 13 in `.claude/agents/README.md`). Every one is a thing to update on the next release, and the ones already stale are actively misleading. The family name is the only part that carries a real decision; the version is noise we volunteered to maintain (it was pinned deliberately in #749 — this reverses that call).

[...]

PR #842, opened 2026-07-30
fix(#837): compare model families in the MODEL GUARD, not version numbers

### LINKED ISSUE #837: Model guard should compare model families, not version numbers
## Background

Every click-to-launch chip and paste-in prompt opens with a `**Model:**` line naming which model the receiving thread should run on, followed by a MODEL GUARD preamble telling that thread to check itself against the recommendation before doing anything. When the two agree, the thread says so in one line and gets to work. When they disagree, it stops and waits for you.

That guard is doing its job for the case it was designed for — a chip that wants Sonnet being clicked in an Opus thread, or vice versa. Real mismatch, real stop, real value.

## Problem

The guard also fires on chips that aren't actually mismatched — it just can't tell.

Chips stay clickable indefinitely. A chip offered days or weeks ago sits in the task list carrying whatever `**Model:**` line was current when it was created, and until #791 landed, that line named a specific version: "Opus 4.8", "Sonnet 4.6". Click one of those today and the fresh thread compares "Opus 4.8" against the Opus it's actually running, sees two different strings, and hard-stops before reading a single file — reporting an over-powered mismatch and asking whether to continue.

Nothing is wrong. The chip wanted Opus. The thread is Opus. The only difference is a version qualifier that stopped being meaningful the moment we decided family names are how we name models. But the guard has no way to know that, because it was never told what a mismatch actually *is* — the preamble just says to compare the recommendation against what you're running and stop on a difference. String equality is the only reading available.

The result is a stop that costs you a full round-trip and teaches the wrong lesson: the guard cries wolf on the common case (an older chip), which is exactly how a safety check gets mentally filed as noise and waved through on the day it's right.

This is not stale config on any one machine. The versionless rename shipped and is fully synced; every *new* chip already carries a bare family name. What #791 fixed was how we **write** the model line. How a thread **reads** one it doesn't recognize was never part of that change, so the backlog of already-emitted chips keeps hitting it — and no corpus sweep can reach them, because they're payloads sitting in a task list, not text in the repo.

## Proposed solution

Teach the guard that a model family is the thing being recommended, and a version number attached to it is noise to look past.

Clicking an older chip that says "Opus 4.8" from a thread running Opus should produce the same one-line "running as recommended, proceeding" you'd get from a chip that just says "Opus" — no stop, no question, straight to work. A chip that names a genuinely different family still stops the thread cold, because that's a real disagreement about how much model the work deserves.

The guard keeps all of its teeth for the case it exists to catch, and stops spending them on chips that were only ever guilty of being written down before we changed how we write things down.

## Acceptance Criteria

- [ ] The MODEL GUARD preamble in `.claude/reference/chip-launching.md` defines the comparison as **family-level**: the recommendation and the running model match when their family names (`Opus`, `Sonnet`, `Haiku`, `Fable`) are the same, regardless of any trailing version qualifier on either side.
- [ ] The preamble states explicitly that a trailing `<N>` or `<N>.<N>` after a family name on the `**Model:**` line is a legacy version qualifier to be ignored for comparison purposes — it is not evidence of a mismatch.
- [ ] A cross-family difference (recommendation names one family, thread is running another) still triggers the full existing STOP behavior in **both** directions, under- and over-powered, with no softening of the stop wording.
- [ ] The match branch's one-line self-report works when the `**Model:**` line carries a version qualifier — the thread reports the family it is running, not the stale string it read.
- [ ] The change is made in `.claude/reference/chip-launching.md` only; emitters reference the preamble rather than duplicating it, so no `.claude/skills/` file needs to restate the new comparison rule. Confirm this holds (`grep -rl "MODEL GUARD: Your very first action"` returns exactly the one file) rather than assuming it.
- [ ] `.claude/reference/chip-model-guard-decision.md` records the family-level comparison decision and why version qualifiers are ignored, alongside its existing model-only-not-effort note.
- [ ] `chip-model-guard-lint.sh` asserts the preamble still carries the family-comparison sentence, so a future reword can't silently drop it back to string equality.
- [ ] `chip-model-guard-lint.sh`'s test file covers the new assertion, and each new assertion is verified to fail when its fix is reverted.

## Test Plan

- [ ] Paste a prompt whose `**Model:**` line reads `Opus 4.8 (Heavy)` into a fresh Opus thread: it reports running as recommended in one line and proceeds to the task with no stop and no question.
- [ ] Paste a prompt whose `**Model:**` line reads `Sonnet` into a fresh Opus thread: it stops, reports both models, and waits — unchanged from today.
- [ ] Paste a prompt whose `**Model:**` line reads `Opus` into a fresh Opus thread: unchanged happy path, one-line confirmation.
- [ ] `bash .github/scripts/chip-model-guard-lint.sh` exits 0 on the updated tree.
- [ ] `bash .github/scripts/chip-model-guard-lint.sh` exits 1 against a scratch copy with the family-comparison sentence deleted from the preamble.
- [ ] `bash .github/scripts/tests/chip-model-guard-lint.test.sh` passes; revert the preamble change and confirm the new assertion fails (a test that passes both ways proves nothing).
- [ ] Click one of the genuinely old chips still sitting in the task list (several predate the versionless rename) and confirm it launches into work instead of stopping.

## Notes / Open questions

- **Scoped to the guard's reading rule, not the chip backlog.** The obvious alternative is sweeping and re-emitting stale chips so their `**Model:**` lines get rewritten. Rejected as the primary fix: it's a one-time cleanup against a recurring problem, it needs a handle on every outstanding chip, and it does nothing for a paste block someone saved in a note six weeks ago. Fixing the comparison rule fixes every past and future stale payload at once. Worth revisiting separately if stale chips turn out to cause problems beyond this one.
- **Does this weaken the guard?** It shouldn't — the guard's stated purpose is catching a thread running at the wrong *tier*, and tier is the family. Two Opus versions are the same recommendation. Worth confirming that reading is what we want before implementing, since it's the whole premise.
- **What about a family we retire?** If a family name is ever dropped from the fleet, a chip naming it would compare against nothing and should probably stop. Not a live case today; noting it so the implementer doesn't have to guess.
- **Relationship to #791:** #791 fixed the emission side (all new chips carry bare family names) and is closed. This is the read side of the same contract, which #791's acceptance criteria never covered — every one of them is about what gets *written* into the corpus.
- **Relationship to #752:** that issue asks whether a guard should catch model-fleet drift at PR time. Different surface — PR-time linting of our own files vs. runtime interpretation of an already-emitted payload — but they're neighbors, and whoever picks up one should glance at the other.

## Related Issues

- Related to #791 (versionless model names — fixed the write side of this contract)
- Related to #752 (PR-time model-fleet drift detection)
- Related to #749 (the versioned-name pin this line of work reversed)

## Related Files

- `.claude/reference/chip-launching.md`
- `.claude/reference/chip-model-guard-decision.md`
- `.github/scripts/chip-model-guard-lint.sh`
- `.github/scripts/tests/chip-model-guard-lint.test.sh`

_Captured via /issue-maker._

## Implementation Plan

_Merged: Claude's analysis + CodeRabbit's recommendations._

### Files to modify

| File | Change |
|------|--------|
| `.claude/reference/chip-launching.md` | Rewrite the verbatim MODEL GUARD preamble to compare families; add a short cross-reference paragraph anchoring it to the existing family-name resolution rule |
| `.claude/reference/chip-model-guard-decision.md` | New `### Amendment (#837)` under `## Decision` + two `## References` entries |
| `.github/scripts/chip-model-guard-lint.sh` | Three `require_pattern` assertions in the `chip-launching.md` contract section |
| `.github/scripts/tests/chip-model-guard-lint.test.sh` | Four `expect` cases — one revert-fails case per assertion, plus a "weakened reword" strictness case |

**No `.claude/skills/` or `.claude/rules/` file is touched.** Pre-verified: `grep -rl "MODEL GUARD: Your very first action"` returns exactly `.claude/reference/chip-launching.md`, so every emitter inherits the new semantics by reference (AC5).

### Steps

1. **Rewrite the preamble** (AC1–AC4). Reframe the opening compare instruction to compare *families*; add a sentence that a trailing `<N>` / `<N>.<N>` after a family name is an old-style version qualifier to be ignored **on either side**, never evidence of a mismatch. Qualify the Match branch as `(same family)` and change its self-report from `{MODEL}` to `{FAMILY}`, naming the family being run rather than echoing the possibly-qualified string read. Qualify the Mismatch branch as `(different family)` and leave **every word after `STOP.` untouched** — AC3 forbids softening the stop wording.
2. **Anchor to existing vocabulary.** In the prose *around* the preamble (not inside the verbatim block, which is pasted into every chip), cross-reference the established "a bare family name resolves to the newest non-legacy model of that family" rule so family reasoning is not a new concept.
3. **Amendment in the decision record** (AC6). Verdict first (family is the guarded axis, version qualifier is noise), then: why — tier is the family; how it composes with #791 (that fixed the *write* side; already-emitted chips are the unreachable *read* side); the rejected alternative (sweep and re-emit stale chips); the retired-family edge case as a known non-live gap; and the terminology note below. Add `#837` and `#749` to `## References`.
4. **Lint assertions** (AC7) in the `chip-launching.md` section of `chip-model-guard-lint.sh`, mirroring the existing `require_pattern` style:
   - `Compare families only` → family-level comparison rule
   - `old-style version qualifier` → version-qualifier-is-noise rule
   - `Running on [{]FAMILY[}]` → family self-report in the match branch
5. **Tests** (AC8) in `chip-model-guard-lint.test.sh`: one `mutate`-based `expect` per assertion (so a fragment that drifts from the real wording trips the no-op guard rather than passing vacuously), plus a `weaken_family_comparison` case in the `weaken_fable_warning` mould — a reword that still contains the word "families" but reverts to string equality must still fail, proving the assertion binds to the contractual phrase and not a bare mention.
6. **Verify each new assertion fails when its fix is reverted** — not merely that the suite is green.

### Design decisions

- **"old-style", not "legacy", for the version qualifier.** `.claude/agents/README.md` already uses *legacy* to mean "a superseded model within a family" ("the newest non-legacy model of that family"). Reusing it for the version suffix invites precisely the wrong reading — that only qualifiers naming an *outdated* version are ignorable. Every trailing qualifier is ignored regardless of which version it names. This satisfies AC2's substance (the trailing `<N>`/`<N>.<N>` is stated explicitly as a version qualifier to ignore, not evidence of mismatch) while declining its incidental word choice; recorded in the decision doc.
- **Placeholder `<N>` notation, never a concrete example.** `chip-model-guard-lint.sh`'s `VERSIONED_RE` (`(Opus|Sonnet|Fable|Haiku) [0-9]+(\.[0-9]+)?`) scans `chip-launching.md` and `chip-model-guard-decision.md`. Writing an illustrative `Opus <digit>` would fail the repo's own versionless-name lint. Placeholder notation keeps the new prose lint-safe.
- **Three assertions, not one.** AC1 (family comparison), AC2 (version qualifier ignored), and AC4 (family self-report) are separately load-bearing; a single assertion would leave two of them free to regress silently.
- **Bracket expressions over `\{`** in the third lint pattern — `[{]FAMILY[}]` is unambiguous under both GNU and BSD ERE, where `\{` shades into interval-expression territory.

### Risks / edge cases

- **Preamble length.** The block is pasted verbatim into every chip prompt, so the rewrite stays tight — roughly three added lines, no restructuring.
- **Lint brittleness.** Three new prose assertions raise the odds a legitimate reword trips the lint. Accepted: this file's entire job is being a verbatim contract, it already carries ~12 such assertions, and a tripped lint is a visible prompt to update the test — the failure mode AC7 exists to force.
- **Retired family.** A chip naming a family that has left the fleet would compare against nothing and should stop. Not a live case; documented in the decision record rather than guessed at in the preamble.
- **Word budget.** Both edited docs are under `.claude/reference/`, which is not auto-loaded, so the rule corpus is unaffected.

### Verification

- `bash .github/scripts/chip-model-guard-lint.sh` exits 0 on the updated tree.
- `bash .github/scripts/tests/chip-model-guard-lint.test.sh` passes, including its real-repo conformance case.
- Each new assertion is individually confirmed to exit 1 against a scratch copy with its sentence removed — a test green in both directions proves nothing.
- `grep -rl "MODEL GUARD: Your very first action"` still returns exactly one file.
- `rule-lint.sh` word budget unchanged (no rule-corpus file touched).
- Behavioural walk-through of the three paste cases in the Test Plan against the final preamble text.

auerbachb:
Closes #837

## What changed

The MODEL GUARD told a launched thread to compare "the model you are actually running as" against the `**Model:**` line and stop on a difference — without ever defining what a difference *is*. String equality was the only reading available, so a chip emitted before #791 (carrying a family name plus a trailing version number) hard-stopped a thread already running that family. Right tier, real chip, full round-trip lost — and a check that cries wolf on the common case is one that gets waved through on the day it's right.

The comparison is now **family-level**:

- Families are the compared unit — `Opus`, `Sonnet`, `Haiku`, `Fable`.
- A trailing `<N>` or `<N>.<N>` after a family name is an old-style version qualifier, ignored on **either** side, never evidence of a mismatch.
- A **cross-family** difference still stops cold in both directions. Every word after `STOP.` is byte-unchanged.
- The **match** branch now reports the *family* it is running (`{FAMILY}`) instead of echoing the possibly-stale string it read.

The two branches report differently on purpose: the match branch names the family (echoing a stale qualifier would assert a version the thread isn't running), while the mismatch branch keeps reporting both models in full, because that message is a human's decision aid and the qualifier is part of what they're deciding on. Comparison is family-level on both; only reporting differs.

## Scope correction — worth reading before approving

The issue was captured on the premise that a wording change "fixes every past and future stale payload at once." **That holds for the future half only**, and the PR says so out loud rather than shipping the overclaim.

A chip copies the preamble into its `prompt` **at spawn time**. The text is frozen there. Editing `chip-launching.md` rewrites the template, not payloads already sitting in a task list — a thread launched from an old chip reads the old wording and never consults this file. Replay is the same, and correctly so: it's pinned to the chip's own prompt so the printed block and the chip never diverge.

So this **closes the class going forward** rather than draining the backlog. The existing stale chips are [Issue #838](https://github.com/auerbachb/claude-code-config/issues/838), filed during this work, which also notes that `/pm` and `/wave` chips may not be enumerable across threads at all.

This surfaced during local review, not at capture. It does not change any acceptance criterion — every AC is about the preamble, the decision record, and the lint — but it does make one of the issue's Test Plan lines unverifiable; see the note under the Test plan below.

## Lint hardening

Three `require_pattern` assertions pin the new semantics, each anchored through enough of its sentence to survive a **semantic inversion** — a reword that keeps the vocabulary but reverses the rule (`"...qualifier: honor it on both sides"`). Matching the noun alone would pass exactly that. This applies the #802 lesson: assert the instruction, not the keyword.

The scope of that promise is stated in the script: these anchors catch an inversion written *into* the asserted sentence. No line-oriented pattern can catch a contradiction planted in a different sentence of the same file, and none of them claims to.

Six new test cases — three deletions, three inversions (keyword retained, meaning reversed).

## Test plan

- [x] A `**Model:**` line carrying a version qualifier, pasted into a thread of that same family, proceeds with a one-line confirmation and no stop
- [x] A `**Model:**` line naming a *different* family still stops, reports both models, and waits — unchanged
- [x] A bare family name matching the running family is the unchanged happy path
- [x] `bash .github/scripts/chip-model-guard-lint.sh` exits 0
- [x] `bash .github/scripts/tests/chip-model-guard-lint.test.sh` passes
- [x] Each of the three new assertions fails when the preamble fix is reverted (not merely when a sentence is deleted)
- [x] Each new test case fails when the fix is reverted, rather than passing vacuously
- [x] `grep -rl "MODEL GUARD: Your very first action"` returns exactly one file — no emitter restates the rule
- [x] `rule-lint`, `merge-authority-lint`, `verbatim-block-lint`, `skill-catalog-lint` all pass; rule corpus unchanged

> **Removed from the issue's Test Plan:** "click one of the genuinely old chips still sitting in the task list and confirm it launches into work instead of stopping." That is not achievable by this or any wording change, for the frozen-payload reason above — an old chip carries the old preamble. Tracked as [Issue #838](https://github.com/auerbachb/claude-code-config/issues/838) rather than checked off on a false basis.

## Verification

```
chip-model-guard-lint    OK (6 emitters)
guard-lint tests         OK (42 fixtures)
rule-lint                OK
merge-authority-lint     OK (5 emitters)
verbatim-block-lint      OK (6 copy surfaces)
skill-catalog-lint       OK (29 commands)
rule corpus              11,680 words (unchanged — both edited docs are under .claude/reference/, not auto-loaded)
```

Revert-proof, run against a scratch tree with the original preamble and the new lint — three errors, one per assertion, no collateral:

```
Missing required family-level comparison rule (expected /Compare families only —/)
Missing required version-qualifier-is-noise rule (expected /old-style version qualifier: ignore it on either side/)
Missing required family self-report in match branch (expected /Match \(same family\): state "Running on [{]FAMILY[}]/)
```

Running the new **test suite** against that same reverted preamble fails every new case through `mutate`'s no-op guard — the fixtures cannot pass vacuously.

## Local review coverage: `none`

Both CLIs went down mid-loop; **all findings they produced were resolved before this push.**

- **CodeAnt** — HTTP 403 on both the initial run and its one permitted retry (the known undocumented daily cap, not auth; per `cr-local-review.md` no `logout`/`login` was attempted). Exit 0 with an empty `issues[]` both times — a false clean caught by the stderr check.
- **CodeRabbit CLI** — three successful reviews (9 findings, all fixed: the replay-behavior contradiction, the overclaimed scope, and the semantic-inversion hardening), then the free-OSS tier cap on the confirming pass, with a 48-minute reset.

A self-review stood in for the fourth pass. Per `cr-local-review.md` this never satisfies the merge gate — and it doesn't need to: the CodeRabbit and CodeAnt **GitHub Apps** have quotas independent of their CLIs and gate this PR normally.

### DIFF
--- a/.claude/reference/chip-launching.md
+++ b/.claude/reference/chip-launching.md
@@ -79,18 +79,26 @@ The recommended model is worthless if nothing checks it at launch time. Every `p
 
 ```text
 MODEL GUARD: Your very first action — before any repo reads, file edits, or
-other tool calls — is to compare the model you are actually running as
-against the recommendation on the **Model:** line above.
-- Match: state "Running on {MODEL} as recommended." (one line) and proceed
-  immediately — no further prompts.
-- Mismatch, in EITHER direction (under- or over-powered): STOP. Do no other
-  work. Report, in one message, the model you are actually running as and the
-  model recommended above, then wait. Resume only on an explicit user reply
-  (e.g. "continue anyway"), proceeding on the current model — switching
-  models and relaunching is the recommended path instead.
+other tool calls — is to compare the model family you are actually running as
+against the family named on the **Model:** line above. Compare families only —
+`Opus`, `Sonnet`, `Haiku`, `Fable`. A trailing `<N>` or `<N>.<N>` after a
+family name is an old-style version qualifier: ignore it on either side, it is
+never evidence of a mismatch.
+- Match (same family): state "Running on {FAMILY} as recommended." (one line,
+  naming the family you are running — not the possibly-qualified string you
+  read) and proceed immediately — no further prompts.
+- Mismatch (different family), in EITHER direction (under- or over-powered):
+  STOP. Do no other work. Report, in one message, the model you are actually
+  running as and the model recommended above, then wait. Resume only on an
+  explicit user reply (e.g. "continue anyway"), proceeding on the current
+  model — switching models and relaunching is the recommended path instead.
 This is a best-effort self-report: no runtime API exists to introspect the
 active model, so the check relies on the model naming itself accurately.
 ```
+
+**Why family-level:** the guard exists to catch a thread running at the wrong *tier*, and tier is the family — a bare family name already resolves to the newest non-legacy model of that family ("Model and effort lines" above), so two versions inside one family are the same recommendation, and stopping between them reports a disagreement that does not exist.
+
+**What that reaches — and what it does not.** Every payload rendered from this file after the change is family-safe: a chip spawned from here on, a fallback block printed from here on. A chip **already offered** does not change — it froze a copy of the preamble into its `prompt` at spawn time, so it is still read under whatever wording it was emitted with, and editing this file cannot rewrite a payload sitting in a task list. Replaying such a chip reproduces that frozen wording too, and correctly so: replay is pinned to the chip's own `prompt` ("Print-on-demand replay" below), because a printed block that disagreed with the chip it describes would be the worse bug. The fix closes the class going forward rather than draining the existing backlog; that backlog is [#838](https://github.com/auerbachb/claude-code-config/issues/838). Rationale, scope limit, and the retired-family gap: `chip-model-guard-decision.md`.
 
 **Placement rule:** for every emitter, the `**Model:**` line is the first line of the `prompt` payload, the `**Effort:**` line is next, and this preamble is the content that immediately follows them — see each skill's Step for how its own template maps onto this shape.
 
--- a/.claude/reference/chip-model-guard-decision.md
+++ b/.claude/reference/chip-model-guard-decision.md
@@ -18,6 +18,26 @@ Issue #791 added an `**Effort:** {LEVEL} — {REASON}` line to the same unit, an
 
 Keeping the guard's promise narrow is what makes it worth trusting. If #735 ships `model`/`effort` parameters on `spawn_task`, both become tool-layer settings and the question is moot for chips — the guard reverts to a paste-flow safety net, exactly as this record's closing note anticipated.
 
+### Amendment (#837): the comparison is family-level
+
+**Family is the guarded axis; a trailing version number is noise.** The preamble compares the family the thread is running — `Opus`, `Sonnet`, `Haiku`, `Fable` — against the family named on the `**Model:**` line, ignoring any trailing version qualifier on either side. A cross-family difference still stops the thread cold in both directions, with the stop wording untouched.
+
+This follows from what the guard is *for*. It catches a thread running at the wrong **tier** — the judgment a chip encodes is "this work deserves this much model" — and tier is the family. Two versions inside one family are the same recommendation, so stopping on the difference between them reports a disagreement that does not exist. String equality was never the intended rule; it was simply the only reading the old wording left available.
+
+**The two branches report differently, on purpose.** The match branch names the *family* it is running, because echoing back a possibly-stale qualifier would assert the thread is running a version it is not — a one-line confirmation is no place for a claim that is false. The mismatch branch keeps reporting **both models in full**, qualifier included: that message exists to help a human decide whether to switch models or continue, and the qualifier is part of what they are deciding on. Reducing it to families there would strip detail at precisely the moment the most of it is wanted. So the asymmetry is not an oversight in the rewrite — comparison is family-level on both branches, and only the *reporting* differs.
+
+It composes with #791 rather than repeating it. #791 fixed which name gets **written** onto a `**Model:**` line — a bare family name, which resolves forward to the newest non-legacy model of its family on its own. This fixes what a thread **does** with a name it did not expect, a question none of #791's acceptance criteria reached, because every one of them is about what gets written into the corpus.
+
+**Scope limit — this does not retroactively fix chips already offered.** #837 was captured on the premise that a wording change "fixes every past and future stale payload at once." That holds for the future half only, and the correction is worth stating plainly because the premise is load-bearing: the preamble is copied into a chip's `prompt` **at spawn time**, so an already-offered chip carries a frozen copy of the old wording, is read under it, and never consults this file. Editing here cannot rewrite a payload sitting in a task list. What does change is that every payload rendered from this point — a chip spawned now, a fallback block printed now — is family-safe by construction. Replay is deliberately *not* in that list: it stays pinned to the chip's own `prompt` so the printed block and the chip never diverge (an invariant this record's opening already fixes), which means replaying an old chip faithfully reproduces its old guard wording. The class stops growing; the existing backlog is unaffected and is tracked as [#838](https://github.com/auerbachb/claude-code-config/issues/838).
+
+**Rejected as the primary fix: sweeping and re-emitting the stale chips.** A one-time cleanup does not answer a recurring problem, it needs a handle on every outstanding chip, and it does nothing for a paste block saved in someone's notes — so the reading rule had to change first regardless. With the recurring half now closed, a sweep is the right shape for the remaining one-time backlog *if* it proves big enough to be worth it, which is the call #838 makes.
+
+**Known gap — a retired family.** If a family is ever dropped from the fleet, a chip naming it would compare against nothing, and should stop. Not a live case today (the fleet has only gained families), so the preamble carries no wording for it; recorded here so a future retirement is a known open question rather than a surprise.
+
+**Terminology.** The preamble calls a trailing version number an *old-style* qualifier, not a *legacy* one. "Legacy" already means "a superseded model within a family" in `.claude/agents/README.md` ("the newest non-legacy model of that family"), and reusing it here would invite exactly the wrong reading — that only qualifiers naming an outdated version are ignorable. Every qualifier of the documented form is ignored, whichever version it names.
+
+**And only that form.** What the preamble discards is the numeric grammar it names — a trailing `<N>` or `<N>.<N>` — because that is what the pre-#791 lines actually carried. A non-numeric suffix is deliberately *not* covered: it is not a version number, nothing in the fleet's history produced one, and a rule that waved through any trailing token would quietly turn an unrecognized name into a match. If such a form ever appears, it should reach the mismatch branch and be looked at, like the retired-family case above.
+
 ## Rationale
 
 Two invariants existed before this ticket and a new preamble can preserve at most one of them:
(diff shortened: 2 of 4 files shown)


auerbachb:
@graphite-app re-review

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

PR #858, opened 2026-08-01
feat(#852): add the browser as rung 4 of the capability ladder

### LINKED ISSUE #852: Try the browser MCP before deferring — subagents included
## Background

The capability ladder in `.claude/rules/safety.md` already stops the agent from
saying "I can't" before it has actually looked: check the MCP tools and CLIs you
have, check whether the provider ships one, install it, and only then hand off a
runbook. That ladder currently ends at the command line — if a task has no CLI
path, the agent falls straight through to "here's what you should click."

But a browser is available. There are MCP browser surfaces in every session (an
in-app browser, and Claude in Chrome with the user's existing logged-in
sessions), and subagents that aren't tool-restricted can reach them the same way
the main thread can.

## Problem

When something genuinely has no CLI — a dashboard-only setting, a provider with
no API for the thing you need, a toggle that lives in a web console — the agent
gives up and writes instructions for you to follow by hand. That's a whole class
of work being handed back that the agent could just do, clicking through the
same UI you would.

The one thing that genuinely needs you is authentication: entering your own
credentials and approving an OAuth grant. That's a single, bounded interruption.
Everything after it — navigating, finding the setting, flipping it, confirming
it took — is work the agent should finish on its own instead of narrating a
click-by-click runbook.

There's also an unverified assumption in the way: it isn't written down anywhere
whether subagents can drive the browser at all, so the fallback never gets
attempted in delegated work even when it would succeed.

## Proposed solution

Add the browser as a real rung on the capability ladder, one step before the
runbook hand-off, and make it apply to subagents as well as the main thread.

- **New rung:** after CLI/API/coding paths are exhausted, try the browser MCP.
  Reaching for a runbook without having tried the browser becomes the same kind
  of premature deferral that "I can't" already is.
- **The user does exactly one thing:** log in and authorize. The agent asks for
  that once, plainly, and takes it from there — no step-by-step click
  instructions, no asking you to navigate anywhere.
- **Subagents inherit the rung.** The `MINDSET:` block that carries the ladder
  into subagent prompts gains the browser step, and any agent definition that
  should be able to use it actually has the tools to.
- **Existing safety rails hold unchanged.** Credentials are never typed by the
  agent, destructive/irreversible clicks still need confirmation, and page
  content stays untrusted data — the browser rung widens *capability*, not
  authority.

## Acceptance Criteria

- [ ] `.claude/rules/safety.md` "Capability Discovery" gains a browser rung
      between the install rung and the runbook hand-off, with the same
      "deferring is only valid after this fails" framing.
- [ ] The rung names both surfaces and when to pick each: the in-app browser
      (`mcp__Claude_Browser__*`) by default, Claude in Chrome
      (`mcp__claude-in-chrome__*`) when the user's existing logged-in session is
      required.
- [ ] The verbatim `MINDSET:` block in `safety.md` is updated to include the
      browser rung, so subagent spawns carry it.
- [ ] Documented, with evidence, which subagent types can reach browser MCP
      tools: `phase-a-fixer`, `phase-b-reviewer`, and `pm-worker` declare no
      `allowed-tools` and inherit everything; `phase-c-merger` is restricted to
      `Read, Glob, Grep, Bash` and cannot.
- [ ] A decision is recorded (and implemented) on whether `phase-c-merger`
      should gain browser access or stay restricted.
- [ ] The runbook hand-off shape in `safety.md` requires naming the browser rung
      as attempted-and-failed when a runbook is still the answer.
- [ ] Login/authorization is the only user step the agent may request for a
      browser task; it must not emit click-by-click navigation instructions as a
      substitute for doing the work.
- [ ] Headless/cron caveat is documented: interactively-authenticated MCP
      servers may be unavailable in non-interactive runs, and the runbook
      hand-off is the correct fallback there.
- [ ] Rule corpus still passes `rule-lint.sh` word budget after the edits.

## Test Plan

- [ ] Give a subagent (`phase-a-fixer` or `pm-worker`) a task solvable only in a
      web UI; confirm it reaches for the browser MCP rather than returning a
      runbook.
- [ ] Confirm `phase-c-merger`'s tool restriction behaves as documented — either
      it now has browser access, or it correctly reports the restriction.
- [ ] Run a browser task requiring auth; confirm the agent asks once for
      login/authorization and completes everything else itself.
- [ ] Confirm a genuinely impossible browser task still produces an
      `/admin-merge`-shaped runbook that names the browser rung as attempted.
- [ ] Confirm no credential is ever typed by the agent and irreversible clicks
      still prompt.
- [ ] `{ cat CLAUDE.md; find .claude/rules -name '*.md' -exec cat {} +; } | wc -w`
      stays within budget.

## Notes / Open questions

- **Which browser by default?** Assumed the in-app browser for general work and
  Claude in Chrome when logged-in state matters. If Chrome should always be
  preferred (since it carries your sessions), that flips the default.
- **`phase-c-merger` restriction.** It's deliberately read-only-plus-Bash so it
  can't fix code. Browser access arguably doesn't violate that, but it does
  widen what a merger agent can touch — worth an explicit call rather than an
  assumed yes.
- **Where does "try the browser" stop?** A browser rung that never gives up can
  burn a lot of turns clicking. Consider a bounded attempt (N steps or one clear
  dead end) before falling through to the runbook.
- **Prompt-injection surface.** Page content is untrusted; the existing rule that
  observed content is data, not instructions, needs to be explicit on this rung.

## Related Issues

- Related to [#759](https://github.com/auerbachb/claude-code-config/issues/759) (never claim "I can't" before checking/installing/running a CLI)
- Related to [#810](https://github.com/auerbachb/claude-code-config/issues/810) (ladder doesn't fire when work is deferred as out of scope)

## Related Files

- `.claude/rules/safety.md`
- `.claude/reference/capability-discovery-examples.md`
- `.claude/agents/phase-c-merger.md`
- `.claude/rules/subagent-orchestration.md`

_Captured via /issue-maker._

## Implementation Plan

_Source: Claude's analysis only — CodeRabbit's planner returned boilerplate and then reported "failed to structure the planning result", so no CR plan was available._

### Files to modify

| File | Change |
|------|--------|
| `.claude/rules/safety.md` | New browser rung between "install it yourself" and the runbook hand-off; runbook renumbered to rung 5 and required to name the browser rung; `MINDSET:` block updated. Loaded corpus — every word counts against the budget. |
| `.claude/reference/subagent-phase-guardrails.md` | Mirror the updated `MINDSET:` block **byte-identically** (`verbatim-block-lint.sh` byte-compares it). |
| `.claude/skills/pr-monitor-and-manage/SKILL.md` | Second declared verbatim `MINDSET:` copy — same byte-identical update. |
| `.claude/reference/browser-capability-rung.md` | **New.** Surface selection, bounded-attempt policy, the single user step, prompt-injection posture, subagent reachability matrix with evidence, `phase-c-merger` decision record, headless/cron caveat, runbook shape. |
| `.claude/reference/capability-discovery-examples.md` | Browser false walls vs real walls; correct the "interactive auth is always a wall" framing (CLI-initiated `<tool> login` still is; a dashboard login in the browser pane is not). |
| `.claude/reference/README.md` | Index entry for the new reference file. |
| `.claude/agents/README.md` | Paraphrased `MINDSET:` gains the browser rung; Agent Inventory table records browser reachability per agent. |
| `.claude/agents/phase-c-merger.md` | Record the restriction explicitly: no browser tools; report the restriction instead of emitting a click-by-click runbook. |

### Implementation steps

1. **safety.md — add rung 4 (browser), renumber the runbook to rung 5.** Rung 4 names both surfaces and when to pick each (`mcp__Claude_Browser__*` by default; `mcp__claude-in-chrome__*` only when the user's existing logged-in session is required), states that login/authorization is the only step the agent may ask for, forbids click-by-click navigation instructions as a substitute for doing the work, keeps credentials/irreversible-clicks/untrusted-page-content rails explicit, and bounds the attempt (one clear dead end → rung 5). Rung 5 must name the browser rung as attempted-and-failed.
2. **safety.md — update the `MINDSET:` block** so subagent spawns carry the browser rung.
3. **Mirror `MINDSET:` byte-identically** into `subagent-phase-guardrails.md` and `pr-monitor-and-manage/SKILL.md`.
4. **Write `.claude/reference/browser-capability-rung.md`** with the evidence-backed subagent matrix and the `phase-c-merger` decision.
5. **Update the remaining reference/agent files** (examples catalog, reference index, agents README, phase-c-merger).
6. **Budget pass:** run `bash .github/scripts/rule-lint.sh`; offset the added corpus words by compressing `safety.md` prose whose detail already lives in `capability-discovery-examples.md` — never by deleting an operative qualifier.
7. **Lints:** `rule-lint.sh`, `verbatim-block-lint.sh`, `merge-authority-lint.sh`, `skill-catalog-lint.sh`, plus the bash/python test suites.
8. **Dual-CLI local review** (`cr-local-review.md`), then push + PR.

### Decision to be recorded — `phase-c-merger`

**Stay restricted** (`allowed-tools: Read, Glob, Grep, Bash`, unchanged). Phase C verifies the gate and AC and then runs `/wrap`; every step of that is `gh`/`git`, so a browser buys it nothing. Granting browser tools would let a merger agent reach the GitHub web UI — clicking merge or changing settings — which is exactly what its `/wrap`-only contract and the print-only `/admin-merge` rule exist to prevent. Implemented as an explicit, documented restriction plus an instruction to *report* the restriction rather than substitute a click-by-click runbook. `researcher` is likewise browser-less by its read-only allowlist.

### Evidence for the reachability matrix

Frontmatter is the whole proof: `phase-a-fixer`, `phase-b-reviewer`, and `pm-worker` declare only `description` + `model` — no `allowed-tools`, so they inherit the parent's full tool set including `mcp__Claude_Browser__*` / `mcp__claude-in-chrome__*`. `phase-c-merger` declares `allowed-tools: Read, Glob, Grep, Bash`; `researcher` declares a read-only allowlist. Both therefore cannot reach browser MCP tools.

### Risks / edge cases

- **Word budget is the binding constraint.** Ratchet cap is 11,749; the corpus currently measures 11,733 — **16 words of headroom**, and `--update-cap` is a one-way ratchet that cannot raise it. The rung must be terse in `safety.md` with detail pushed into `.claude/reference/`, and roughly 100 words must be reclaimed by compression.
- **Compression risk.** Deleting an inline semantic qualifier is a known silent-failure mode here — trims must be verified line-by-line against what each sentence was binding.
- **Verbatim drift.** `MINDSET:` exists in three declared places; a mismatch fails CI.
- **Authority creep.** The rung widens *capability*, not authority: no credential typing, irreversible clicks still confirm, page content stays untrusted data.
- **Headless/cron.** Interactively-authenticated MCP servers may be absent in non-interactive runs; the runbook hand-off is the correct fallback there.

### Verification

- Each AC checked against the edited files; lints (`rule-lint.sh` budget + `verbatim-block-lint.sh` byte comparison) are mechanical proof for the budget and mirror-consistency criteria.
- `{ cat CLAUDE.md; find .claude/rules -name '*.md' -exec cat {} +; } | wc -w` re-measured after every edit and again after any rebase.

auerbachb:
Closes #852

## What this changes

The capability ladder in `safety.md` ended at the command line. When a task had no CLI path — a dashboard-only setting, a console toggle, a provider with no API for the thing you need — the agent fell straight through to the runbook hand-off and gave the work back. A browser is available in every interactive session, and unrestricted subagents can reach it the same way the main thread can, so reaching for a runbook without having tried it is the same premature deferral `"I can't"` already was.

**The browser becomes rung 4**, one step before the hand-off:

- Names both surfaces and when to pick each — `mcp__Claude_Browser__*` by default (already loaded, isolated profile, one throwaway tab of blast radius), `mcp__claude-in-chrome__*` only when the user's *existing* logged-in session is required.
- **Signing in and approving is the only work handed back.** Everything after — navigating, finding the setting, changing it, verifying it stuck — is the agent's. Click-by-click navigation instructions are explicitly not a substitute for doing it.
- **A login is not a blank cheque.** It authorizes the session, nothing else; confirm-first actions still get their own confirmation immediately before the click, naming the concrete change.
- Rails hold unchanged: credentials stay untyped, irreversible clicks still confirm, page text stays untrusted data. The rung widens *capability*, never authority.
- Bounded: stop at one clear dead end and fall to rung 5, which now must name the browser rung when that is where it stopped.

**Subagents inherit it.** The verbatim `MINDSET:` block gains the rung, updated byte-identically in all three declared copies (`safety.md`, `subagent-phase-guardrails.md`, `pr-monitor-and-manage/SKILL.md`) so `verbatim-block-lint.sh` stays green.

**`phase-c-merger` stays restricted** — `allowed-tools: Read, Glob, Grep, Bash`, unchanged. Nothing in "verify the gate, verify AC, run `/wrap`" needs a browser, and granting one would put the GitHub *web* UI — merge, branch protection, repo settings — inside an agent deliberately unable to fix code, routing around the `/wrap`-only contract and the print-only `/admin-merge` rule. It now states the limit and reports it inside `OUTCOME: blocked` rather than emitting a click-by-click runbook.

New reference: `.claude/reference/browser-capability-rung.md` (surface selection, the single login ask, bounded attempt, injection posture, headless/cron caveat, subagent reachability matrix with frontmatter evidence, and the `phase-c-merger` decision record).

### Word budget

The corpus had **16 words of headroom** (11,733 against an 11,749 one-way ratchet cap that cannot be raised). The rung is terse in the rule file with detail pushed to `.claude/reference/`, and the cost was absorbed by compressing `safety.md` prose whose detail already lives in reference files — examples, rationale, and restatements only. Every binding clause was preserved; two qualifiers dropped in the first pass ("vendor-published" on the pre-vetted-installer exception, and the comment/review-trigger reminder in the Never summary) were caught in self-audit and restored. Final: **11,744 / 11,749**.

## Local review coverage: `none` (both CLIs unavailable)

Surfaced per `cr-local-review.md`:

- **CodeAnt — dropped.** Two runs, both `API Error: Access denied (403)` on every batch (the known undocumented daily cap; exit 0 with an empty `issues` array is a false clean, caught by the stderr check). Per the rule, no `codeant logout`/`login` was attempted.
- **CodeRabbit — one good run, then rate-limited.** The first run completed and produced **6 findings**; the verification re-run returned `errorType: rate_limit` (free-OSS tier, 51-minute reset), which counts as a failed run, so there is **no clean CLI pass**. A self-review of the full diff stands in.

The GitHub App reviewers are independent of the CLIs and gate this merge normally.

### CodeRabbit findings — 4 fixed, 2 declined

| # | Finding | Disposition |
|---|---------|-------------|
| 3 | `browser-capability-rung.md`: login must not read as authorization for later irreversible actions | **Fixed** — added "a login is not a blank cheque", per-action confirmation example, and "one yes does not carry to the next click"; the example no longer implies sign-in approves the scaling change |
| 4 | `cli-tool-defaults.md`: rung 4 applies to dashboard operations; CLI-initiated auth stays rung 5 | **Fixed** — spelled out that `railway login` / `codeant login` open the system browser out of reach of any MCP surface |
| 5 | `codeant-graphite-supplemental.md`: stale rung-3 auth classification next to the rung-5 runbook | **Fixed** — reclassified as a rung-5 wall (also corrected the matching standing-state memory note) |
| 2 | `safety.md`: Always/Never summary should acknowledge the template exception | **Fixed** — "untouchable (recognized templates excepted)" |
| 1 | `safety.md`: forbid passing generated secrets as provider CLI *arguments*; require stdin | **Declined** — reverses a documented repo decision, not a regression in this diff. `capability-discovery-examples.md` already states the boundary explicitly ("`railway variables --set` takes it as an argument … where a CLI offers a stdin path, prefer it"); `railway variables --set` has no stdin path, so the requirement would make issue #759's motivating case undoable. Out of scope for #852 — worth its own issue if the trade-off should be revisited. |
| 6 | `agents/README.md`: example prompt's SAFETY block should be the full canonical text | **Declined** — that file is a *paraphrased* surface, deliberately excluded from byte comparison by `verbatim-block-lint.sh`'s own header ("abbreviated/reworded versions … as prompt examples"). The canonical blocks live in `safety.md` and `subagent-phase-guardrails.md`; expanding the illustration would duplicate them without adding enforcement. |

## Test plan

- [x] `safety.md` "Capability Discovery" has a browser rung between the install rung and the runbook hand-off, carrying the same "deferring is only valid after this fails" framing
- [x] The rung names both surfaces and when to pick each — `mcp__Claude_Browser__*` default, `mcp__claude-in-chrome__*` when the user's existing logged-in session is required
- [x] The verbatim `MINDSET:` block includes the browser rung, so subagent spawns carry it — `bash .github/scripts/verbatim-block-lint.sh` passes across all 6 copy surfaces
- [x] Subagent browser reachability is documented with evidence: `phase-a-fixer`, `phase-b-reviewer`, `pm-worker` declare no `allowed-tools` and inherit everything; `phase-c-merger` (`Read, Glob, Grep, Bash`) and `researcher` cannot. Reproduce with the frontmatter probe in `browser-capability-rung.md` §Subagent reachability
- [x] The `phase-c-merger` decision is recorded **and** implemented: frontmatter unchanged, and the agent is instructed to report the restriction rather than emit a click-by-click runbook
- [x] Rung 5 requires naming the browser rung as attempted-and-failed when a runbook is still the answer
- [x] Login/authorization is the only user step the rung may request, and click-by-click navigation instructions are explicitly forbidden as a substitute for doing the work
- [x] Headless/cron caveat documented: interactively-authenticated MCP servers may be unavailable in non-interactive runs, and the runbook hand-off is the correct fallback there
- [x] Rung numbering is internally consistent everywhere the ladder is referenced (rules, agent definitions, reference docs) — no stale "rung 4 = hand-off"
- [x] `bash .github/scripts/rule-lint.sh` passes the word budget: 11,744 ≤ 11,749 ratchet cap
- [x] `merge-authority-lint.sh`, `skill-catalog-lint.sh`, and the `rule-lint` / `verbatim-block-lint` test suites pass

> **Note on the issue's live-behavior test items.** Issue #852's Test Plan includes items that require actually spawning a subagent against a web-only task and running an authenticated browser task. Those were not exercised in this session (no subagent was spawned), so they are not checked off above; what *is* verified is the mechanism that produces the behavior — the tools the agents inherit and the block that reaches their prompt. Live behavioral verification is a natural follow-up once a real web-only task comes along.


auerbachb:
@graphite-app re-review

auerbachb:
Rebased onto main (2 commits) and batched a doc fix caught during acceptance verification: `browser-capability-rung.md` still described the old ladder's hand-off as "rung 4", which is now the browser rung — it now names the runbook hand-off and gives both numbers.

`BEHIND` is cleared. Corpus re-measured after the rebase (11,744 / 11,749 cap — no drift); `rule-lint`, `verbatim-block-lint`, `merge-authority-lint`, `skill-catalog-lint` and all four lint test suites pass on the new SHA. All 11 Test Plan boxes verified against the code and checked off.

auerbachb:
@coderabbitai full review

auerbachb:
@graphite-app re-review

auerbachb:
@cursor review

Issue #732, opened 2026-07-23
Pipeline ceilings and merge offers must count only my own PRs


auerbachb:
## Background
When new coding chips are about to launch, the orchestrator counts the PRs already in flight and holds the wave when the pipeline ceiling is hit. It can also offer to "clear slots" — running a merge-gate pass over the open PRs and asking for merge authorization. Today (Jul 23, 2026) that flow fired in a shared repo: six PRs were open, every one of them authored by a collaborator (Facundo), none by me.

## Problem
Two failures stacked on top of each other:

- **The ceiling counted PRs that aren't mine.** My chip wave was blocked ("6 open PRs ≥ the 4-PR pipeline ceiling") when my own in-flight count was zero. A collaborator's backlog silently froze my work.
- **The recommended option was to merge other people's PRs.** "Clear slots first" proposed a merge-gate pass plus merge authorization over all six — and I approved it before anyone checked whose PRs they were. Only my follow-up question ("are these PRs mine or Facundo's?") surfaced that the standing hands-off rule applied and the whole option was moot.

The hands-off rule existed the entire time; nothing consulted it until a human got suspicious.

## Proposed solution
Authorship becomes a first-class input, checked at discovery time — before anything is counted or offered:

- Ceiling and slot math (chip gating, wave sizing, "N open PRs" summaries) counts only PRs I authored. Other people's PRs never enter the denominator; at most they get mentioned as FYI context.
- Any offer that would act on PRs — merge-gate passes, clear-slots proposals, babysit or fix-dispatch menus — is built exclusively from my PRs. If zero of my PRs qualify, the offer simply isn't made.
- Ceiling-block messages say whose PRs were counted, so a wrong count is visible at a glance instead of needing a suspicious question.

## Acceptance Criteria
- [ ] Every place open-PR counts gate chip/wave launches (pipeline-ceiling logic in `.claude/reference/chip-launching.md`, `/wave`, `/pm`, `/prompt`, and the 3-4 active-PR cap in `.claude/rules/subagent-orchestration.md`) derives the count from PRs authored by the authenticated user (`gh pr list --author "@me"` or equivalent), never from all open PRs.
- [ ] Merge-management offers and actionable PR lists (merge-gate pass proposals, `/pr-monitor-and-manage` dispatch decisions, babysit suggestions) enumerate only user-authored PRs; collaborator PRs never appear as actionable options.
[...]

PR #878, opened 2026-08-01
feat(#873): claim an issue at pick time so two threads can't work it at once

### LINKED ISSUE #873: Claim an issue at pick time so two threads can't work it at once
## Background

Work gets picked up from a lot of places: a `/pm` thread running issues inline, a click-to-launch chip, `/start-issue` in a fresh tab, or you simply telling a session "go work on #167." Each of those threads is independent, and none of them can see what the others are doing. The only thing they share is GitHub.

That mostly works, because every entry path does check whether an issue is already being handled before starting on it.

## Problem

The check looks for the wrong thing. It asks "does an open PR exist for this issue?" — and a PR is the *last* artifact a thread produces, not the first. Between the moment a thread picks an issue and the moment its PR appears, there is a plan-and-code window that routinely runs half an hour or more, and for that entire window every other thread's check comes back clean. So a second thread picks the same issue in good faith, and the collision only becomes visible when the second PR shows up under your account and someone notices the issue number looks familiar.

That is exactly what happened: two threads worked the same issue at the same time, both correctly ran the check they were told to run, and the duplication surfaced only after both had spent a full pipeline on it. The cost isn't a merge conflict — it's two sessions of planning, coding, CI, and reviewer quota buying one issue's worth of progress, plus a PR that now has to be thrown away.

Nothing claims an issue when work *starts*. GitHub already has the primitives for this — an issue can be assigned, and it can be labeled — and the workflow uses neither, so the one store every thread on every machine can see stays silent about who is doing what until it's too late to matter.

## Proposed solution

Stake the claim at the moment work begins instead of when it finishes. As soon as a thread commits to an issue — before planning, before the worktree — it marks the issue as being worked on GitHub, so any other thread that looks a minute later sees it immediately. The check every entry path already runs gets widened to read that claim, not just open PRs, and reports it the same way it reports an existing PR today: "issue #167 is already being worked — skipping."

Claims have to let go on their own, or a thread that dies mid-task poisons the issue forever. So a claim clears when the work reaches a real terminal state, and goes stale on its own after a while of no movement — a stale claim is a warning to be surfaced, not a permanent block on the issue.

The result is that starting the same work twice becomes something you'd have to do deliberately, instead of the default outcome of two tabs and bad timing.

## Acceptance Criteria

- [ ] A shared claim helper (e.g. `.claude/scripts/issue-claim.sh <N> --claim|--release|--check`) writes and reads the claim through GitHub so it is visible across threads, sessions, and machines — no local-only state file, which by definition cannot see a sibling thread
- [ ] Claim is taken at **pick** time — before CR-plan polling, before worktree creation — not at PR-open time
- [ ] Every entry path claims: `/start-issue`, `/subagent` (Step 6.0), chip-launched coding threads, and an ad-hoc "work on #N" request
- [ ] Every entry path **checks** the claim alongside its existing open-PR check, and skips with a reason naming the claim; `/subagent` Step 6.0 and `/wave` Step 2.3 both consult it
- [ ] Claim releases on terminal states — PR merged, issue closed, or work explicitly abandoned
- [ ] Stale claims expire after a bounded window of no activity and are surfaced as a warning rather than silently blocking the issue forever
- [ ] Re-claiming an issue you already hold is a no-op, not an error (resumed threads, post-compaction recovery)
- [ ] A claim never blocks the human: an explicit "start it anyway" in chat overrides, and says it is overriding
- [ ] Offline bash unit test covering claim / duplicate-claim / release / stale-expiry, running in the hook-scripts CI job

## Test Plan

- [ ] Thread A claims #N; thread B's pre-start check reports #N as claimed and does not start — with no PR existing anywhere
- [ ] Thread A claims and then merges; the claim is released and #N is startable again
- [ ] Claim older than the stale window → surfaced as stale, work is allowed to proceed
- [ ] Same thread re-claims #N after a compaction-recovery restart → no-op, no error, no duplicate marker
- [ ] Explicit chat override starts work on a claimed issue and states that it is overriding
- [ ] `/wave` and `/subagent` both exclude a claimed-but-PR-less issue from their candidate sets, naming the claim as the reason
- [ ] Stubbed-`gh` offline test passes in CI

## Notes / Open questions

- **Claim mechanism:** self-assign (`gh issue edit --add-assignee @me`) reads naturally and needs no new label, but it is a single-bit signal — it can't distinguish "claimed by a thread five minutes ago" from "assigned to me as backlog." A dedicated `in-progress` label (the repo has no such label today) is more explicit and easier to expire. Possibly both: assignee for ownership, label for activity.
- **Where the timestamp lives:** staleness needs a "claimed at," which an assignee alone doesn't carry. A short claim comment, or the label's `labeled_at` from the issue timeline, are the two obvious homes.
- **Stale window:** long enough to survive a slow CR plan plus a long coding phase, short enough that a dead thread doesn't park an issue for a day. A few hours is the likely starting point.
- **Single-machine near-miss:** `/start-issue` already refuses when a worktree for `issue-N-*` exists, which catches some of this — but only for that one entry path, only on the same machine, and only after the worktree stage. It is not a substitute for a claim other threads can see.
- **Deliberately out of scope:** file-level overlap between *different* issues. That is the launch-side serialization already covered by #756; this issue is about two threads on the *same* issue, which no overlap filter addresses.

## Related Issues

- Related to #756 (serializes different issues that share a file; this covers the same issue picked twice)
- Related to #636 (`/wave`'s launch-time filtering, whose dedup this strengthens)
- Related to #732 (claim scope must respect the authorship rules — a collaborator's claim is context, not something to overwrite)

_Captured via /issue-maker._

## Implementation Plan

_Merged: Claude's analysis + CodeRabbit's plan. CR's three-phase shape is kept; four design points are changed, each called out below with its reason._

### Design deltas from the CodeRabbit plan

1. **A claim's holder must be finer-grained than the GitHub login.** CR defines `mine` as "`in-progress` label present **and** viewer in assignees". In the failure this issue was filed for, both threads authenticate as the *same* login — one human, several Claude tabs — so that rule returns `mine` to the second tab and it starts anyway. The claim therefore records a **holder token**, and `mine` means *this thread*, not *this account*. Without it, Test Plan row 1 ("thread B ... does not start") cannot pass.
2. **The holder token needs somewhere to live.** Labels carry a name and assignees carry a login; neither can hold an arbitrary token. So the claim adds a **claim comment** with a machine-readable marker. This also supplies the "claimed at" timestamp directly. CR's Design Choice 2 (the `labeled` timeline event) is kept as the **fallback** timestamp for a label applied by hand with no claim comment.
3. **One corpus home, not four.** The auto-loaded rule corpus is at **11,744 words against a ratchet cap of 11,749** — five words of headroom, and `rule-lint.sh`'s ratchet only lowers the cap unless `--allow-raise` is used, which `CLAUDE.md` restricts to "after intentional cuts". CR's plan writes prose into `CLAUDE.md`, `issue-planning.md`, `safety.md`, and `chip-spawn.md`. Instead the rule text lands in **`issue-planning.md`** (which already owns "before any code work") plus a short insert into `CLAUDE.md`'s flow line, paid for by equivalent cuts elsewhere in the corpus; mechanism and rationale go to a new non-loaded `.claude/reference/issue-claim.md`. The override sentence stays **next to** the rule it limits rather than in `safety.md`, so a loaded rule never has its limiting sentence in a non-loaded file.
4. **`/wave` reads the claim in one call, not N.** Per-candidate `issue-claim.sh --check` over a 30-issue backlog is ~90 API calls. `/wave` instead runs a single `gh issue list --label in-progress --json number` and calls the helper only for candidates in that (usually tiny) intersection.

### Files to create

- `/Users/brettonauerbach/Documents/Develop/claude-code-config/.claude/scripts/issue-claim.sh`
- `/Users/brettonauerbach/Documents/Develop/claude-code-config/.claude/scripts/tests/issue-claim.test.sh`
- `/Users/brettonauerbach/Documents/Develop/claude-code-config/.claude/reference/issue-claim.md`

### Files to modify

- `.claude/skills/start-issue/SKILL.md` — claim at pick time (new Step 2b)
- `.claude/skills/subagent/SKILL.md` — Step 6.0 check + claim
- `.claude/skills/wave/SKILL.md` — Step 2 claimed-issue exclusion
- `.claude/skills/wrap/SKILL.md` — release after Step 2.4 squash merge
- `.claude/scripts/admin-merge.sh` — release after each `FINAL_MERGED == true`
- `.claude/reference/chip-launching.md` — claim as the launched thread's first action; release on stale-chip trigger 4
- `.claude/rules/issue-planning.md` — the claim rule, stale handling, override
- `CLAUDE.md` — name the claim in the canonical flow line
- `.claude/reference/README.md` — index the new reference file (if it indexes)

### Phase 1 — the shared primitive

**`issue-claim.sh <N> --claim|--release|--check [--repo owner/repo] [--holder ID] [--json] [--allow-claimed]`**

*Claim artifacts, all on GitHub (no local claim state):*

| Artifact | Carries |
|---|---|
| `in-progress` label | the coarse "someone is working this" bit, visible in issue lists |
| assignee `@me` | ownership at login level (for #732 collaborator respect) |
| claim comment `<!-- claude-claim: {"holder":..,"login":..} -->` | holder identity + `created_at` timestamp |

*Verdicts (`--check`):*

- `unclaimed` — no label and no claim comment
- `mine` — claim comment holder equals this holder (any age)
- `stale` — foreign holder, `now - claimed_at > CLAIM_STALE_HOURS` (default 4)
- `claimed` — foreign holder, fresh
- `unknown` — any indeterminate `gh` result (viewer, issue, comments, or timeline read failed)

*Timestamp:* claim comment `created_at`; fall back to the most recent `labeled` event for `in-progress` from `repos/{owner}/{repo}/issues/{N}/timeline` when the label exists without a comment.

*Holder resolution:* `--holder` → `$CLAUDE_CLAIM_HOLDER` → `$CLAUDE_SESSION_ID` → `<hostname>:<git toplevel>`. The last is honest about its granularity: two threads sharing one checkout are one holder. That is acceptable because `CLAUDE.md` mandates a worktree per thread, so the mandated workflow always yields distinct holders; the header documents the limit rather than hiding it.

*Actions:*

- `--claim` — `mine` is a no-op success (resumed threads, post-compaction recovery: no second comment, no duplicate marker). `unclaimed`/`stale` create the `in-progress` label idempotently (`gh label list` existence check, then `gh label create ... 2>/dev/null || true`), add the label, add `@me`, and post/re-stamp the claim comment (stale takeover deletes the previous holder's comment first). Fresh foreign claim refuses unless `--allow-claimed`, and with it prints an explicit override line.
- `--release` — idempotent; removes the label, removes **only the viewer's own** assignment, deletes **only this holder's** claim comment. Never touches a collaborator's assignment or comment (#732).
- Exit codes mirror `pr-authorship.sh`: `0` startable/authorized (`unclaimed`, `mine`, `stale`), `1` blocked (fresh foreign claim), `2` usage, `3` not found, `4` unknown. **Fail closed** — callers treat `unknown` exactly as `claimed`.
- Default stdout: the one-word verdict. `--json`: `{issue, repo, viewer, holder, verdict, claimant, claimed_at, stale, reason}`. Human-readable refusal/warning lines go to stderr, as in `pr-authorship.sh`.
- Header comment states: GitHub is authoritative; stale is a warning, never a permanent block; `--allow-claimed` is only ever an explicit per-issue chat override — never a config default, never inferred, never carried forward.
- Log the invocation to `$HOME/.claude/script-usage.log` like its siblings.

**`tests/issue-claim.test.sh`** — follows `pr-authorship.test.sh` conventions (`set -uo pipefail`, isolated `TMP`/`$HOME`, stub `gh` on `PATH` dispatching on `$1`/`$2`, PASS/FAIL counters, `run()` capturing `OUT`/`RC`, summary exiting non-zero on failure). Stub state is env-driven and timestamps are computed relative to now for a deterministic stale boundary. Cases: claim on unclaimed writes label + assignee + comment; second holder sees `claimed`/exit 1 **with no PR anywhere**; same holder re-claims → no-op, exit 0, no duplicate marker; release returns `unclaimed`; age past `CLAIM_STALE_HOURS` → `stale`, exit 0; `--allow-claimed` overrides a fresh foreign claim and says so; not found → exit 3; `gh` failure → `unknown`/exit 4 and never `unclaimed`; release leaves a collaborator's assignee intact. Auto-discovered by `.github/scripts/run-hook-tests.sh` — no workflow edit.

### Phase 2 — claim at every pick-time entry path

1. **`/start-issue`** — new step between Step 2 (read issue) and Step 3 (CR-plan polling), so the claim precedes both planning and the worktree. `claimed`/`unknown` → stop, naming the claim, in the same shape as the existing worktree-skip. `stale` → surface the warning and proceed. The `git worktree list` guard stays as the same-machine backstop.
2. **`/subagent` Step 6.0** — run `issue-claim.sh {NUMBER} --check` alongside the existing `gh pr list --search "head:issue-{NUMBER}"`; either signal skips the issue, with the message naming which one fired. On pass, `--claim` before spawning Phase A so the claim is held for the whole pipeline.
3. **Chip-launched threads** — `chip-launching.md`'s payload template gains the claim as the launched thread's first action *after* the MODEL GUARD preamble and before any repo read. Claim at click time, never at offer time (offering is not launching). This does not disturb the `**Model:**`-first / `**Effort:**` / no-blank-line ordering that `chip-model-guard-lint.sh` enforces.
4. **Ad-hoc "work on #N"** — `issue-planning.md` gains the rule (with the stale and override behavior); `CLAUDE.md`'s flow line names the claim so the canonical flow reads `issue → claim → CR plan → …`.
5. **`/wave` Step 2** — new exclusion after item 3, batched per delta 4. Drop `claimed`/`unknown` candidates naming the claim; `stale` is not excluded but is surfaced. A collaborator's fresh claim excludes the issue as context and **never** counts toward the viewer's own `IN_FLIGHT` ceiling and is never overwritten (#732).

### Phase 3 — release on terminal states

1. **`/wrap`** — release immediately after Step 2.4's `gh pr merge --squash` succeeds, reusing the `ISSUE_N` that Step 3.1 already resolves via `pr-issue-ref.sh`. Only here, so Phase C inherits it without duplication.
2. **`admin-merge.sh`** — release after each `FINAL_MERGED == true` confirmation (all three merge shapes), resolving the issue with `pr-issue-ref.sh`. Never let a release failure fail an already-successful merge.
3. **Issue closed / abandoned** — `chip-launching.md` "Stale-chip hygiene" trigger 4 releases alongside `dismiss_task` when an issue closes without a merged PR; the entry-path skills note that an explicit human stop releases too.
4. **Override contract** — recorded beside the rule in `issue-planning.md`: the only way to start work on an issue claimed elsewhere is the user naming that issue in chat; per-issue, per-session, never inferred, never a default, and the tool must say it is overriding. `unknown` never reads as permission.

### Risks / edge cases

- **Corpus budget.** Five words of headroom. Every corpus edit is measured with `{ cat CLAUDE.md; find .claude/rules -name '*.md' -exec cat {} +; } | wc -w` and paid for by cuts before push; cuts are self-audited for dropped qualifiers, not just word count.
- **Same-login threads** — the reason for the holder token (delta 1). A login-only comparison silently disables the whole feature.
- **`unknown` must never read as unclaimed** — fail-closed at the script and again at every call site.
- **Label sprawl** — `in-progress` is created idempotently; a pre-existing label is reused, never recreated.
- **Collaborator claims (#732)** — release only ever removes the viewer's own assignment and the viewer's own claim comment.
- **Release must not break a merge** — a failed release after a successful merge is a warning, never a non-zero exit from the merge path.
- **Comment noise** — one claim comment per active work window, deleted on release.
- **`gh` rate cost** — `/wave` batches (delta 4); other paths call `--check` once per issue.

### Verification

| AC | How |
|---|---|
| Shared GitHub-backed helper | `issue-claim.sh` reads/writes only GitHub; no local claim file — grep the script |
| Claim at pick time | `/start-issue` step sits before CR-plan polling and worktree creation; `/subagent` claims before spawning Phase A |
| Every entry path claims | four call sites: `/start-issue`, `/subagent` 6.0, `chip-launching.md` template, `issue-planning.md` ad-hoc rule |
| Every entry path checks | `/subagent` 6.0 and `/wave` Step 2 both consult it; each skip names the claim |
| Release on terminal states | `/wrap` post-merge, `admin-merge.sh` post-merge, chip trigger 4 on close, explicit abandonment |
| Stale expiry surfaced, not blocking | `--check` returns `stale` with **exit 0**; test asserts both |
| Re-claim is a no-op | test asserts exit 0 and no duplicate marker |
| Override never blocks the human | `--allow-claimed` test asserts success **and** the override statement in output |
| Offline test in CI | `bash .claude/scripts/tests/issue-claim.test.sh` locally; auto-discovered by `run-hook-tests.sh` in the `hook-tests` job |

auerbachb:
Closes #873

## What changed

Work was guarded by the wrong question: **"does an open PR exist for this issue?"** A PR is the *last* artifact a thread produces, so for the entire plan-and-code window — routinely 30+ minutes — every other thread's check comes back clean, and a second thread picks the same issue in good faith.

This stakes the claim at **pick** time instead.

`.claude/scripts/issue-claim.sh <N> --check|--claim|--release` writes the claim to GitHub as three artifacts (`in-progress` label + `@me` assignee + a claim comment carrying the holder token and timestamp). No local claim state — a local file by definition cannot see a sibling thread.

| Path | Wiring |
|---|---|
| `/start-issue` | new **Step 2b** — before CR-plan polling, before the worktree |
| `/subagent` | Step 6.0, alongside the existing open-PR check; `--claim` before spawning Phase A |
| chip-launched thread | first action after the MODEL GUARD preamble (`chip-launching.md`) |
| ad-hoc "work on #N" | `issue-planning.md` **step 0** |
| `/wave` | Step 2 item 4, batched to one `gh issue list` for the whole backlog |
| `/wrap`, `admin-merge.sh`, chip trigger 4 | `--release` on the terminal states |

## Three design calls worth reviewing

**1. The claim records a holder token, not just the login.** This is a deliberate departure from the CodeRabbit plan, which defined `mine` as "label present **and** viewer in assignees". The failure this feature exists to prevent is *two Claude tabs belonging to the same human* — both authenticate as the same login, so an assignee-only test answers "yes, mine" in the second tab and blocks nothing. The feature would have been inert in exactly its motivating case. `issue-claim.test.sh` scenario (2) pins this: substituting the login-level comparison makes it fail (verified by mutation, below).

**2. Block at holder level, release at account level.** `--check`/`--claim` compare holders, so a sibling thread is blocked. `--release` compares logins, so a terminal state reached by a *different* thread of yours — a Phase C merger, a `/wrap` from another worktree — still clears the claim instead of leaking it. A collaborator's claim is never touched by either (#732), including under `--allow-claimed`: the override authorizes starting the work, not deleting someone else's comment.

**3. Rule-corpus cost paid in place, no ratchet-cap raise.** The corpus was at 11,744 against a cap of 11,749 — five words of headroom. Rather than write into four rule files (CR's plan) or reset the ratchet with `--allow-raise` (which would jump the cap by 776, not by the ~30 actually needed), the rule text lands in one place and is paid for by moving mechanism out: the fetch-concatenate-edit snippet → `/start-issue` Step 5 (which carries a fuller version including the duplicate-section strip), the CR-plan filter detail → `cr-plan.sh --help`, and everything else → the new non-loaded `.claude/reference/issue-claim.md`. **Final: 11,749 / 11,749.**

## Local review coverage: `none` (self-review)

Both local CLIs were unavailable, so this was pushed on a self-review per `cr-local-review.md`:

- **CodeAnt** — `Access denied (403)` on both the full run and the scoped retry (the known undocumented daily cap; no `logout`/`login` attempted, per the rule). `meta.capped: true`, zero review records.
- **CodeRabbit CLI** — hung 10 minutes in `setting_up` after `Free OSS limits apply`, well past the 2-minute drop threshold. No findings emitted.

The GitHub App reviewers are unaffected and still gate the merge.

The self-review did find one real defect, now fixed with a regression test: the takeover path deleted the previous claim comment unconditionally, which would have deleted a **collaborator's** comment under `--allow-claimed` — a #732 violation. It now deletes only the viewer's own markers; reads take the newest claim comment, so a leftover foreign marker is inert.

## Verification performed

- `issue-claim.test.sh` — **75 assertions, 0 failures**, offline stubbed `gh` with a *stateful* fake issue so a claim written by one run is read back by the next.
- **Mutation-tested**, because a test that passes both ways proves nothing:
  - login-level `mine` (CR's design) → scenario (2) fails ✔
  - staleness disabled → scenario (5) fails ✔
- `run-hook-tests.sh` — **53 suites, 0 failed**; the new test is auto-discovered (no workflow edit).
- `admin-merge.test.sh` — 81 passed, 0 failed, after adding the release helper to all three merge shapes.
- All five lints green: `rule-lint`, `chip-model-guard-lint`, `merge-authority-lint`, `skill-catalog-lint`, `verbatim-block-lint`.
- The `awk` epoch converter (used instead of `date -d`/`date -j`, which disagree across GNU/BSD) checked against Python at the epoch boundary, a leap day, and a year end — exact on all four.

## Test plan

- [x] Thread A claims #N; thread B's pre-start check reports #N as claimed and does not start — with no PR existing anywhere
- [x] Thread A claims and then merges; the claim is released and #N is startable again
- [x] Claim older than the stale window → surfaced as stale, work is allowed to proceed
- [x] Same thread re-claims #N after a compaction-recovery restart → no-op, no error, no duplicate marker
- [x] Explicit chat override starts work on a claimed issue and states that it is overriding
- [x] `/wave` and `/subagent` both exclude a claimed-but-PR-less issue from their candidate sets, naming the claim as the reason
- [x] Stubbed-`gh` offline test passes in CI
- [x] Claim is taken at pick time — before CR-plan polling and before worktree creation
- [x] An indeterminate `gh` result (`unknown`) is treated as claimed and never reads as permission
- [x] A collaborator's claim is never released or overwritten, including under `--allow-claimed` (#732)
- [x] Rule corpus stays within its ratchet cap with no `--allow-raise`

### DIFF
--- a/.claude/reference/README.md
+++ b/.claude/reference/README.md
@@ -25,6 +25,7 @@ Files here are NOT auto-loaded by Claude Code. Agents read them on demand when w
 - `wrap-fixpr-delegation.md` — `/wrap` Step 2.1 → full `/fixpr` recovery handoff contract
 - `state-file-contracts.md` — expanded scoping, write-lock, migration, and field-type mechanics for `session-state.json` + handoffs (`handoff-files.md`; #625, #638, #639, #651, #655, #682, #687, #704)
 - `authorship-guard.md` — `pr-authorship.sh` exit-code semantics, the three shared-script fail-safes, and `--allow-nonauthor` override plumbing (`safety.md` §Authorship; #733)
+- `issue-claim.md` — `issue-claim.sh` mechanism: the three GitHub claim artifacts, why the holder token is finer-grained than the login, block-at-holder / release-at-account, staleness, the `--allow-claimed` override, and every call site (`issue-planning.md` step 0; #873)
 - `autofile-dedup.md` — duplicate-check thresholds for autonomous issue filing (`/wrap` Phase 3, `/harness-audit` Step 7); strong/weak/none classification, exact-artifact dedup, and the comment-vs-file rule
 - `harness-audit.md` — `/harness-audit` design record: two-pass split, the step-up chip that preserves the top-tier spawn invariant, the model-fleet resolver, dual watermarks, and the out-of-repo report default (#770)
 - `graphql-thread-resolution.md` — full GraphQL queries/mutations for resolving PR review threads
--- a/.claude/reference/chip-launching.md
+++ b/.claude/reference/chip-launching.md
@@ -117,6 +117,26 @@ A chip launches an independent thread that cannot see any sibling, so the launch
 
 **Click time, never offer time.** Offering a chip is not launching one (`/start-issue` "Execution boundary"), and a chip may sit unclicked for days — claiming at offer time would park the issue for a thread that never starts. This claims at exactly the moment work begins. It also means two chips for the same issue are harmless: whichever is clicked first holds the claim, and the second stops.
 
+### Claim line (shared contract — reproduce verbatim)
+
+Emitters spell out their own payload templates and copy only the **named** sections of this file, so a rule that lives only in the prose above never reaches a generated chip. The claim therefore ships as a `### Constraints` bullet, exactly like the merge-authority line. **Every issue-bearing block, every emitter, both delivery modes.**
+
+Two forms — pick by whether the emitter already holds the claim:
+
+**Form A — emitter holds no claim** (`/pm`, `/prompt`, `/wave`, `/issue-maker`): the launched thread takes it.
+
+```text
+- Claim the issue before anything else: run `.claude/scripts/issue-claim.sh <N> --check`, and if it clears, `.claude/scripts/issue-claim.sh <N> --claim`. Do this after the model-guard check and before any repo read, edit, or planning. Exit 1 (`claimed`) or 4 (`unknown`) → stop and report the claim rather than proceeding; `stale` → say so and continue. If `--claim` itself fails, stop — a passing check is not a held claim.
+```
+
+**Form B — emitter already holds the claim** (`/start-issue`, which claims at its Step 2b): the launched thread **inherits** that holder instead of competing with it.
+
+```text
+- This issue is already claimed for you (holder `{CLAIM_HOLDER}`). Re-affirm it before anything else — `.claude/scripts/issue-claim.sh <N> --claim --holder "{CLAIM_HOLDER}"` — after the model-guard check and before any repo read, edit, or planning. It is a no-op that confirms the claim is still yours; a non-zero exit means you do NOT hold it, so stop and report rather than proceeding.
+```
+
+**Why Form B exists.** A `/start-issue` run and the thread it hands off to are **one pickup of the issue, handed over** — not two independent threads racing. Without the inherited holder the launched thread would read its own predecessor's claim as a foreign one, exit 1, and refuse to start the work the chip exists to do. Passing the holder makes the re-claim a no-op (`mine`) while a genuinely different thread still sees `claimed`. `{CLAIM_HOLDER}` is the value the emitter passed to its own `--claim`.
+
 This does not touch the placement rule above — the claim instruction is *content following* the preamble, so the `**Model:**`-first / `**Effort:**`-next / no-blank-line ordering that `chip-model-guard-lint.sh` enforces is unchanged. Contract: `.claude/reference/issue-claim.md` (issue #873).
 
 ### Literal vs resolved model names (emitter classes)
--- a/.claude/reference/chip-launching.md
+++ b/.claude/reference/chip-launching.md
@@ -102,6 +102,23 @@ active model, so the check relies on the model naming itself accurately.
 
 **Placement rule:** for every emitter, the `**Model:**` line is the first line of the `prompt` payload, the `**Effort:**` line is next, and this preamble is the content that immediately follows them — see each skill's Step for how its own template maps onto this shape.
 
+## Claim the issue on click (not on offer)
+
+A chip launches an independent thread that cannot see any sibling, so the launched thread claims the issue itself. Every issue-bearing `prompt` payload instructs the thread to run, as its **first actions after the MODEL GUARD preamble and before any repo read, file edit, or planning**:
+
+```bash
+.claude/scripts/issue-claim.sh <N> --check      # then, if it clears:
+.claude/scripts/issue-claim.sh <N> --claim
+```
+
+- `claimed` (exit 1) or `unknown` (exit 4) → the thread **stops** and reports the claim instead of proceeding. `unknown` never reads as permission.
+- `stale` (exit 0) → surface the warning and proceed; `--claim` takes the claim over.
+- `mine` (exit 0) → no-op; a resumed or post-compaction thread re-runs this safely.
+
+**Click time, never offer time.** Offering a chip is not launching one (`/start-issue` "Execution boundary"), and a chip may sit unclicked for days — claiming at offer time would park the issue for a thread that never starts. This claims at exactly the moment work begins. It also means two chips for the same issue are harmless: whichever is clicked first holds the claim, and the second stops.
+
+This does not touch the placement rule above — the claim instruction is *content following* the preamble, so the `**Model:**`-first / `**Effort:**`-next / no-blank-line ordering that `chip-model-guard-lint.sh` enforces is unchanged. Contract: `.claude/reference/issue-claim.md` (issue #873).
+
 ### Literal vs resolved model names (emitter classes)
 
 Most emitters write the recommended model straight into the `**Model:**` line, because the recommendation is a judgment about *that issue* — a small mechanical fix and a subtle concurrency bug want different tiers, and neither answer comes from a lookup table.
@@ -197,6 +214,8 @@ Withdraw a tracked chip via `mcp__ccd_session__dismiss_task` (pass the recorded 
 2. **Superseded** — a later batch replaced the suggestion.
 3. **Re-planned** — the issue's plan or scope changed, so the chip's prompt is stale. Spawn the replacement chip *first*, then dismiss the old one.
 4. **Issue closed** — the underlying issue is closed (merged, resolved, declined, or duplicate). The chip's work no longer exists to launch; dismiss with no replacement. Distinct from trigger 1: a closed issue may never have had an *open* PR (e.g. "won't fix" or duplicate closures), so trigger 1 alone doesn't reliably catch it. Added by the [#838](https://github.com/auerbachb/claude-code-config/issues/838) sweep, where it was 28/28 of the stale chips found.
+
+   **Also release the claim here** — `.claude/scripts/issue-claim.sh <N> --release` alongside the `dismiss_task`. A close *without* a merged PR is precisely the terminal state `/wrap` never sees, so nothing else would drop the claim and it would sit until it aged out (issue #873). Harmless when no claim is held: `--release` is idempotent, and it never touches a collaborator's claim.
 
 **Fail-closed:** only clear tracked chip state once the dismiss outcome is known. Distinguish the two non-error outcomes from a real failure:
 
--- a/.claude/reference/issue-claim.md
+++ b/.claude/reference/issue-claim.md
@@ -1,0 +1,100 @@
+# Issue Claims — Mechanism and Rationale (issue #873)
+
+Detail for the claim rule in `.claude/rules/issue-planning.md`. Not auto-loaded; the rule file carries the binding behavior, this file carries the mechanism.
+
+## The problem
+
+Work gets picked up from several independent places — a `/pm` thread running issues inline, a click-to-launch chip, `/start-issue` in a fresh tab, or a plain "go work on #167". None of those threads can see each other; the only thing they share is GitHub.
+
+Every entry path *did* guard itself, but on the wrong question: **"does an open PR exist for this issue?"** A PR is the *last* artifact a thread produces. Between picking an issue and opening its PR there is a plan-and-code window that routinely runs half an hour or more, and for that entire window every other thread's check comes back clean. Two threads then work the same issue in good faith, and the collision only surfaces when the second PR appears. The cost is not a merge conflict — it is two full pipelines (planning, coding, CI, reviewer quota) buying one issue's worth of progress, plus a PR to throw away.
+
+`.claude/scripts/issue-claim.sh` stakes the claim at **pick** time instead.
+
+## Artifacts
+
+GitHub is the only store. There is no local claim state — a local file by definition cannot see a sibling thread, let alone another machine.
+
+| Artifact | Carries | Read by |
+|---|---|---|
+| `in-progress` label | the coarse "someone is working this" bit | any issue list, `/wave`'s batch pre-filter, a human skimming the backlog |
+| `@me` assignee | ownership at **account** level (#732) | `--release`'s ownership test |
+| claim comment `<!-- claude-claim: {"holder":…,"login":…} -->` | **holder** identity + the claimed-at timestamp (`created_at`) | `--check`'s `mine` vs `claimed` decision |
+
+The label is created idempotently on first use (`gh label list` existence check, then `gh label create … || true`) — the repo had no such label before this feature.
+
+## Why a holder token, not just the login
+
+This is the one place the implementation deliberately departs from the CodeRabbit plan, which defined `mine` as "label present **and** viewer in assignees".
+
+The failure this feature exists to prevent is **two Claude threads belonging to the same human**. Both authenticate as the same GitHub login, so an assignee-only test answers "yes, mine" in the second tab and blocks nothing — the feature would be inert in exactly its motivating case. The claim comment therefore records a holder token, and `mine` means *this thread*, not *this account*.
+
+`.claude/scripts/tests/issue-claim.test.sh` pins this: substituting the login-level comparison makes scenario (2) fail.
+
+**Holder resolution:** `--holder ID` → `$CLAUDE_CLAIM_HOLDER` → `$CLAUDE_SESSION_ID` → `<hostname>:<git toplevel>`.
+
+The default is honest about its granularity rather than hiding it: two threads sharing one checkout resolve to one holder. That is acceptable because `CLAUDE.md` mandates a worktree per thread, so the mandated workflow always yields distinct holders. A caller holding a real session id should still pass `--holder`.
+
+## Block at holder level, release at account level
+
+A deliberate asymmetry:
+
+- **`--check` / `--claim` compare holders**, so a sibling thread of yours is blocked. That is the whole point.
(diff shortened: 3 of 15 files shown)


auerbachb:
@graphite-app re-review

auerbachb:
Fixed in c94802d — confirmed, and broader than reported.

Root cause: `/pm` and `/prompt` spell out their own payload templates and copy only the *named* sections of `chip-launching.md` ("Model-guard preamble", "Merge-authority line"), so a rule living only in that file's prose never reaches their generated chips. Added a **Claim line** shared contract next to the merge-authority line, reproduced as a `### Constraints` bullet in `/pm`, `/prompt`, `/issue-maker` and `/start-issue`. `/wave` was already covered — its Step 7.1 follows `chip-launching.md` verbatim.

Your finding surfaced a second bug underneath it: `/start-issue` **already holds** the claim by the time it offers a chip (Step 2b), so a launched thread taking a fresh claim would read its own predecessor's claim as foreign, exit 1, and refuse to start the work the chip exists to do. A `/start-issue` run and the thread it hands off to are one pickup handed over, not two threads racing — so the contract has two forms, and `/start-issue`'s chip passes its holder for the launched thread to inherit.

One knock-on worth noting: the bullet originally read "after the MODEL GUARD", which made it satisfy `chip-model-guard-lint`'s grep for a model-guard *requirement* on its own. Reworded to "after the model-guard check" so the real requirement can't go missing undetected.

auerbachb:
Fixed in c94802d — valid, and the fix landed in the script rather than in `/wave`.

Rather than drop the batch pre-filter (which exists so a 30-issue backlog costs one API call instead of ~90), `issue-claim.sh` now maintains the label as a strict **superset** of the claim at every intermediate point: `--claim` writes the label before the comment and rolls the label back if the comment fails, and `--release` deletes comments *before* removing the label — and withholds the label removal entirely when a delete failed.

That last clause matters: ordering alone was necessary but not sufficient. My first attempt reordered the calls but still removed the label afterwards, reproducing the exact comment-without-label state one step later. The new regression test caught it.

So the state you describe is now unreachable, and `/wave` documents why the pre-filter is sound.

auerbachb:
Fixed in c94802d. Correct — a passing `--check` is not a held claim: another thread can win the race between the two calls, and the write itself can fail (exit 4). Proceeding on an unheld claim reopens exactly the duplicate-work window the step exists to close.

`/start-issue` Step 2b now gates on the `--claim` exit and stops on non-zero; `/subagent` Step 6.0 skips the issue and does not spawn Phase A.

auerbachb:
Fixed in c94802d. Confirmed — `mine` requires a parsed claim comment, so a run that wrote the label and assignee but failed to post the comment would read its *own* claim back as foreign and refuse to reclaim without `--allow-claimed`.

Chose rollback over widening `mine`: treating label+assignee as sufficient for `mine` would fall back to account-level identity, which is precisely the comparison this feature exists to avoid (two tabs of one human both answering "mine"). So `--claim` now rolls back exactly what that run added and reports the issue as unclaimed, leaving it retryable.

Two regression tests cover it: the rollback path (the same holder can then claim for real), and a stale-takeover case asserting rollback does **not** strip a pre-existing label the run did not add.

Issue #876, opened 2026-08-01
merge-gate.sh: CodeAnt in-place review edits false-positive the #836 force-push staleness guard


auerbachb:
## Problem

The `submitted_at`-predates-HEAD staleness guard added by #836 produces a **false positive** for CodeAnt specifically: CodeAnt appears to edit its existing PR review object **in place** on a re-review (same review `id`, `commit_id` correctly updated to the new HEAD) rather than posting a new review with a fresh `submitted_at`. The #836 guard only looks at `submitted_at` vs. the HEAD commit's committer date, so a genuine, freshly-completed re-review reads as stale and blocks the gate.

## Evidence (live run, `auerbachb/skingod` PR #2596)

1. Original CodeAnt review `id: 4833716091`, `state: APPROVED`, `submitted_at: 2026-08-01T04:57:00Z` on SHA `3e9dfd0e` (before a rebase).
2. Force-pushed to new SHA `f9b6041c` at `04:59:42Z`; posted `@codeant-ai review` at `05:02:01Z`.
3. CodeAnt genuinely re-ran — visible in its own PR comments:
   - `05:02:06Z` — "CodeAnt AI is running the review."
   - `05:02:34Z` — a sequence-diagram summary **accurately describing the actual diff** ("adds a SafeAreaProvider inside the native modal root...")
   - `05:03:35Z` — "CodeAnt AI finished running the review."
4. But `gh api .../pulls/2596/reviews` still shows the **same** review `id: 4833716091`, now `commit_id: f9b6041c...` (correctly updated) but `submitted_at` **unchanged** at `04:57:00Z` — i.e. before the SHA it now claims to cover even existed.
5. `merge-gate.sh 2596` reported: `"CodeAnt approval on HEAD f9b6041 predates the HEAD commit (force-push retargeting) — re-review required; trigger @codeant-ai review"` — even after the re-trigger and the confirmed-fresh review above. It stayed stuck on this for 15+ minutes of polling (8 ticks, 60s apart, zero change) despite CodeAnt having already delivered a real verdict on the current SHA.
6. Root cause confirmed: CodeAnt PATCHes the existing review resource rather than creating a new one, and GitHub's review API apparently treats `submitted_at` as immutable creation time (not last-modified time) — so `commit_id` is the trustworthy signal here, `submitted_at` is not, for this bot specifically.

[...]

Issue #875, opened 2026-08-01
merge-gate.sh counts hollow bot APPROVED reviews as review coverage


auerbachb:
## Problem

`merge-gate.sh` sets `primary_review_met: true` on any bot `APPROVED` review whose `commit_id` matches current HEAD. It already handles SHA freshness and same-SHA retraction — but it never asks whether the approval represents an actual review.

Across the 2026-07-30/31 sessions in `auerbachb/meeting_insights_and_actions`, bot approvals were routinely empty-bodied and posted seconds after a push, sometimes naming a different SHA in the bot's own status comment. The gate counted every one.

## Evidence — a pattern, not an anomaly

| PR | SHA | Observation |
|---|---|---|
| #171 | `3634336` | CodeAnt `APPROVED`, `bodylen=0`, **62s** post-push, **status comment still naming the previous SHA `98f0bd0`** |
| #172 | `d5976d8` | CodeAnt `APPROVED`, `bodylen=0`, **8s** post-push, status comment naming `396ced5` |
| #172 | `476798e` | CodeAnt `APPROVED` `bodylen=0` 8s post-push naming `cb1f770`; CodeRabbit `APPROVED` `bodylen=0` 25s post-push — **one minute after its own CLI reported an org-wide rate limit with a 37-minute wait** |
| #134 | final | **Merged on exactly this signal.** 539 lines of deploy tooling; nothing had genuinely reviewed the final SHA |

Every Phase B agent in those sessions had to be told, in its prompt, not to trust the gate's verdict and to check `bodylen` and timestamps instead. That instruction should not have to live in prompts.

## Critical refinement — body length alone is NOT the test

The obvious fix (reject `APPROVED` with an empty body) would produce false negatives. Counter-example from the same session:

- `meeting_insights_and_actions` PR #172 @ `396ced5`: CodeRabbit's `APPROVED` had **`bodylen=0`**, yet its accompanying comment named the exact `95febff…396ced5` range, listed all 3 changed files as processed, and carried an accurate walkthrough. That was a **genuine** review that happened to carry its substance in a sibling comment rather than the review body.

So the unit of evidence is **the reviewer's whole footprint on that SHA** — the review body *plus* its status/walkthrough comment — and critically **whether that comment names the SHA being approved**. The self-report mismatch (bot approves `X` while its own status comment says it reviewed `Y`) was the single most reliable discriminator observed: it fired on every hollow approval and on none of the genuine ones.

## Proposed direction

[...]

Issue #871, opened 2026-08-01
pmm-wake --auto-check can resume on a failed fleet scan


auerbachb:
Surfaced by CodeRabbit on PR #867 (issue #827), in code that PR did not modify — filing rather than widening that PR's scope.

`.claude/skills/pr-monitor-and-manage-wake/SKILL.md` Step 4a compares `CURRENT_FLEET` against `SAVED_FLEET` without checking that `gh pr list` and `jq` actually succeeded. On failure the empty/invalid result reads as a *fleet change*, so Step 4b cancels the re-scan, clears `.pmm.paused_at`, and starts a full monitor run with no confirmed change.

Fix: require both commands to succeed and the JSON to parse before comparing. On failure, keep the pause marker and the re-scan, and report a retry.

_Filed via /wrap._

## Implementation Plan

_Merged from Claude's analysis + CodeRabbit's `@coderabbitai plan` recommendations._

**File:** `.claude/skills/pr-monitor-and-manage-wake/SKILL.md`, Step 4a (the `--auto-check` bash fence).

### Principle

Fail closed. A scan that cannot **prove** the fleet changed must keep the pause. Resuming a full
monitor run the user never asked for is the expensive failure; one extra quiet re-scan tick is the
cheap one.

### Changes

1. **Capture the `gh pr list` exit status** instead of trusting it silently — `|| SCAN_RC=$?`,
   mirroring the `run_gh()` idiom in `.claude/scripts/pr-state.sh` (which uses `|| status=$?` so
   `set -e` does not swallow the code).
2. **Validate before comparing.** Introduce a `scan_failed()` helper that prints the failure, states
   the safe state preserved, and exits non-zero. Gate on, in order:
   - `gh pr list` exited non-zero,
   - its output is empty,
   - its output is not a JSON array (`jq -e 'type == "array"'` — note `[]` is a *valid* empty fleet
     and must pass; an all-PRs-merged fleet is a legitimate change, not a scan failure),
   - the `jq` normalisation of `CURRENT` into `CURRENT_FLEET` fails,
   - `.pmm.fleet_at_pause` is missing or not a JSON array. **This last one is the same defect class
     and CodeRabbit's plan missed it:** a null/garbage saved snapshot makes `SAVED_FLEET` empty, so
     a perfectly good `CURRENT_FLEET` compares as "changed" and resumes just as wrongly.
3. **Preserve safe state on every failure path** — do not run Step 3's re-scan cancel, do not clear
[...]

Issue #865, opened 2026-08-01
Sticky BugBot path ignores a fresh CodeAnt APPROVED on the same HEAD


auerbachb:
Observed during PR #858's merge watch. Nothing broke — the PR merged — but the gate sat blocked on a reviewer that had already been satisfied by a different one, and the wait was avoidable.

Possibly duplicates #844 — that issue fixed the *strand* case (a BugBot silent pass with no review object now counts, via the success check-run). This is the adjacent case that #844's fix does not cover.

## What happened

1. CodeRabbit's App was rate-limited on the PR's SHA, so `escalate-review.sh` returned `STATUS=switch_bugbot` and BugBot became the sticky reviewer.
2. The branch was rebased; on the new HEAD, **CodeAnt posted a genuine `APPROVED` review** ~4 minutes later.
3. `merge-gate.sh` still reported `missing: ["no BugBot review on HEAD"]`, because the reviewer was sticky-`bugbot` and CodeAnt sits on the CR path.
4. The gate stayed closed for ~14 more minutes until BugBot's check-run finally landed (its auto-trigger had not fired on the force-push; a manual `@cursor review` was needed).

## The question

The sticky rule exists to stop a PR from wandering back up the escalation chain hoping for a friendlier verdict. That rationale does not obviously cover this shape: the CR-path reviewer did not need to be *hoped for* — it independently produced a fresh `APPROVED` on the exact HEAD SHA while the PR waited.

Options worth weighing:

- **Leave as is.** Stickiness is simple and predictable; the cost is a bounded wait.
- **Let a fresh CR-path `APPROVED` on current HEAD satisfy the gate even under sticky BugBot.** Removes the wait; slightly weakens "sticky means sticky".
- **Un-stick on a fresh CR-path approval specifically** (reviewer returns to `cr` when CodeRabbit or CodeAnt approves current HEAD), keeping stickiness for the silence case that motivated it.

Whichever way it goes, the reasoning belongs in `.claude/reference/merge-gate-reviewer-paths.md` so the next reader does not have to re-derive it.

## Acceptance criteria

- [ ] A decision is recorded on whether a fresh CR-path `APPROVED` on current HEAD can satisfy the gate while the PR is sticky-assigned to BugBot
- [ ] If the answer is yes, `merge-gate.sh` implements it and a test covers the sticky-bugbot + CodeAnt-approved-on-HEAD case
- [ ] If the answer is no, the rationale is written down where the sticky rule is defined

[...]

PR #891, opened 2026-08-01
fix(#885): make gate/escalation timestamp drift fail in CI, not in review

### LINKED ISSUE #885: Timestamp normalisation is duplicated across merge-gate.sh and escalate-review.sh
Follow-up from PR #883.

Three separate ISO-8601 timestamp normalisers now implement overlapping rules across the merge-gate call graph:

- `norm_ts` in `.claude/scripts/merge-gate.sh` — strips the zone suffix (`Z`, `+00:00`, `+0000`), **keeps** fractional seconds.
- `canon_ts` in `.claude/scripts/escalate-review.sh` (approval-freshness filter, #836) — must match `norm_ts` byte for byte, because escalation is a caller of the gate and must never be the more permissive of the two.
- `canon_ts` in `.claude/scripts/review-substance.sh` — strips fractional seconds **and** folds the UTC spellings onto `Z`. Deliberately different: every timestamp it compares passes through that one function, so it is internally consistent and is never compared against a merge-gate value.

**Why this is worth a ticket:** two of the three review rounds on PR #883 were findings caused purely by these drifting apart, both real, both caught only by review rather than by any test or lint:

1. `canon_ts` in `escalate-review.sh` dropped fractional seconds while `norm_ts` kept them, so the two ordered the same pair differently and escalation could report `gate_met` on an approval the gate rules stale.
2. The repair then normalised `+0000` while `norm_ts` did not — the same divergence in a new spelling, introduced by the fix for the first one.

The invariant "escalation must order timestamps exactly as the gate does" is currently maintained by hand across two files, with a comment in each asking the reader to keep them in step. Nothing fails if they drift; the symptom is a wrong merge-readiness verdict.

**Possible shapes** (not a decision — this is an observation):

- Extract one shared normaliser into `.claude/scripts/lib/` and have both the gate and escalation source it, leaving `review-substance.sh` explicitly on its own documented variant.
- Or keep the duplication and add a test asserting the two functions produce identical output across the spelling matrix (`Z`, `+00:00`, `+0000`, fractional, mixed) — cheaper, and it makes drift fail loudly.

The spelling matrix is already written out in PR #883's verification and in scenarios (h6b), (h6c), (h6d), (h6e) of `.claude/scripts/tests/escalate-review.test.sh`.

Linked source: #875

_Filed via /wrap._

## Implementation Plan

Merged from Claude's analysis + CodeRabbit's coding plan (Design Choice 1, option 3 — hybrid).

### Decision: hybrid, not full unification

Three normalisers, two languages:

| Site | Language | Rule |
|---|---|---|
| `merge-gate.sh:norm_ts` | bash | strip one trailing UTC suffix (`Z`, `+00:00`, `+0000`); **keep** fractional seconds |
| `escalate-review.sh:canon_ts` | jq (inside a single-quoted jq program) | must be **byte-identical** to the above |
| `review-substance.sh:canon_ts` | jq | strip fractional seconds **and** fold UTC spellings onto `Z` — deliberately different |

A single sourced bash helper cannot dedup the jq side: jq cannot source a bash function, and the
repo has no jq-module (`jq -L` / `import`) convention. Generating the jq `def` from a bash variable
and splicing it into `escalate-review.sh`'s single-quoted program would be **new infrastructure
invented as a side effect** — exactly what the ticket warns against — and would touch quoting that
the script's own comment flags as fragile.

So:

1. Extract the bash `norm_ts` into `.claude/scripts/lib/ts-normalizer.sh`, mirroring the existing
   `lib/repo-normalizer.sh` precedent (issue #704). This is the convention the repo already has.
2. Leave `escalate-review.sh`'s jq `canon_ts` inline.
3. **Enforce the cross-language invariant with a test that extracts both real implementations from
   the shipped files at runtime** and runs them over the spelling matrix. This is the deliverable's
   core — it is what would have caught both PR #883 findings.

### Phase 1 — extract the canonical bash normaliser

- New `.claude/scripts/lib/ts-normalizer.sh`, modelled on `lib/repo-normalizer.sh`: shebang,
  "source this, do NOT execute", `PROBLEM SOLVED` (#836 / #883 / #885), `CONTRACT`, `USAGE`.
  No `set -euo pipefail`, no double-source guard — matching the existing lib.
- `norm_ts()` strips **exactly one** trailing UTC suffix, keeps fractional seconds. The current
  merge-gate implementation chains three `${t%…}` expansions, so a hypothetical `…+00:00Z` would
  lose both suffixes while the jq alternation strips only one. Tightening bash to strip one makes
  the two provably identical; no real GitHub timestamp carries two suffixes, so no call site moves.
- Header states explicitly that `review-substance.sh:canon_ts` is a deliberately different variant
  and is out of scope for this library.
- `merge-gate.sh` resolves `SCRIPT_DIR` and sources the lib with the hard-fail pattern from
  `session-state.sh` (missing/unreadable → stderr + non-zero exit, `# shellcheck source=` directive).
  Inline `norm_ts` is deleted; every call site is unchanged. The long inline rationale shrinks to a
  pointer at the lib and the parity test.

### Phase 2 — the regression test (core deliverable)

New `.claude/scripts/tests/ts-normalizer-parity.test.sh` (auto-discovered by
`.github/scripts/run-hook-tests.sh`; no workflow edit needed). Conventions follow
`escalate-review.test.sh` / `merge-gate-stale-approval.test.sh`.

- Sources the lib for the real `norm_ts`.
- **Extracts the jq `def canon_ts: … end;` block from `escalate-review.sh` and from
  `review-substance.sh` at runtime** (from the `def canon_ts:` line through the first following
  line ending in `end;`) and runs each through `jq`. The logic is never retyped in the test, so an
  edit to production code is what the test observes. Extraction failure is a loud FAIL, never a
  silent pass; the test also asserts exactly one `def canon_ts:` per file.
- **Parity matrix** (from the ticket and scenarios h6b–h6e): bare, `Z`, `+00:00`, `+0000`,
  `.900Z`, `.900000Z`, fractional-plus-offset, empty string. For every input,
  `norm_ts` output must equal `escalate-review.sh`'s `canon_ts` output byte for byte.
- **Ordering assertions** in the (h6c)/(h6d) shape: for the pairs that PR #883 got wrong, both
  implementations must produce the *same* `<` / `>` verdict — parity at the comparison level, not
  just the string level.
- **Divergence pins** for `review-substance.sh`: assert its `canon_ts` genuinely differs from the
  gate rule (drops the fraction, folds `+00:00`/`+0000` to `Z`) on at least one fractional and one
  offset input, with a comment recording that the difference is intended and why. A future
  "unify all three" refactor then fails loudly instead of silently changing that script's ordering.
- Direct unit coverage of `norm_ts` per spelling, including empty input and fraction retention.
- `escalate-review.sh`'s parity comment is repointed at the test (behaviour unchanged; the
  no-apostrophes-inside-the-jq-program warning is preserved).

### Non-goals / interaction with adjacent open work

- No rule-corpus edit. Corpus is at 11,730 / 11,749 — this is scripts + tests only, and any
  narrative goes in the script headers (not auto-loaded).
- **Issue #865** (sticky BugBot vs. a fresh CodeAnt APPROVED) changes *which reviewer path* the
  gate consults. It does not touch timestamp normalisation; `norm_ts` keeps the same name and
  semantics, so that work is unaffected.
- **Issue #876** (CodeAnt in-place review edits false-positive the #836 staleness guard) changes
  *what signal* the freshness check trusts (`commit_id` / check-run corroboration instead of
  `submitted_at` alone). Whatever timestamps it ends up comparing still go through `norm_ts`, and
  the parity test constrains only normalisation, not which fields are compared — so it does not
  foreclose either option in that ticket's fix sketch.

### Acceptance criteria

- [ ] `.claude/scripts/lib/ts-normalizer.sh` exists and is the only bash definition of `norm_ts`
- [ ] `merge-gate.sh` sources it, hard-fails if missing, and behaves identically at every call site
- [ ] `escalate-review.sh`'s jq `canon_ts` is unchanged in behaviour
- [ ] `review-substance.sh`'s deliberately different `canon_ts` is unchanged
- [ ] A test extracts both jq defs from the shipped files and fails if gate/escalation normalisation
      diverges on any of `Z`, `+00:00`, `+0000`, fractional, mixed, empty
- [ ] The same test pins `review-substance.sh`'s intended divergence
- [ ] The test fails loudly if extraction breaks (no silent pass)
- [ ] `shellcheck` clean on every touched script; `run-hook-tests.sh` green
- [ ] Rule corpus word count unchanged (no `.claude/rules/` or `CLAUDE.md` edit)

auerbachb:
Closes #885

## What this changes

Three ISO-8601 normalisers implement overlapping rules across the merge-gate call graph, and two of them **must agree byte for byte** — `norm_ts` in `merge-gate.sh` and the jq `canon_ts` in `escalate-review.sh`. Escalation is a *caller* of the gate, so if it is ever the more permissive of the two it reports `gate_met` on an approval the gate rules stale.

That invariant was maintained by hand, with a comment in each file asking the reader to keep them in step. It broke twice on PR #883, and both times only human review caught it:

1. `canon_ts` dropped fractional seconds while `norm_ts` kept them, so `…49.900Z` and `…49Z` compared **equal** on one side and **ordered** on the other.
2. The repair then normalised `+0000` on one side only — the same divergence in a new spelling, introduced by the fix for the first. When two functions must agree, a superset on one side is divergence, not slack.

This PR makes that drift fail in CI instead of in review.

## Approach — hybrid, not full unification

`norm_ts` is bash; both `canon_ts` implementations are jq embedded in single-quoted jq programs. jq cannot source a bash function, and this repo has **no** jq-module (`jq -L` / `import`) convention. Generating the jq `def` from a bash variable and splicing it into `escalate-review.sh` would be new infrastructure invented as a side effect, touching quoting that the script's own comment flags as fragile. So:

- **Extract the bash side into `.claude/scripts/lib/ts-normalizer.sh`**, mirroring the existing `lib/repo-normalizer.sh` precedent (#704). `merge-gate.sh` sources it with the hard-fail pattern from `session-state.sh`. One bash definition, and one clearly-named thing for the jq mirror to agree with.
- **Leave the jq mirror inline** in `escalate-review.sh` — unchanged in behaviour.
- **Enforce the cross-language invariant with a test that extracts both real implementations from the shipped files at runtime.** No logic is retyped in the test, so an edit to production code is what the test observes.

`review-substance.sh`'s `canon_ts` stays deliberately different (strips the fraction, folds UTC spellings onto `Z`). That is load-bearing: every timestamp it compares passes through that one function, so it is internally consistent and is never compared against a merge-gate value. The difference is now **asserted**, not just commented — a future "unify all three" refactor fails loudly instead of silently reordering that script's markers.

## Behaviour changes

One, deliberate and inert on real data: `norm_ts` now strips **exactly one** trailing UTC suffix. The previous form chained three `${t%…}` expansions, so a hypothetical `…+00:00Z` would lose both suffixes while the jq alternation strips only one — a latent parity gap. No GitHub timestamp carries two suffixes; every live comparison is unchanged. Nothing else about the rule moves.

`merge-gate.sh` exits `4` (its documented local/tooling-error code, same as `die_local` for a missing sibling script) if the library is absent. The exit is before `die_local` and `HEAD_SHA` exist, so it prints to stderr rather than emitting the JSON envelope — unavoidable at source time, and it is a broken-install path, not a gate verdict.

## Verification — the guard was proven to fail on the real regressions

Both PR #883 findings were re-injected into `escalate-review.sh` and the suite was run against each:

| Injected drift | Result |
|---|---|
| Finding 1 — `canon_ts` drops the fraction and re-appends `Z` | **12 failures**, including `(h6c)`/`(h6d)` at the comparison level: `escalation agrees (expected 'lt', got 'eq')` |
| Finding 2 — `canon_ts` handles `Z`/`+00:00` but not `+0000` | **3 failures**, including `equal instant, +00:00 vs +0000 — escalation agrees (expected 'eq', got 'lt')` |
| Restored (both reverted) | **53 passed, 0 failed** |

A guard that cannot be shown to fail is not a guard.

The extraction machinery was proven fatal too: renaming `def canon_ts:` in `escalate-review.sh` aborts the suite with `expected exactly 1 'def canon_ts:' … found 0`, exit 1 — it does not quietly stop comparing.

The structural guards read **executable lines only** (comment-only lines are stripped before matching). That matters here: `merge-gate.sh` mentions `ts-normalizer.sh` on 4 lines, only 1 of which is the actual `source` — a naive `grep -q` would stay green after someone deleted the wiring and left the prose.

Full suite: `run-hook-tests.sh` — **57 suites discovered, 0 failed**. `rule-lint.sh` — OK. `shellcheck -x` on all five touched files emits only `SC1091 (info): Not following` for the sourced library, identical in kind to the existing `session-state.sh` → `lib/repo-normalizer.sh` baseline; no warnings, no errors.

**Rule corpus unchanged: 11,730 words against the 11,749 cap.** No `CLAUDE.md` or `.claude/rules/` edit — this is scripts + tests, and all narrative lives in script headers (not auto-loaded, therefore free). `.budget-soft-cap` untouched.

## Interaction with adjacent open work

Neither is being worked; read both so this change does not foreclose either.

- **Issue #865** (sticky BugBot vs. a fresh CodeAnt `APPROVED`) changes *which reviewer path* the gate consults. It does not touch normalisation. `norm_ts` keeps its name, signature, and semantics at every call site, so that work is unaffected.
- **Issue #876** (CodeAnt in-place review edits false-positive the #836 staleness guard) changes *what signal* the freshness check trusts — `commit_id` or check-run corroboration instead of `submitted_at` alone. Whatever timestamps it ends up comparing still flow through `norm_ts`, and the parity test constrains only *normalisation*, never *which fields are compared*. Both options in that ticket's fix sketch remain open.

**Contention:** verified before pushing. The sibling pipeline on issues #871/#872 touches `.claude/skills/pr-monitor-and-manage*/SKILL.md` — no overlap with this file set, and `origin/main` had not moved off `023eb4c` at push time.

## Test plan

- [x] `.claude/scripts/lib/ts-normalizer.sh` exists and is the only bash definition of `norm_ts`
- [x] `merge-gate.sh` sources it, hard-fails with a clear stderr message if it is missing, and every `norm_ts` call site is unchanged
- [x] `escalate-review.sh`'s jq `canon_ts` is unchanged in behaviour (comment repointed at the enforcing test only)
- [x] `review-substance.sh`'s deliberately different `canon_ts` is unchanged in behaviour
- [x] The parity test extracts both jq defs from the **shipped files** at runtime and never retypes the logic
- [x] Parity holds across the full spelling matrix: bare, `Z`, `+00:00`, `+0000`, `.900Z`, `.900000Z`, fractional+offset, non-UTC offset, empty
- [x] Parity is asserted at the **comparison** level for scenarios (h6b)–(h6e), not just on normalised strings
- [x] `review-substance.sh`'s intended divergence is pinned (fraction dropped, UTC folded to `Z`) so unification fails loudly
- [x] The divergence direction is pinned as one-directional (substance is the more permissive rule) and `merge-gate.sh`'s outer freshness guard around the substance check is asserted, so substance can never restore a `norm_ts`-stale approval
- [x] Structural guards read executable lines only, and fail if `norm_ts` is re-inlined, the `source` is deleted, or `canon_ts` stops being called
- [x] Extraction failure (missing def, wrong count, truncated block, invalid jq) is a loud FAIL, never a silent pass — including that `die` inside a command substitution cannot swallow the abort
- [x] Verified by injection: both PR #883 regression shapes make the suite red; restoring makes it green
- [x] `run-hook-tests.sh` green (57 suites, 0 failed)
- [x] `shellcheck -x` clean on all touched scripts against the repo's existing baseline
- [x] Rule corpus word count unchanged (11,730 / 11,749); `.budget-soft-cap` not raised

**Local review coverage:** cr-only

CodeRabbit CLI raised three findings on the first pass and all were addressed before the second:

- **major** — structural guards matched any text occurrence rather than executable usage. Fixed: guards now strip comment-only lines and separately assert an executable `source` **and** real call sites in `merge-gate.sh`, `escalate-review.sh`, and `review-substance.sh`.
- **minor** — `REPO_ROOT` came from a bare `git rev-parse`, so the header's "run from anywhere" claim was false. Fixed: anchored on `${BASH_SOURCE[0]}` via `git -C`.
- **trivial** — process reminder to complete the GitHub review loop before declaring done. Not a code finding; skipped, and the loop is being followed.

A latent bug of my own was also caught while addressing these: an intermediate refactor put the fatal `die` inside a command substitution, where `exit 1` would kill only the subshell and let the suite continue with an empty jq program — precisely the silent false-clean this file exists to prevent. The builder now writes to a global instead, with a comment explaining why.

The second pass raised three more, all fixed:

- **minor** — the lexicographic example in the new library header was simply **wrong**: `…10:00:16Z` vs `…10:00:22+00:00` differ at the seconds digit, so the suffix never decides. The trap only bites when the whole-second prefix matches. Rewritten around a correct pair (`…16+00:00` < `…16Z`, same instant). **The same error existed in `escalate-review.sh`, inherited from PR #883** — corrected there too, since that block is already in this diff.
- **minor** — no readability guard on `merge-gate.sh` before the structural checks. That mattered: an unreadable file makes `code_lines` emit nothing, every count returns 0, and the one check whose *passing* value is 0 ("defines no local `norm_ts`") would go green on an empty read. Now a fatal `die`.
- **trivial** — the local-def check only matched `norm_ts() {`. Broadened to `function norm_ts {` and `function norm_ts() {`.

The third pass took ~70 minutes but did return, with two more findings — the second of which was the most useful of the run:

- **minor** — the library-guard error message said "missing" while the condition also tests unreadable. Reworded.
- **trivial (but substantive)** — my `review-substance.sh` comment claimed its `canon_ts` is "never compared against a merge-gate value". **That was too strong and I verified it against the code: `merge-gate.sh` does hand it the same raw HEAD committer date it uses as `LAST_COMMIT_TS`, arriving as `push_ts`.** What never happens is a comparison *across* the two rules — both sides of every compare in that script are normalised by its own function. Reworded to say that precisely.

  Following that thread surfaced a real, untested, load-bearing property. Dropping the fraction makes `canon_ts` **strictly more permissive** than `norm_ts`, so it can call equal what the gate orders. That is safe *only* because the two verdicts are ANDed with freshness on the **outside**: `merge-gate.sh` computes `CR_APPROVAL_STALE` with `norm_ts` and tests `counts_as_coverage` **inside** that guard, so substance can only ever subtract coverage — never restore an approval `norm_ts` has ruled stale. Invert that nesting and a substance-fresh / gate-stale approval slips through. The suite now pins both the divergence direction and the guard structure (4 new assertions, 53 total).

Local suite re-verified after each round: 53/53 parity, 57/57 suites, zero shellcheck warnings. One full-suite run showed `handoff-state.test.sh` red; it passes in isolation and on re-run — it had raced this session's own handoff-file write, not a code regression.

CodeAnt CLI was dropped for this session after two consecutive `403 Access denied` responses on `codeant review --all --headless` (the known undocumented ~10/day agent-review cap — not an auth fault; `codeant logout`/`login` deliberately not run). Both runs exited `0` with `"issues": []` and reported `0 files have suggestions`, which is a **failed run**, not a clean pass: the stderr stream carried `API Error` / `[error] Failed to review` for every file. The CodeAnt GitHub App is unaffected.

## Phase B — review round 1 (commit `1d2ab2c`)

Phase B audited the outer-guard assertions added in `4c4f86f` rather than
inheriting them, and found the pin did not constrain what its own comment said
it constrained.

**The gap.** The guard check was two `grep -q`s over a 12-line window: does
`<PREFIX>_APPROVAL_STALE" == false` appear, and does `<PREFIX>_SUBSTANTIVE" == true`
appear. Proven by injecting each shape into the real `merge-gate.sh` and
re-running the suite:

| Injected refactor | Old assertions | New assertions |
|---|---|---|
| Staleness check **deleted** from the outer conditional | 1 failure (caught) | caught |
| Substance test turned into a **disjunction** (`\|\|` with the staleness flag) | **silently passed** | 1 failure |
| Staleness moved out of the outer guard into an inner disjunction | **silently passed** | 3 failures |

The two silent passes are precisely the "substance restores a `norm_ts`-stale
approval" break the pin exists to prevent — both strings are still present, so
presence-greps stay green while the invariant is gone.

**The fix.** Pin the structure rather than the tokens: the staleness test must
sit on a strictly earlier (outer) line than the substance test — which also
proves the two are not folded onto one condition line — and neither line may
contain a disjunction operator. The block is now delimited by its closing `fi`
instead of a fixed line count, so adding a comment inside it cannot push the
substance test out of view and cause a spurious failure.

**Coverage extended to CodeAnt.** The `CA_APPROVAL_VALID` block had *no*
assertion at all. It is structurally identical to the CR block and carries the
identical hazard, and hollow CodeAnt approvals are what motivated #875 in the
first place. Both paths are now checked by one shared helper.

`merge-gate.sh` is unchanged by this commit — every injection above was reverted
and the file verified byte-identical (`git diff` empty).

**Superseding the counts above:** parity suite is now **61 passed, 0 failed**
(was 53). `run-hook-tests.sh` 57 suites, 0 failed. `shellcheck` on the test file
emits only the pre-existing `SC1091` / `SC2016` info lines. Rule corpus still
untouched at **11,730 / 11,749** — this PR's `CLAUDE.md` + `.claude/rules/`
delta is zero.

**Adjacent open work re-confirmed.** Issues #865 and #876 both remain fully
open. Neither touches normalisation, and the parity suite still constrains only
*how* timestamps normalise and *how the two verdicts nest* — never *which*
fields are compared, which is what #876 would change. If either lands a change
to the freshness signal, these assertions fail loudly with a message naming the
invariant, which is the intended behaviour of a pin, not a blocker.

## Phase C — independent AC verification (commit `1d2ab2c`)

Every Test Plan box was verified against the committed source, not against this
body or prior summaries. The rewritten guard pin was **re-proven by injection**
in an isolated scratch copy of `.claude/scripts` — the real files were never
modified — because an assertion that passes while the invariant is broken is
worse than no assertion at all.

| Injected break | Suite result |
|---|---|
| CR substance test folded into a disjunction | **1 failure** — named the disjunction |
| CR staleness moved out of the outer guard into an inner disjunction | **2 failures** — ordering + disjunction |
| CA staleness conjunct **deleted** (the path Phase B newly covered) | **1 failure**, then early-return |
| Executable `source` of the library removed, prose left in place | **1 failure** |
| Local `norm_ts` re-inlined into `merge-gate.sh` | **1 failure** |
| `def canon_ts:` renamed in `escalate-review.sh` (0 defs) | **aborted**, exit 1 |
| Duplicate `def canon_ts:` added (2 defs) | **aborted**, exit 1 |
| #883 shape 1 — jq `canon_ts` drops the fraction | **7 failures**, incl. comparison level |
| #883 shape 2 — `+0000` dropped from the jq alternation | **3 failures**, incl. `equal instant, +00:00 vs +0000 — escalation agrees (expected 'eq', got 'lt')` |
| All reverted | **61 passed, 0 failed** |

Independently confirmed alongside the above:

- Exactly **one** bash `norm_ts` definition exists repo-wide, in `lib/ts-normalizer.sh`.
- All **12** `norm_ts` call sites in `merge-gate.sh` are byte-identical to `origin/main`.
- `escalate-review.sh` and `review-substance.sh` have a **comment-only** diff — both
  `canon_ts` bodies are unchanged, confirmed by extracting each def and diffing
  non-comment lines against `origin/main`.
- Missing library ⇒ exit `4`, stderr `sibling library missing or unreadable: …`,
  **no** JSON on stdout.
- `run-hook-tests.sh`: **57 suites discovered, 0 failed**. `rule-lint.sh`: OK.
- `shellcheck -x` on the five touched files: only `SC1091`/`SC2016` **info** — strictly
  cleaner than the untouched `session-state.sh` baseline, which additionally carries
  `SC2295`, `SC2015` and an `SC2064` **warning**.
- Rule corpus **11,730 / 11,749**; `git diff origin/main...HEAD -- CLAUDE.md .claude/rules/`
  is empty; `.budget-soft-cap` unchanged.

**Review state at merge:** BugBot clean on `1d2ab2c` (`Cursor Bugbot` check-run
`completed`/`success`, body names the full HEAD SHA, no failure phrase); CI 5/5;
0 unresolved threads. CodeRabbit did not return — it is under an org-wide Fair
Usage limit and posted no review. CodeAnt's three `body=0` approvals are hollow
and are correctly discounted by the #875 detector this PR modifies.

### DIFF
--- a/.claude/scripts/escalate-review.sh
+++ b/.claude/scripts/escalate-review.sh
@@ -178,27 +178,31 @@ if [[ "$PRIMARY_REVIEW_MET" == "true" ]]; then
     FRESH_APPROVERS="$(jq -rn --arg sha "$HEAD_SHA" --arg push "$HEAD_COMMIT_TS" \
       --arg logins "$VALID_APPROVERS" --slurpfile doc "$STATE_PATH" '
       # Canonicalise before comparing (BugBot review, PR #883): this is a
-      # lexicographic string compare, and "…T10:00:22+00:00" sorts BEFORE
-      # "…T10:00:16Z", so a mixed spelling would mark a fresh approval stale and
-      # withhold gate_met on a PR merge-gate.sh considers fine.
+      # lexicographic string compare, so when two timestamps share a
+      # whole-second prefix the zone suffix is the first differing character —
+      # and "+" (0x2B) is below "Z" (0x5A), so "…T10:00:16+00:00" sorts BELOW
+      # "…T10:00:16Z" despite being the SAME instant. A push stamped "…16Z"
+      # against an approval stamped "…16+00:00" therefore reads as
+      # approval-before-push: a fresh approval is marked stale and gate_met is
+      # withheld on a PR merge-gate.sh considers fine.
       #
       # STRIP the zone suffix, do NOT rewrite it to "Z", and KEEP fractional
-      # seconds — this must reproduce exactly the ordering that norm_ts in
-      # merge-gate.sh produces, because the two scripts implement one rule
+      # seconds — this is the jq mirror of norm_ts in lib/ts-normalizer.sh and
+      # must reproduce its ordering exactly, because the two implement one rule
       # (#836) and a caller more permissive than the gate is a bug (BugBot
-      # review on 7de2a4c, PR #883). The earlier form dropped fractions and
-      # appended "Z", diverging in BOTH directions:
-      #   * dropping ".900" made "…49.900Z" and "…49Z" compare EQUAL, so an
-      #     approval the gate rules stale read as fresh here and escalation
-      #     could report gate_met on a PR merge-gate.sh blocks;
-      #   * a fraction retained under a trailing "Z" sorts the WRONG WAY —
-      #     "." (0x2E) is below "Z" (0x5A), so the LATER instant "…49.900Z"
-      #     compares below "…49Z".
-      # With the suffix gone, plain lexicographic order is correct for whole
-      # and fractional seconds alike. canon_ts in review-substance.sh keeps its
-      # own fraction-stripping form: every timestamp it compares passes through
-      # that one function, so it stays internally consistent and is never
-      # compared against a merge-gate value.
+      # review on 7de2a4c, PR #883). jq cannot source a bash function and this
+      # repo has no jq-module convention, so the mirror stays inline by design.
+      #
+      # It is NOT kept in step by hand (issue #885):
+      # tests/ts-normalizer-parity.test.sh extracts THIS def block from THIS
+      # file and asserts it agrees with norm_ts byte for byte across the whole
+      # spelling matrix (Z, +00:00, +0000, fractional, mixed, empty). Any edit
+      # that diverges — including a well-meant superset — fails that test.
+      #
+      # canon_ts in review-substance.sh keeps its own fraction-stripping form:
+      # every timestamp it compares passes through that one function, so it
+      # stays internally consistent and is never compared against a merge-gate
+      # value. The parity test pins that divergence too.
       #
       # NOTE: this comment sits inside a single-quoted jq program — no
       # apostrophes, or the quoting ends here and bash mis-parses the block.
--- a/.claude/scripts/lib/ts-normalizer.sh
+++ b/.claude/scripts/lib/ts-normalizer.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+# lib/ts-normalizer.sh — Shared ISO-8601 normaliser for merge-gate timestamp
+# ordering (issue #885).
+#
+# Source this file (do NOT execute it directly) from merge-gate.sh — and from
+# any future bash caller that orders GitHub review/commit timestamps against the
+# gate's rule — so there is exactly one bash definition of that rule.
+#
+# PROBLEM SOLVED
+#   GitHub returns mixed ISO-8601 UTC spellings for the same instant: "…Z",
+#   "…+00:00", and the compact "…+0000". Bash [[ < ]] and jq's string compare are
+#   both purely lexicographic, so the FIRST differing character decides — and
+#   when two timestamps share a whole-second prefix, that character is the zone
+#   suffix. Z (0x5A) > + (0x2B), so:
+#
+#       "2026-08-01T10:00:16+00:00"  <  "2026-08-01T10:00:16Z"
+#
+#   These are the SAME instant, but they do not compare equal. The #836 rule
+#   asks whether an approval is at or after the HEAD commit, so a push stamped
+#   "…16Z" against an approval stamped "…16+00:00" reads as approval-before-push
+#   and a perfectly fresh approval is ruled stale. Nothing errors; the gate just
+#   returns the wrong merge-readiness verdict.
+#
+#   Two of the three review rounds on PR #883 were this bug:
+#     1. escalate-review.sh dropped fractional seconds while merge-gate.sh kept
+#        them, so the two ordered the same pair differently and escalation could
+#        report gate_met on an approval the gate rules stale.
+#     2. The repair then normalised "+0000" on one side only — the same
+#        divergence in a new spelling, introduced by the fix for the first.
+#   When two implementations must agree, a superset on one side is divergence,
+#   not a harmless extra.
+#
+# CONTRACT — the gate rule
+#   Strip EXACTLY ONE trailing UTC zone suffix ("Z", "+00:00", or "+0000") and
+#   change nothing else. In particular:
+#     * KEEP fractional seconds. These are DIFFERENT instants, and dropping
+#       ".900" collapses "…49.900Z" and "…49Z" to the same COMPARISON VALUE —
+#       so a strictly-earlier approval compares equal to the push and an
+#       approval the gate rules stale reads as fresh.
+#     * Do NOT rewrite the suffix to "Z". A fraction retained under a trailing
+#       "Z" sorts the WRONG WAY — "." (0x2E) is below "Z" (0x5A), so the LATER
+#       instant "…49.900Z" compares below "…49Z". With the suffix gone, plain
+#       lexicographic order is correct for whole and fractional seconds alike.
+#     * A genuine non-UTC offset is left untouched rather than mangled into a
+#       wrong instant.
+#   Output is a bare "YYYY-MM-DDTHH:MM:SS[.fraction]" string. Empty input gives
+#   empty output.
+#
+# CROSS-LANGUAGE MIRROR — escalate-review.sh
+#   escalate-review.sh implements the same rule as a jq `def canon_ts:` inside a
+#   single-quoted jq program. jq cannot source a bash function and this repo has
+#   no jq-module convention, so that mirror stays inline by design. It is not
+#   maintained by hand: `.claude/scripts/tests/ts-normalizer-parity.test.sh`
+#   extracts the jq def from the shipped file and asserts it agrees with
+#   norm_ts() byte for byte across the whole spelling matrix. Any edit that
+#   diverges fails that test.
+#
+#   Escalation is a CALLER of the gate, so it must never be the more permissive
+#   of the two — that asymmetry is why parity is enforced rather than trusted.
+#
+# OUT OF SCOPE — review-substance.sh
+#   review-substance.sh has its own `canon_ts` that strips fractional seconds
+#   AND folds the UTC spellings onto "Z". That is deliberately different, not
+#   drift: every timestamp it compares passes through that one function, so it
+#   is internally consistent and is never compared against a merge-gate value.
+#   Do NOT "unify" it into this library. The parity test pins the divergence so
+#   a future refactor that erases it fails loudly.
+#
+# USAGE
+#   # In merge-gate.sh:
+#   source "$SCRIPT_DIR/lib/ts-normalizer.sh"
+#   if [[ "$(norm_ts "$approved_at")" < "$(norm_ts "$commit_ts")" ]]; then …
+#
+# norm_ts <iso8601-timestamp>
+#   Prints the comparison-safe form. Idempotent — an already-stripped value is
+#   unchanged.
+#   Examples:
+#     norm_ts "2026-08-01T10:00:16Z"          -> "2026-08-01T10:00:16"
+#     norm_ts "2026-08-01T10:00:16+00:00"     -> "2026-08-01T10:00:16"
+#     norm_ts "2026-08-01T10:00:16+0000"      -> "2026-08-01T10:00:16"
+#     norm_ts "2026-08-01T10:00:16.900000Z"   -> "2026-08-01T10:00:16.900000"
+#     norm_ts "2026-08-01T10:00:16-04:00"     -> "2026-08-01T10:00:16-04:00"
+#     norm_ts ""                              -> ""
+
+# Exactly one suffix is stripped per call, matching the jq alternation
+# `sub("(Z|\\+00:00|\\+0000)$"; "")` in escalate-review.sh. The earlier
+# merge-gate.sh form chained three `${t%…}` expansions, which would strip BOTH
+# suffixes off a (never-observed) "…+00:00Z" while the jq mirror strips one —
+# a latent parity gap. No real GitHub timestamp carries two suffixes, so this
+# changes no live comparison.
+norm_ts() {
+  local t="${1-}"
+  case "$t" in
+    *Z)      printf '%s\n' "${t%Z}" ;;
+    *+00:00) printf '%s\n' "${t%+00:00}" ;;
+    *+0000)  printf '%s\n' "${t%+0000}" ;;
+    *)       printf '%s\n' "$t" ;;
+  esac
+}
--- a/.claude/scripts/merge-gate.sh
+++ b/.claude/scripts/merge-gate.sh
@@ -130,7 +130,7 @@ printf '%s\t%s\t%s\n' "$(date -u +%FT%TZ)" "$(basename "$0")" "${*//$'\n'/ }" >>
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 TS_NORMALIZER_LIB="${SCRIPT_DIR}/lib/ts-normalizer.sh"
 if [[ ! -f "$TS_NORMALIZER_LIB" || ! -r "$TS_NORMALIZER_LIB" ]]; then
-  echo "merge-gate.sh: missing sibling library: $TS_NORMALIZER_LIB" >&2
+  echo "merge-gate.sh: sibling library missing or unreadable: $TS_NORMALIZER_LIB" >&2
   exit 4
 fi
 # shellcheck source=./lib/ts-normalizer.sh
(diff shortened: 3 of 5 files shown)


auerbachb:
@cursor review

auerbachb:
@cursor review

Issue #835, opened 2026-07-30
Re-audit the usage-limit signal on runtime upgrade; build the pre-emptive wind-down if one appears


auerbachb:
## Background

Issue #824 asked for a session that detects an approaching usage limit, writes a handoff, and shuts down gracefully. The gating investigation (PR #829) found **no trustworthy upstream signal** reaches a hook, skill, or session on Claude Code **2.1.219**, so the pre-emptive half was deliberately not built. Full evidence, including reproduction commands: `.claude/reference/usage-limit-signal-audit-2026-07.md`.

What shipped instead is a post-hoc recorder (`.claude/hooks/usage-limit-record.sh`, `StopFailure`/`rate_limit`) that leaves a breadcrumb *after* the wall is hit. Useful, but it is not the thing #824 actually wanted.

This issue exists so the original goal is not silently forgotten. The verdict is **version-scoped** — it is expected to change.

## Re-check triggers

Re-run the audit when any of these become true:

1. **A hook event gains usage fields.** The shared hook-payload base currently returns only `{session_id, transcript_path, cwd, prompt_id, permission_mode, agent_id, agent_type, effort}`. This single change would unblock the full pre-emptive wind-down.
2. **statusLine gains a non-TUI invocation path**, or the desktop app starts rendering a status line. Today `rate_limits` reaches only the status line, which never executes in a headless `--output-format stream-json` session.
3. **A proactive warning event ships** (e.g. an `ApproachingLimit` event, or `Notification` carrying `rate_limits`).
4. **A machine-readable usage export appears** (CLI subcommand, env var, or file).

## Acceptance Criteria

- [ ] Re-run the three binary checks in `usage-limit-signal-audit-2026-07.md` against the current runtime and record the version + SHA-256 audited
- [ ] If a trustworthy signal now reaches a hook/skill/session: build the wind-down per #824's original AC — stop accepting new work, let in-flight work reach a safe stopping point, write the handoff via `handoff-state.sh`/`session-state.sh`, emit one plain user message
- [ ] If still no signal: update the audit doc with the new version and close again, leaving the recorder in place
- [ ] `safety.md`'s local-estimation prohibition stays intact either way — the trigger must remain an upstream signal, never a locally-derived estimate

## Notes

- Do **not** rebuild a local spend estimator. #499 rolled one back because agents gated real decisions on it (one declined a correct wave "because quota is critical"), which is why the safety rule exists.
[...]

Issue #84, opened 2026-03-30
Rewrite README as a clear onboarding guide for new users


auerbachb:
## Problem

The current README documents what the config does and how the review loop works, but it doesn't walk a new user through the exact steps they need to take after cloning the repo. Someone landing on this repo for the first time has to piece together the setup from scattered sections.

## Goal

Rewrite the README so a new user can clone the repo and follow a numbered checklist to get everything working with the Claude Code desktop app. The guide should cover:

1. Clone the repo
2. Symlink `CLAUDE.md` → `~/.claude/CLAUDE.md`
3. Symlink `.claude/rules/` → `~/.claude/rules/`
4. Symlink each skill in `.claude/skills/*` → `~/.claude/skills/*`
5. Copy `global-settings.json` → `~/.claude/settings.json` and replace placeholder paths
6. Install prerequisites (Claude Code, `gh` CLI, CodeRabbit, CodeRabbit CLI, optionally Greptile)
7. Verify the setup works

Keep the existing content about what the config does, design decisions, FAQ, etc. — just restructure so onboarding is front and center.

## Acceptance Criteria

- [ ] README has a clear step-by-step "Getting Started" section that a new user can follow from clone to working setup
- [ ] All symlink commands use the correct relative structure (repo → `~/.claude/`)
- [ ] `global-settings.json` → `~/.claude/settings.json` copy + path replacement is clearly documented
- [ ] Skills symlinking is documented (each skill dir, not a single symlink)
- [ ] Prerequisites section lists all required tools with install commands
- [ ] Existing reference content (design decisions, FAQ, slash commands, review loop diagrams) is preserved
- [ ] Hook scripts and their purpose are documented in the setup flow

## Implementation Plan

1. Restructure README with 'Getting Started' as the primary section — a numbered step-by-step checklist from clone to verified setup
2. Document all symlinks: CLAUDE.md, .claude/rules/, and each skill directory individually
3. Document global-settings.json copy + path replacement with sed command
4. Add prerequisites table with install commands
5. Add verification step with commands to check symlinks and settings
6. Preserve existing content (design decisions, FAQ, review loop diagrams, slash commands) below the setup guide
7. Update review loop diagram to show Greptile instead of Macroscope
8. Add per-project override section as optional alternative

PR #910, opened 2026-08-01
feat(#901): /pause — clean wind-down plus a portable, lint-enforced handoff

### LINKED ISSUE #901: Clean pause on demand: a portable handoff I can hand to Cursor
## Background

Long sessions end at a wall. You are three hours into a productive stretch, a couple of pipelines are running, a PR is mid-review, and the 5-hour window (or the weekly one) runs dry. The thread stops, and everything it knew about where things stood stops with it.

Issue #824 went after exactly this and came back with a hard answer: the runtime does surface how much of the 5-hour and weekly allowance is spent, but that number reaches **only** the status line — which never renders in the desktop app at all. It does not reach a hook, a skill, or the session. So the agent cannot notice the wall coming. That verdict is version-scoped and tracked in #835, which re-checks it whenever the runtime changes. What shipped instead was a breadcrumb written *after* the wall is hit — better than silence, but it arrives too late to be a handoff.

There is a second half to this that nobody has built, and it is not blocked on anything upstream. **You** can see the allowance — it is right there in the usage view, which is the authoritative source. What you cannot do today is say "we are getting close, wrap it up cleanly" and get back something you can actually carry somewhere else.

That last part matters more than it used to. When the Claude window runs out, the work does not have to stop — Cursor now imports a Claude thread as a first-class feature. So the natural move when the allowance gets thin is to pause here and pick up there. What makes that move painful is not the import; it is that nothing in this harness produces a document worth importing.

## Problem

Two gaps, one felt as a single frustration.

**There is no clean way to stop.** Ending a heavy session well means remembering, by hand, every loose thread: which PRs are mid-flight and what they are waiting on, which pipelines are running, what the next move was going to be, what you had just decided and why. `/wrap` finishes a merge and `/pm-handoff` sets up a fresh PM thread, but neither is the thing you want when the tank is nearly empty and you want to walk away safely in the next few minutes.

**And what handoffs we do write only work in Claude.** They lean on this repo's own state files and vocabulary — phase names, gate rules, script paths, session-state records. Paste one into Cursor and it reads like instructions to a machine that is not there. The result is that "move my work into Cursor" means reconstructing context by hand at precisely the moment you have no budget left to do it.

The perverse part: the closer you are to the limit, the more valuable the handoff is and the less you can afford to write it.

## Proposed solution

One command you run when you can see the window getting thin — you are the one watching the usage view, so nothing here guesses at a number or decides on your behalf. Ask for a clean pause, and you get three things.

**A safe stopping point.** It stops picking up new work, lets whatever is running reach a place it can be left, and tells you plainly what it stopped and what it let finish. Nothing is left half-pushed or orphaned.

**A document you can take anywhere.** Self-contained and plainly written — what you were working on and why, what is in flight and what each thing is waiting on, decisions made in the session and the reasoning behind them, and an explicit "start here" for whoever picks it up. Crucially it is written for a reader that has never heard of this harness: no phase names, no script invocations, no assumed state files. Someone — or something — with just the repo and this document can continue. Cursor is the case that prompted it, but a portable document is portable everywhere, and that is what makes it worth writing once.

**A short summary in the thread**, so the pause is visible rather than something you discover later.

Two things this deliberately is not. It is not a token estimator — the human reads the authoritative usage view and decides; the agent never infers a budget or refuses work over one, so the existing safety prohibition stands untouched and unamended. And it is not waiting on #835: when a real approaching-limit signal eventually reaches a hook, the wind-down that #835 builds emits *this* document. Building it now means that day is a wiring change instead of a design problem.

Worth folding in while we are here: the after-the-wall breadcrumb currently points at nothing in particular. If a portable handoff exists, it should point at the most recent one — so even the session that dies without warning leaves you somewhere useful.

## Acceptance Criteria

- [ ] A user-invoked command exists (working name `/pause`; final name is an open question below) that produces a portable handoff document on demand — no schedule, no automatic trigger, no inferred budget
- [ ] The command is invocable from any thread type (coding, PM, monitoring, capture) and degrades gracefully when a category is absent (e.g. no in-flight PRs)
- [ ] **Portability is testable, not aspirational:** the emitted document contains no `.claude/` script invocations, no phase-letter vocabulary (`Phase A/B/C`, `merge_ready`), no `session-state.json` / handoff-file paths, and no skill names as required steps — verified by a grep-based check in the test plan
- [ ] The document captures, in prose a non-Claude agent can act on: current objective; every in-flight PR/issue with its number, URL, and what it is blocked on; decisions made this session with rationale; explicit "do this first" next step; and any uncommitted or unpushed local state
- [ ] The document is written to a predictable path under `~/.claude/handoffs/` **and** printed in-thread in a single fenced block, so it can be copied without a file read
- [ ] Wind-down behavior before the document is written: stop launching new subagents/pipelines, allow in-flight work to reach a safe stopping point, and report what was stopped vs. allowed to finish
- [ ] Existing in-repo state (`session-state.json`, `~/.claude/handoffs/`) is read through `session-state.sh` / `handoff-state.sh` per `handoff-files.md` — never inline `jq` read-modify-write
- [ ] No code path computes or consults a local token/spend estimate to gate any decision; `.claude/rules/safety.md` "Anthropic Quota & Spend Authority" is left **unchanged** — verified by grep over the change set
- [ ] `.claude/hooks/usage-limit-record.sh` `resume_hint` points at the most recent portable handoff when one exists, and degrades to its current behavior when none does
- [ ] #835's Acceptance Criteria updated (or a comment added) noting that the wind-down it builds should emit this document rather than define its own
- [ ] Rule-corpus word budget respected (`rule-lint.sh`) if any `CLAUDE.md` / `.claude/rules/` file is touched; prefer `.claude/reference/` for mechanism detail

## Test Plan

- [ ] Invoke the command in a session with 2+ open PRs and an active subagent → verify it reports what it stopped vs. let finish, and that no new subagent launches afterward
- [ ] Invoke it in a clean session with nothing in flight → verify it still emits a valid document rather than erroring or emitting an empty shell
- [ ] **Portability grep:** run the emitted document through a check for `\.claude/`, `Phase [ABC]`, `merge_ready`, `session-state`, `handoff-state`, and `/[a-z-]+` skill invocations → zero hits in the required-action sections
- [ ] **Cold-read test:** hand the document to a fresh agent with repo access and no harness rules loaded (a plain `claude -p` run or a Cursor thread) and confirm it can state what to do first and which PRs are open, using the document alone
- [ ] Verify the document is both written to disk and printed in-thread, and that the two match
- [ ] Simulate a `rate_limit` StopFailure with a portable handoff on disk → confirm `resume_hint` names it; repeat with none on disk → confirm current behavior is unchanged
- [ ] `grep` the change set for local spend/token estimation gating a decision → zero hits; confirm `safety.md` is untouched in the diff
- [ ] All hook test suites pass; `rule-lint.sh` passes on any corpus edit

## Notes / Open questions

- **Command name.** `/pause` is the working name. `/sunset`, `/handoff`, `/eject`, `/carry-out` are candidates. `/sundown` was the name in #466, which was closed NOT_PLANNED in favor of folding that sweep into `/wrap` — reusing the name would be confusing, so it is off the table.
- **Relationship to `/wrap` and `/pm-handoff`.** `/wrap` is end-of-*PR* (merge, sweep, lessons); `/pm-handoff` is Claude-thread-to-Claude-thread continuity. This is end-of-*allowance* and cross-*tool*. The three overlap in what they read but not in what they produce. Open question: should this reuse `/pm-handoff`'s state-gathering internals rather than re-implement them, and just differ in the rendering step? Leaning yes — one collector, two renderers (Claude-native and portable).
- **How much does "let in-flight work reach a safe stopping point" cost?** Letting a Phase A subagent finish may cost tokens the window does not have. #824 leaned toward: stop launching new ones, let running ones write their handoffs, then stop. Same call likely applies here, but it is worth confirming rather than inheriting.
- **Do not build a token estimator.** #499 rolled one back precisely because agents started gating real decisions on it — one declined a correct wave "because quota is critical." The human-as-sensor design is what keeps this issue on the right side of that line, and it is not a detail to optimize away later.
- **The Cursor import path is the motivating case, not a dependency.** The document should be plain Markdown that any agent or human can act on; nothing should hard-couple to one vendor's import format, which may change.

## Related Issues

- Follows #824 (closed on the no-signal finding; shipped the post-hoc recorder)
- Related to #835 (re-audit + pre-emptive wind-down — this issue supplies the document that wind-down should emit)
- Related to #499 (the `/quota` rollback that produced the local-estimation prohibition)
- Related to #710 (spend telemetry — measurement, not a limit signal)

## Related Files

- `.claude/rules/safety.md` (deliberately unchanged — "Anthropic Quota & Spend Authority")
- `.claude/scripts/handoff-state.sh`
- `.claude/skills/pm-handoff/SKILL.md`
- `.claude/skills/wrap/SKILL.md`
- `.claude/hooks/usage-limit-record.sh`
- `.claude/reference/usage-limit-signal-audit-2026-07.md`

_Captured via /issue-maker._

## Implementation Plan

_Merged: Claude's codebase analysis + CodeRabbit's plan (posted 2026-08-01T15:40:51Z). CR's four design choices are adopted; the deltas below are where this plan diverges, each with its reason._

### Command name

`/pause` (CR design choice 1, the ticket's working name). The skill directory and README row are the only places the name is hardcoded, so a later rename is a two-file change.

### Deltas from CodeRabbit's plan

1. **The portability contract is enforced by a real script, not a prose self-check.** CR put the grep inside the skill body as an instruction ("include a self-check that greps…"). An instruction an agent may or may not follow is exactly what the issue's "testable, not aspirational" criterion rules out. It becomes `.claude/scripts/portable-handoff-lint.sh` with its own auto-discovered test suite, and the skill is gated on its exit code.
2. **Slash-command detection reads the actual skill catalog** rather than the literal `/[a-z-]+` from the Test Plan. That regex matches `/tmp`, `/usr`, and every lowercase leading path segment, so it would fail every document that names a real filesystem location. Enumerating `.claude/skills/*/` plus a short built-in list detects exactly "skill names as required steps" with no false positives.
3. **Enforcement is section-scoped**, matching the issue's own Test Plan wording ("zero hits **in the required-action sections**"). This resolves a real conflict: the uncommitted-work location the reader needs is an absolute path that necessarily contains `.claude/worktrees/`. Harness *vocabulary* (phase letters, `merge_ready`, state-file names, harness script paths, skill invocations) is banned document-wide; only a `.claude/worktrees/` path inside the local-state section is exempt, and only there.
4. **A quota-authority guard test ships with the change.** The issue makes a silent contradiction with `safety.md` a blocking failure, so it gets a test rather than a promise: the verbatim prohibition must still be present in `safety.md`, and the executable code of every new artifact is grepped for spend/estimation vocabulary.
5. **`refill.reason` reuses the existing bounded enum value `full_stop`** rather than adding a `pause_command` value. The semantics are identical (a full stop that only an explicit human resume clears), and the enum is bounded on purpose — a new value would need handling in every future reader for zero behavioral gain.
6. **Golden + negative fixtures** in the lint test, so each violation class is proven caught rather than assumed caught.

### Files to create

| Path | Purpose |
|------|---------|
| `.claude/skills/pause/SKILL.md` | The command: wind-down → collect → render → verify → emit |
| `.claude/skills/pause/references/portable-handoff-template.md` | Document layout + rendering rules (follows the `wrap/references/` pattern) |
| `.claude/scripts/portable-handoff-lint.sh` | The enforceable portability check |
| `.claude/scripts/tests/portable-handoff-lint.test.sh` | Auto-discovered suite: golden pass, one negative fixture per violation class, quota-authority guard |
| `.claude/reference/session-state-collector.md` | The one shared collector both `/pm-handoff` and `/pause` consume |
| `.claude/reference/portable-handoff.md` | Artifact convention: naming, format, atomic single-writer write, why it is a sibling to the PR-scoped JSON handoffs |

### Files to modify

| Path | Change |
|------|--------|
| `.claude/skills/pm-handoff/SKILL.md` | Steps 4 / 5b / 5b2 / 5c point at the shared collector; rendering stays put. Behavior unchanged. |
| `.claude/hooks/usage-limit-record.sh` | `resume_hint` names the most recent portable handoff when one exists; byte-identical when none does |
| `.claude/hooks/tests/usage-limit-record.test.sh` | Two new cases (handoff present / absent); every existing assertion intact |
| `README.md` | One `/pause` row; both hardcoded counts 29 → 30 |
| `.claude/reference/README.md` | Index rows for the two new reference files |

Issue #835 gets a comment (not an AC rewrite) noting its wind-down should emit this document rather than define its own.

### Implementation steps

1. **`portable-handoff-lint.sh`** — takes a file path, parses `## ` headings, classifies each section, and reports every violation with file:line and a named rule. Rules: `harness-path` (`.claude/{scripts,hooks,rules,reference,agents,skills}/`), `phase-vocabulary` (`Phase A/B/C`, `merge_ready`, `phase-{a,b,c}-*`), `state-file` (`session-state`, `handoff-state`, `pr-N-handoff`), `skill-invocation` (catalog-driven), and `required-sections` (every required heading present and non-empty — this is what makes "never an empty shell" checkable). Exit 0 clean, 1 violations, 2 usage, 3 unreadable file. `--help` documents the contract.
2. **`portable-handoff-template.md`** — fixed H2 headings so the lint can classify sections: `Start here`, `What we're working on`, `Open work`, `Decisions made this session`, `Local state on this machine`. Fenced literal template with `{PLACEHOLDER}` tokens, then `## Rendering rules` — including the plain-English translations for phase/gate concepts and the requirement that the printed block and the on-disk file are identical.
3. **`session-state-collector.md`** — the reads only, no rendering: live GitHub state, repo-scoped PR list via `--session-view` (never `--get .`), per-PR handoff reads, active agents, polling jobs, memory index. Lock-free, helper-scripts-only, no inline `jq` read-modify-write.
4. **`pause/SKILL.md`** — frontmatter leading with when to invoke; then: wind-down (`refill.paused=true` + `reason`/`scope`/`at` in one atomic `--set`, never killing running units), report stopped-vs-allowed-to-finish, collect, render, **run the lint and fix any hit before emitting**, then atomically write to `~/.claude/handoffs/portable-handoff-{SESSION_ID}-{TIMESTAMP}.md` and print the identical content. Graceful degradation per section. An explicit statement that the human reads the authoritative usage view — the skill never estimates.
5. **`usage-limit-record.sh`** — `newest_handoff()` picks the newest `portable-handoff-*.md` by mtime using bash `-nt` (no `stat` portability switch), ties broken by name. Appends one sentence naming it to `resume_hint` and adds a `portable_handoff` field. All existing invariants preserved.
6. **Tests** — the new lint suite plus the two recorder cases; then `run-hook-tests.sh`, `rule-lint.sh`, `skill-catalog-lint.sh`, `verbatim-block-lint.sh` locally.

### Risks and edge cases

- **The worktree-path conflict** (delta 3) is the one place the portability rule and the document's usefulness genuinely collide. Section scoping is the resolution; the lint encodes it so it cannot drift back into prose.
- **`/pm-handoff` regression.** The collector extraction is a refactor of a working skill. Mitigation: the reference holds the collection bash verbatim; only rendering stays in `pm-handoff`.
- **A naive slash-command regex would false-positive** on any absolute path (delta 2) — the reason for the catalog-driven rule.
- **Lint self-reference.** The template and the lint's own `--help` name the banned patterns, so neither is linted; only rendered documents are.
- **`refill.paused` is a real user-visible stop.** `/pause` is a human typing in chat, which is exactly the authorized writer of that field. It stays paused until an explicit human resume, and the wind-down report says so out loud.

### Verification

Each AC maps to a check: portability → the lint plus its negative fixtures; empty-shell → the `required-sections` rule; disk-equals-thread → the skill writes then prints one buffer; recorder → the two new hook cases; quota authority → the guard test asserting `safety.md` verbatim and grepping the new artifacts' executable code; corpus budget → `rule-lint.sh` (no `CLAUDE.md` or `.claude/rules/` edits are planned, so this should be a no-op).

auerbachb:
## What this does

Adds `/pause` — a command you run when you can see your usage allowance getting thin, or when you want to carry a session's work into another tool.

It does two things. It **winds down** without killing anything: new launches stop, running units are left to reach their own stopping point, and the report says plainly what was stopped and what was allowed to finish. Then it writes a **portable handoff document** — plain Markdown for a reader who has the repo and nothing else. No phase vocabulary, no state-file paths, no skill names as steps. Written to disk and printed byte-identical in the thread.

You decide when. Nothing here estimates tokens, spend, or quota.

## The part that matters: portability is enforced

The obvious way to build this is a paragraph in the skill saying "write it portably." That would not have worked. The failure is silent — a document saying "resume at Phase B" still looks like a handoff, still writes successfully, and only fails much later, for a reader who by then has no context to repair it with.

So `portable-handoff-lint.sh` gates the emit. Seven rules: harness paths, phase vocabulary, state-file names, skill invocations, unfilled template placeholders, missing-or-empty required sections, and a working directory that must be an absolute path. Exit 1 and the document does not ship as written.

Two judgment calls worth flagging:

**Skill-invocation detection reads the live skill catalog** instead of the ticket's literal `/[a-z-]+`. That pattern matches `/tmp`, `/usr`, and every lowercase leading path segment — it would fail every document that names a real location, which is every useful handoff. Enumerating `.claude/skills/*/` catches "a skill name used as a required step" exactly, with no false positives, and keeps working as skills are added.

**One `.claude/` form is permitted:** an *absolute* path containing `/.claude/worktrees/`. That is the literal address of the reader's uncommitted work. The leading slash is the whole rule — a bare `.claude/worktrees/…` is a this-checkout-only pointer and still fails. My first draft scoped this exemption by section instead, and writing the golden fixture immediately showed why that was wrong: the path belongs in "Start here" too.

## Two pre-existing bugs fixed in passing

Both were in `/pm-handoff` collection code this PR moves into a shared collector, so they were about to gain a second consumer:

- **`gh pr list` had no `--limit`.** It defaults to 30, so every handoff silently dropped the 31st open PR onward while reading as complete.
- **The per-PR handoff lookup was unscoped.** Two repos both tracking PR #84 made `head -1` a coin flip; picking the other repo's entry made the payload guard skip the PR entirely, so *this* repo's handoff silently vanished from the report.

## Quota authority is untouched

`.claude/rules/safety.md` §"Anthropic Quota & Spend Authority" is unamended — `git diff` against it is zero lines. The human reads the authoritative usage view and invokes the command; no code path computes or consults a local estimate.

That is asserted by a test rather than by this paragraph: the rule text must still be present verbatim, and no shipped executable code — the checker, the recorder, or the skill's bash blocks — may carry spend/estimation vocabulary.

## Also

- **One collector, two renderers.** `/pm-handoff` and `/pause` now share `.claude/reference/session-state-collector.md` and differ only in rendering. `/pm-handoff`'s output shape is unchanged.
- **The breadcrumb now points somewhere.** `usage-limit-record.sh` names the most recent portable handoff in `resume_hint` (newest by mtime, filesystem lookup only, never opens the file), and is byte-identical when none exists. A session that dies without warning lands somewhere useful.
- **Issue #835 aligned** by comment: when a real approaching-limit signal appears, the wind-down it builds emits *this* document rather than defining its own.

## Local review coverage: `none` (degraded)

Both local CLIs ended the session unavailable, so this had **no verified clean CLI pass**:

- **CodeAnt CLI** — HTTP 403 on every batch (the undocumented daily cap), retried once per the rules, then dropped. `meta.capped: true`.
- **CodeRabbit CLI** — two full reviews completed and **all 15 findings were verified and addressed** (9 in round one, 6 in round two, of which 5 were fixed and 1 declined with reasoning below). The third pass hit the free-OSS tier cap: `rate_limit`, 46-minute lockout. That last run is a failed run, not a clean pass.

A self-review stood in, and caught two real defects the CLIs had not: `/pause` referenced helper scripts by repo-relative path despite being invocable from any thread, and the "checker could not run" branch was unreachable because an unresolved checker exits 127, not 2 or 3.

CLI quotas are independent of the GitHub App reviewers, so the merge gate is unaffected.

**Declined finding, with reasoning:** CodeAnt-style byte caps on handoff-file and memory-index reads. The collector's contract is to *extract named fields*, not to pass files through, and a blanket truncation risks dropping exactly the `remaining_work` note the handoff exists to carry. Made the extract-don't-dump rule explicit instead.

## Test plan

- [x] `portable-handoff-lint.sh` catches every violation class — one negative fixture per rule, each asserted against the specific rule id
- [x] It does **not** fire on what a real handoff contains: absolute paths, `/tmp`, `/usr/local/bin`, GitHub URLs, `git status`, `gh pr view`
- [x] The worktree exemption is exactly as narrow as documented — absolute form passes, `repo/.claude/worktrees/…`, `./.claude/worktrees/…`, and `/…/.claude/scripts/…` all fail
- [x] A missing skill catalog fails **closed** (falls back to shape matching) rather than silently checking one rule fewer
- [x] Structural rules are real: a missing required section, an all-headings-no-content shell, and a relative/empty/missing working directory each fail
- [x] The checker's required-section list and the template's headings agree (drift between them fails the suite)
- [x] A real session renders a document that passes every rule — four pull requests with numbers, URLs and blockers, a concrete first action, decisions with reasoning, branch and absolute working directory; zero harness-vocabulary hits
- [x] The emit path renders **one** buffer, checks that buffer, and publishes that same buffer — verified end-to-end on a real handoff built from this session's live state (2,960 bytes, sha256 `db482659…`, mode 644). The in-thread print is the same buffer by construction; what is mechanically verified here is the single-render/atomic-publish path, not a live `/pause` invocation
- [x] `resume_hint` names the most recent portable handoff when one exists — newest by mtime, ignoring non-handoff files and in-progress drafts
- [x] `resume_hint` is **byte-identical** to its previous value when no handoff exists (asserted as full equality, not a substring)
- [x] These recorder tests fail against the pre-change hook (verified by reverting it)
- [x] Zero local spend/token estimation in shipped executable code; `safety.md` diff is zero lines
- [x] All 59 bash test suites pass; `rule-lint`, `skill-catalog-lint`, and `verbatim-block-lint` pass
- [x] Rule corpus unchanged at 11,735 words (no `CLAUDE.md` or `.claude/rules/` edits)

**Deferred to [Issue #912](https://github.com/auerbachb/claude-code-config/issues/912):** the cold-read test — handing the document to an agent with no harness rules loaded. The `claude` CLI is not installed on this machine and a headless run needs interactive auth, so it was not run rather than claimed. The lint proves the *absence* of harness jargon; only a real reader can prove the *presence* of enough information, and that distinction is why it is a manual test rather than another rule.

Closes #901

## Summary by CodeRabbit

* **New Features**
  * Added the `/pause` command to stop new work, save session state, and create a portable handoff document.
  * Portable handoffs now include session details, local state, next steps, and validation.
  * Usage records can reference recent portable handoffs and flag repository mismatches.

* **Documentation**
  * Added guidance for portable handoffs and shared session-state collection.
  * Updated handoff workflows and command documentation.

* **Quality Improvements**
  * Added validation and extensive coverage for handoff formatting, discovery, retention, and concurrent usage-record updates.

___

## **CodeAnt-AI Description**
Add `/pause` for clean wind-downs and portable session handoffs

### What Changed
- Added `/pause`, which stops new work without interrupting running tasks and reports what was stopped, what is finishing, and whether the pause was recorded.
- Generates a self-contained Markdown handoff for use in another tool or fresh session, including current work, open items, decisions, and local changes.
- Validates handoffs before publishing them, rejecting internal harness references, unresolved placeholders, incomplete sections, and non-actionable paths.
- Publishes handoffs atomically and prints the same verified document in the conversation.
- Usage-limit records now point to the newest portable handoff from the last seven days, while preserving the existing transcript-only behavior when none is available.
- Shared session-state collection now supports both portable `/pause` documents and native `/pm-handoff` prompts.
- Added comprehensive tests for portability checks, handoff selection, stale files, concurrent usage-limit records, and failure-safe behavior.

### Impact
`✅ Clean session wind-down without killing running work`
`✅ Portable handoffs for fresh sessions and other tools`
`✅ Fewer lost-context recovery steps after usage limits`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

### DIFF
--- a/.claude/hooks/tests/usage-limit-record.test.sh
+++ b/.claude/hooks/tests/usage-limit-record.test.sh
@@ -372,4 +372,21 @@ for bad in "not-a-number" "" "-3" "7; rm -rf /" "0"; do
 done
 [[ -e / ]] || fail "the injection fixture did something catastrophic"
 
+# 15i. The durable record is written BEFORE the pointer lookup, so no amount of
+# slowness in the optional enrichment can cost the user the record. Assert the
+# ordering structurally: the events append must appear before the lookup call.
+APPEND_LINE=$(grep -n '>>"\$EVENTS_LOG"' "$HOOK" | head -1 | cut -d: -f1)
+LOOKUP_LINE=$(grep -n '^PORTABLE_HANDOFF=' "$HOOK" | head -1 | cut -d: -f1)
+[[ -n "$APPEND_LINE" && -n "$LOOKUP_LINE" ]] \
+  || fail "could not locate the append and lookup lines to check their ordering"
+(( APPEND_LINE < LOOKUP_LINE )) \
+  || fail "the handoff lookup (line $LOOKUP_LINE) runs before the durable append (line $APPEND_LINE) — a slow lookup could consume the hook's timeout and lose the record"
+
+# And behaviourally: the events log carries the base record while
+# usage-limit-last.json carries the enriched pointer.
+jq -e --arg h "$BASE_HINT" '.resume_hint == $h' <"$SOME_DIR/usage-limit-events.jsonl" >/dev/null 2>&1 \
+  || true   # first line only; the assertion below is the load-bearing one
+jq -e '.portable_handoff != null' "$SOME_LAST" >/dev/null \
+  || fail "usage-limit-last.json lost its handoff pointer after the phase split"
+
 echo "PASS: usage-limit-record.sh"
--- a/.claude/hooks/tests/usage-limit-record.test.sh
+++ b/.claude/hooks/tests/usage-limit-record.test.sh
@@ -389,4 +389,31 @@ jq -e --arg h "$BASE_HINT" '.resume_hint == $h' <"$SOME_DIR/usage-limit-events.j
 jq -e '.portable_handoff != null' "$SOME_LAST" >/dev/null \
   || fail "usage-limit-last.json lost its handoff pointer after the phase split"
 
+# 15j. Concurrency with a handoff present: `last` must still agree with the
+# newest event. Splitting the write into two phases opened the door for a slow
+# invocation to finish its enrichment after a newer one published and roll
+# `last` back to an older session — breaking the one thing that file promises.
+# Concurrency is the EXPECTED case here: one account limit fails every active
+# session at once.
+RACE_DIR="$TMP_DIR/race"
+mkdir -p "$RACE_DIR/handoffs"
+RACE_HANDOFF="$RACE_DIR/handoffs/portable-handoff-20260801T120000Z-sess.md"
+printf 'handoff\n' >"$RACE_HANDOFF"
+touch_ago "$RACE_HANDOFF" 300
+RACE_N=8
+for i in $(seq 1 "$RACE_N"); do
+  printf '%s' "{\"hook_event_name\":\"StopFailure\",\"error\":\"rate_limit\",\"session_id\":\"race$i\"}" \
+    | CLAUDE_USAGE_LIMIT_DIR="$RACE_DIR" CLAUDE_HANDOFF_DIR="$RACE_DIR/handoffs" bash "$HOOK" &
+done
+wait
+
+RACE_LOG="$RACE_DIR/usage-limit-events.jsonl"
+RACE_COUNT=$(wc -l <"$RACE_LOG" | tr -d ' ')
+[[ "$RACE_COUNT" == "$RACE_N" ]] || fail "concurrent writes with a handoff lost records: got $RACE_COUNT, expected $RACE_N"
+
+RACE_LAST_SID=$(jq -r '.session_id' "$RACE_DIR/usage-limit-last.json")
+RACE_TAIL_SID=$(tail -1 "$RACE_LOG" | jq -r '.session_id')
+[[ "$RACE_LAST_SID" == "$RACE_TAIL_SID" ]] \
+  || fail "usage-limit-last.json ($RACE_LAST_SID) disagrees with the newest event ($RACE_TAIL_SID) — an older invocation's phase 2 clobbered a newer record"
+
 echo "PASS: usage-limit-record.sh"
--- a/.claude/hooks/tests/usage-limit-record.test.sh
+++ b/.claude/hooks/tests/usage-limit-record.test.sh
@@ -355,4 +355,21 @@ printf '%s' '{"hook_event_name":"StopFailure","error":"rate_limit","session_id":
 jq -e --arg p "$FRESH_HANDOFF" '.portable_handoff == $p' "$AGED_LAST" >/dev/null \
   || fail "the age bound suppressed a fresh handoff too"
 
+# 15h. A bad age setting must not cost the user the pointer. Passed straight to
+# `find`, a non-numeric value makes it fail — and its stderr is discarded, so a
+# perfectly good handoff would vanish from the breadcrumb with no trace.
+BADAGE_DIR="$TMP_DIR/bad-age"
+mkdir -p "$BADAGE_DIR/handoffs"
+BADAGE_HANDOFF="$BADAGE_DIR/handoffs/portable-handoff-present.md"
+printf 'present\n' >"$BADAGE_HANDOFF"
+touch_ago "$BADAGE_HANDOFF" 300
+for bad in "not-a-number" "" "-3" "7; rm -rf /" "0"; do
+  printf '%s' '{"hook_event_name":"StopFailure","error":"rate_limit","session_id":"badage"}' \
+    | CLAUDE_USAGE_LIMIT_DIR="$BADAGE_DIR" CLAUDE_HANDOFF_DIR="$BADAGE_DIR/handoffs" \
+      CLAUDE_HANDOFF_MAX_AGE_DAYS="$bad" bash "$HOOK"
+  jq -e --arg p "$BADAGE_HANDOFF" '.portable_handoff == $p' "$BADAGE_DIR/usage-limit-last.json" >/dev/null \
+    || fail "a fresh handoff was lost with CLAUDE_HANDOFF_MAX_AGE_DAYS='$bad'"
+done
+[[ -e / ]] || fail "the injection fixture did something catastrophic"
+
 echo "PASS: usage-limit-record.sh"
--- a/.claude/hooks/tests/usage-limit-record.test.sh
+++ b/.claude/hooks/tests/usage-limit-record.test.sh
@@ -421,4 +421,22 @@ RACE_TAIL_SID=$(tail -1 "$RACE_LOG" | jq -r '.session_id')
 [[ "$RACE_LAST_SID" == "$RACE_TAIL_SID" ]] \
   || fail "usage-limit-last.json ($RACE_LAST_SID) disagrees with the newest event ($RACE_TAIL_SID) — an older invocation's phase 2 clobbered a newer record"
 
+# 15k. Newest-by-mtime holds beyond any window: the newest file is given the
+# lexically SMALLEST name among many, so a name-ordered scan that truncated
+# would miss it. This is the case the removed scan cap got wrong.
+MANY_DIR="$TMP_DIR/many"
+mkdir -p "$MANY_DIR/handoffs"
+# 10# forces base 10: seq -w zero-pads, and bash reads a leading zero as octal,
+# so "008" would abort the arithmetic mid-loop.
+for i in $(seq -w 1 120); do
+  printf 'x\n' >"$MANY_DIR/handoffs/portable-handoff-2026080$(( 10#$i % 9 + 1 ))T120000Z-s$i.md"
+done
+MANY_NEWEST="$MANY_DIR/handoffs/portable-handoff-00000000T000000Z-aaa.md"
+printf 'newest\n' >"$MANY_NEWEST"
+touch_ago "$MANY_NEWEST" 60          # newest mtime, lexically smallest name
+printf '%s' '{"hook_event_name":"StopFailure","error":"rate_limit","session_id":"many"}' \
+  | CLAUDE_USAGE_LIMIT_DIR="$MANY_DIR" CLAUDE_HANDOFF_DIR="$MANY_DIR/handoffs" bash "$HOOK"
+jq -e --arg p "$MANY_NEWEST" '.portable_handoff == $p' "$MANY_DIR/usage-limit-last.json" >/dev/null \
+  || fail "newest-by-mtime lost among many handoffs (got $(jq -r '.portable_handoff' "$MANY_DIR/usage-limit-last.json"))"
+
 echo "PASS: usage-limit-record.sh"
--- a/.claude/hooks/tests/usage-limit-record.test.sh
+++ b/.claude/hooks/tests/usage-limit-record.test.sh
@@ -382,10 +382,15 @@ LOOKUP_LINE=$(grep -n '^PORTABLE_HANDOFF=' "$HOOK" | head -1 | cut -d: -f1)
 (( APPEND_LINE < LOOKUP_LINE )) \
   || fail "the handoff lookup (line $LOOKUP_LINE) runs before the durable append (line $APPEND_LINE) — a slow lookup could consume the hook's timeout and lose the record"
 
-# And behaviourally: the events log carries the base record while
-# usage-limit-last.json carries the enriched pointer.
-jq -e --arg h "$BASE_HINT" '.resume_hint == $h' <"$SOME_DIR/usage-limit-events.jsonl" >/dev/null 2>&1 \
-  || true   # first line only; the assertion below is the load-bearing one
+# And behaviourally: every events-log line carries the BASE record (phase 1),
+# while usage-limit-last.json carries the enriched pointer (phase 2).
+while IFS= read -r ev_line; do
+  [[ -n "$ev_line" ]] || continue
+  printf '%s' "$ev_line" | jq -e --arg h "$BASE_HINT" '.resume_hint == $h' >/dev/null \
+    || fail "an events-log record carries an enriched hint; phase 1 must write the base record"
+  printf '%s' "$ev_line" | jq -e '.portable_handoff == null' >/dev/null \
+    || fail "an events-log record carries a handoff pointer; only usage-limit-last.json is refined"
+done <"$SOME_DIR/usage-limit-events.jsonl"
 jq -e '.portable_handoff != null' "$SOME_LAST" >/dev/null \
   || fail "usage-limit-last.json lost its handoff pointer after the phase split"
 
--- a/.claude/hooks/tests/usage-limit-record.test.sh
+++ b/.claude/hooks/tests/usage-limit-record.test.sh
@@ -429,11 +429,13 @@ mkdir -p "$MANY_DIR/handoffs"
 # 10# forces base 10: seq -w zero-pads, and bash reads a leading zero as octal,
 # so "008" would abort the arithmetic mid-loop.
 for i in $(seq -w 1 120); do
-  printf 'x\n' >"$MANY_DIR/handoffs/portable-handoff-2026080$(( 10#$i % 9 + 1 ))T120000Z-s$i.md"
+  f="$MANY_DIR/handoffs/portable-handoff-2026080$(( 10#$i % 9 + 1 ))T120000Z-s$i.md"
+  printf 'x\n' >"$f"
+  touch_ago "$f" 7200                # all the decoys are two hours old
 done
 MANY_NEWEST="$MANY_DIR/handoffs/portable-handoff-00000000T000000Z-aaa.md"
 printf 'newest\n' >"$MANY_NEWEST"
-touch_ago "$MANY_NEWEST" 60          # newest mtime, lexically smallest name
+touch_ago "$MANY_NEWEST" 60          # newest mtime, lexically SMALLEST name
 printf '%s' '{"hook_event_name":"StopFailure","error":"rate_limit","session_id":"many"}' \
   | CLAUDE_USAGE_LIMIT_DIR="$MANY_DIR" CLAUDE_HANDOFF_DIR="$MANY_DIR/handoffs" bash "$HOOK"
 jq -e --arg p "$MANY_NEWEST" '.portable_handoff == $p' "$MANY_DIR/usage-limit-last.json" >/dev/null \
--- a/.claude/hooks/tests/usage-limit-record.test.sh
+++ b/.claude/hooks/tests/usage-limit-record.test.sh
@@ -22,8 +22,13 @@ cleanup() { rm -rf "$TMP_DIR"; }
 trap cleanup EXIT
 
 run_hook() {
-  # $1 = JSON payload. Isolated output dir so tests never touch ~/.claude.
-  printf '%s' "$1" | CLAUDE_USAGE_LIMIT_DIR="$TMP_DIR" bash "$HOOK"
+  # $1 = JSON payload. Both dirs are isolated so tests never touch ~/.claude:
+  # CLAUDE_USAGE_LIMIT_DIR for the hook's own records, CLAUDE_HANDOFF_DIR for
+  # the portable-handoff lookup. Leaving the latter unset would read the
+  # developer's real ~/.claude/handoffs, making resume_hint depend on whether
+  # they happen to have run /pause — green here, different there.
+  printf '%s' "$1" \
+    | CLAUDE_USAGE_LIMIT_DIR="$TMP_DIR" CLAUDE_HANDOFF_DIR="$TMP_DIR/handoffs" bash "$HOOK"
 }
 
 EVENTS="$TMP_DIR/usage-limit-events.jsonl"
@@ -157,7 +162,7 @@ mkdir -p "$CONC_DIR"
 N=25
 for i in $(seq 1 "$N"); do
   printf '%s' "{\"hook_event_name\":\"StopFailure\",\"error\":\"rate_limit\",\"session_id\":\"s$i\"}" \
-    | CLAUDE_USAGE_LIMIT_DIR="$CONC_DIR" bash "$HOOK" &
+    | CLAUDE_USAGE_LIMIT_DIR="$CONC_DIR" CLAUDE_HANDOFF_DIR="$CONC_DIR/handoffs" bash "$HOOK" &
 done
 wait
 
@@ -181,7 +186,7 @@ STRUCT_DIR="$TMP_DIR/struct"
 mkdir -p "$STRUCT_DIR"
 BIG_VAL=$(printf 'y%.0s' $(seq 1 4000))
 printf '%s' "{\"hook_event_name\":\"StopFailure\",\"error\":\"rate_limit\",\"error_details\":{\"blob\":\"$BIG_VAL\"},\"last_assistant_message\":{\"nested\":\"$BIG_VAL\"}}" \
-  | CLAUDE_USAGE_LIMIT_DIR="$STRUCT_DIR" bash "$HOOK"
+  | CLAUDE_USAGE_LIMIT_DIR="$STRUCT_DIR" CLAUDE_HANDOFF_DIR="$STRUCT_DIR/handoffs" bash "$HOOK"
 
 STRUCT_LAST="$STRUCT_DIR/usage-limit-last.json"
 [[ -f "$STRUCT_LAST" ]] || fail "structured payload produced no record"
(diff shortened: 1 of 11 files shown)


auerbachb:
Fixed in 5836db0. Both Step 1 call sites are now guarded on $SESSION_STATE_SH being non-empty. An unresolved helper sets WINDDOWN_PERSISTED=0, so Step 2 reports the pause as *not* recorded instead of printing a "Stopped:" line that claims otherwise — and the session-view read reports "could not read what was running" rather than "nothing was running", which are different claims. Good catch: Step 0 promised non-fatal and Step 1 did not honor it.

auerbachb:
Fixed in 5836db0. The collector now carries a "Resolving the helper scripts" section: a consumer that already resolved those paths must substitute its resolved paths, and one without a resolver should use the same three-candidate lookup. This mattered more than it looks — an unresolved path fails silently, so the handoff would have reported empty categories as though they were genuinely empty, which is the one failure mode a handoff must never have.

auerbachb:
Fixed in 5836db0. The handoff location is its own concept, so it now has its own override: CLAUDE_HANDOFF_DIR, defaulting to ~/.claude/handoffs, instead of being derived from the recorder's output dir. Deriving it meant that setting CLAUDE_USAGE_LIMIT_DIR — which exists purely to isolate the recorder's own records — pointed the lookup at a directory no handoff is ever written to. Separating them also exposed a hermeticity leak in the tests, where invocations that did not set the new variable read the developer's real ~/.claude/handoffs; every invocation now isolates both dirs, and a new case pins the property (a decoy handoff under the records dir must lose to the real one under the handoff dir).

auerbachb:
@codeant-ai review

auerbachb:
@cursor review

auerbachb:
Fixed in 48845ac. The fallback now uses {1,}. You are right that this defeated the fail-closed intent in exactly the case it exists for — the shortest skill names are the most common ones, so a 3-character floor let /pm and /go through the one branch whose whole job is to be over-broad. Added fixtures for both.

auerbachb:
Fixed in 48845ac, and the fix is broader than the report. Sanitizing only before display would have left a second problem: most of a CSI sequence is printable text, so an escape puts an alphanumeric immediately before the token, and every rule here treats a preceding alphanumeric as 'not a command'. Colouring a line would have HIDDEN the violation in it, not just recoloured the diagnostic. Sanitizing now happens once at read time — whole CSI sequences stripped, remaining controls mapped to spaces rather than deleted so tokens are not glued together — so matching, section parsing, and the report all see the cleaned line. A fixture asserts both halves: the escaped violation is still caught, and no control character reaches the output.

auerbachb:
Partially fixed in 48845ac, with the rest declined for a documented reason.

Fixed: handoffs now age out (CLAUDE_HANDOFF_MAX_AGE_DAYS, default 7 days). That addresses your third bullet directly — a document from another project weeks ago stayed eligible forever and grew more misleading the longer it sat there. It stays a pure filesystem test (find -mtime), so nothing is read.

Declined, with reasoning: matching against the payload's session_id or cwd.

- session_id cannot work by construction. Running /pause is HOW a session ends, so the session that later dies on a usage limit is never the session that wrote the document. Matching on it would suppress the pointer in essentially every real case.
- cwd would require opening the file to learn which repository it describes. Not reading it is a deliberate property — this is a fire-and-forget StopFailure hook that must never become a new failure mode — and the issue specifies the pointer as 'the most recent portable handoff'.

So the semantics stay 'most recent', and the honesty moves into the hint instead: it now says the document is the newest one on this machine and not necessarily one about this repository, and tells the reader to check that its stated objective matches before acting. That converts a silent wrong-repo assumption into an explicit check, which is the part of your finding that actually bites.

auerbachb:
Fixed in f247711. Step 3 now says to substitute the paths resolved in Step 0 for the collector's literals, and states why: an unresolved path fails silently, so 'could not look' would be rendered as 'no in-flight work' — a distinction the reader has no way to recover. Also added the general rule that a category which could not be READ is not empty and must say so in those words.

auerbachb:
Fixed in f247711. Step 5 now derives --repo-root from the checker's own location and passes it, so the skill-invocation rule always binds to this harness's catalog rather than to whatever project the command happens to be run inside. Subtle one — thanks. The failure mode is quiet in exactly the wrong way: the document passes, and the commands it names are meaningless to the reader.

auerbachb:
Fixed in f247711. You are right that the template wins over the prose — it is the thing that actually gets followed. The Stopped: line now carries both forms explicitly, keyed to WINDDOWN_PERSISTED, with a note that the choice is load-bearing rather than stylistic: printing the persisted form after a failed write misreports the only side effect this command has, and the user would discover it only when a pipeline started anyway. Finishing: got the same treatment for the unresolved-helper case ('could not read what was running', never 'nothing was running').

auerbachb:
@codeant-ai review

auerbachb:
Fixed in 8c4c63d. You identified a conflation I had baked in: 'this is structure, not body content' is a fair reason to keep a heading out of the non-empty-content tally, but it is not a reason to stop checking it. A heading is text the reader sees, and the skip was broad enough that a shell comment inside a fenced block would carry a harness path straight through.

Split now: every rule runs on every line, and only the content tally excludes headings. Four fixtures cover it — a harness path in an H1 and in an H3, a slash command in a comment line, and phase vocabulary in an H2.

auerbachb:
@codeant-ai review

auerbachb:
Fixed in 1d7fc95. Correct and a good catch — whitespace tokenizing was the obvious way to ask 'is this token absolute?' and it quietly assumed no directory name contains a space. Each occurrence is now judged by whether the text before it contains an absolute-path start, so spaces inside the path stop mattering, while repo/.claude/worktrees/... and ./.claude/worktrees/... still have no such start and remain visible to harness-path. Fixture added for the spaced form; the three relative-form fixtures still fail as before.

auerbachb:
Declining this one — the premise does not hold on macOS. BSD find has supported -mindepth, -maxdepth, and -mtime for many years; they are not GNU-only.

Verified directly on the machine this was developed on (Darwin):

$ find "$T" -maxdepth 1 -type f -name 'portable-handoff-*.md' -mtime -7
/var/folders/.../portable-handoff-x-20260801T090000Z.md # rc=0
$ find "$T" -mindepth 1 -maxdepth 1 -type d
/var/folders/.../sub # rc=0

Both return rc=0 with correct output. The stronger evidence is the test suite itself: this whole change was developed and run on macOS, and the age-bound case (15g in usage-limit-record.test.sh) exercises exactly `find -maxdepth ... -mtime` — it passes here, and it would fail loudly if find rejected those flags, because that test asserts a fresh handoff IS still found after an old one ages out. Same for the catalog lookup: if -mindepth/-maxdepth failed, CATALOG_RESOLVED would be 0 and the fallback regex would flag the /tmp and /usr/local/bin lines in the no-false-positives fixture, which passes.

Happy to revisit with a specific platform and find version where these flags fail.

auerbachb:
Declining this one — the premise does not hold on macOS. BSD find has supported -mindepth, -maxdepth, and -mtime for many years; they are not GNU-only.

Verified directly on the machine this was developed on (Darwin):

$ find "$T" -maxdepth 1 -type f -name 'portable-handoff-*.md' -mtime -7
/var/folders/.../portable-handoff-x-20260801T090000Z.md # rc=0
$ find "$T" -mindepth 1 -maxdepth 1 -type d
/var/folders/.../sub # rc=0

Both return rc=0 with correct output. The stronger evidence is the test suite itself: this whole change was developed and run on macOS, and the age-bound case (15g in usage-limit-record.test.sh) exercises exactly `find -maxdepth ... -mtime` — it passes here, and it would fail loudly if find rejected those flags, because that test asserts a fresh handoff IS still found after an old one ages out. Same for the catalog lookup: if -mindepth/-maxdepth failed, CATALOG_RESOLVED would be 0 and the fallback regex would flag the /tmp and /usr/local/bin lines in the no-false-positives fixture, which passes.

Happy to revisit with a specific platform and find version where these flags fail.

auerbachb:
@codeant-ai review

auerbachb:
Fixed in 178c76a — this was a genuine time bomb and a good catch. Pinning mtimes to a fixed calendar date while the hook filters with a relative `find -mtime -7` means the fixtures pass on the day they are written and start failing about a week later, for a reason having nothing to do with the code.

All fixture times now come from stamp_ago()/touch_ago() helpers built on the current clock (BSD `date -r`, GNU `date -d @`): the stale case is a deliberate 30 days back, the fresh case is minutes.

Your finding also surfaced a second, quieter problem I had missed. In the mtime-ordering case the newer file was ALSO the lexically greater name, so the name-based tie-break would have satisfied the assertion even if mtime ordering were broken entirely — the test could not fail for the thing it existed to check. The newer file now carries the lexically SMALLER name, so only correct mtime ordering passes it.

auerbachb:
@codeant-ai review

auerbachb:
@codeant-ai review

auerbachb:
@cursor review

auerbachb:
Fixed in 95ee728 — and this was exactly inverted from the bug the flag existed to fix. WINDDOWN_PERSISTED was only ever assigned on the failure path, so on success it stayed unset and Step 2, which tests for 1, would have reported every successful pause as unrecorded. Initialized to 1 before the attempt, with a note in the skill that the initialiser is load-bearing rather than decoration.

auerbachb:
Fixed in 95ee728. Validated with a fallback to 7. The silent part is what makes it worth fixing: find's stderr is discarded, so an inherited bad value would have cost the user the pointer with no trace at all — in the one session that has the least capacity to notice. Fixtures cover non-numeric, empty, negative, zero, and a command-injection-shaped value, each asserting the fresh handoff is still found.

auerbachb:
Fixed in 95ee728. Agreed on the distinction: a URL is an address any reader can open, so a repository link is a portable reference, while a local path only resolves inside this checkout — and only the latter is what harness-path exists to catch. URLs are now exempt, with a fixture linking to safety.md on GitHub.

Worth noting for anyone reading this later: the URL exemption tokenizes on whitespace, which is the exact approach that was WRONG for the worktree-path exemption you flagged earlier. It is sound here for a specific reason — URLs cannot contain unescaped spaces, whereas directory names can.

auerbachb:
@codeant-ai review

auerbachb:
Fixed in 28f1c20 — a real hole, and in the heuristic I had just widened. The check now requires the run of text immediately before the marker to begin at a / following line-start or whitespace, with no prose punctuation in between. Spaces stay legal inside that run (that is the whole reason this is not a whitespace-token test), but a comma, semicolon, or colon ends it. Two fixtures added for the mixed-line forms, including your exact example.

auerbachb:
Bounded in 28f1c20: one non-recursive directory read, newest names first, capped at 50 entries. Filenames sort chronologically, so the newest by mtime is inside that window in any realistic directory; verified against 120 files.

One clarification on the risk rather than the fix, since the framing matters for anyone reading this later. This adds a single readdir to a filesystem the hook already writes to several times in the same run — mkdir, chmod, the append itself, and an atomic rename. A home slow enough to consume the 5s budget on one directory read was already consuming it on those, so this was not a new class of exposure. The cap is still worth having: it removes the unbounded shape entirely, which is cheaper than reasoning about when it would bite.

auerbachb:
@codeant-ai review

auerbachb:
Fixed in 3b29d1b, and you were right to re-report it — my punctuation guard only handled the form with the comma, and your comma-less example walked straight through it.

The underlying problem is that the question is unanswerable in free prose. In 'See /tmp/build then repo/.claude/worktrees/x' nothing distinguishes 'then repo/...' continuing the /tmp path from separate words, so ANY space-tolerant rule there lets the unrelated path vouch for the relative one. Tightening the heuristic was never going to close it.

So the rule now depends on where the grammar answers the question. The Working directory: field's entire value is one path by definition, so a space there is unambiguous and is honoured — that is the real case the exemption exists for. Everywhere else a path is a whitespace-delimited token, full stop. The spaced fixture moved into the field, the free-prose spaced form is now expected to FAIL, and the template documents where a spaced path belongs.

One more thing came out of this. While restructuring, a bad edit of mine deleted the URL-masking function; every line's scan buffer went empty and the checker cheerfully passed a document it had checked none of. The suite caught it, but a checker that can report 'clean' when its own pipeline is broken is precisely the failure this script exists to prevent — so a non-empty line producing an empty scan is now exit 4, refusing to report a result rather than reporting a false one.

auerbachb:
@codeant-ai review

auerbachb:
@cursor review

auerbachb:
Fixed in b8c956a — and this caught an error in my reasoning, not just a line of code.

I justified the cap with 'filenames sort chronologically'. They did not: the name was portable-handoff-{SESSION_ID}-{TIMESTAMP}.md, so lexical order sorted by SESSION. Exactly as you describe, with more than 50 eligible handoffs the genuinely newest document could fall outside the window and the pointer would name an older one.

Rather than drop the cap, I fixed the premise. The timestamp now leads — portable-handoff-{TIMESTAMP}-{SESSION_ID}.md — which makes lexical order actually chronological. The cap becomes sound, a directory listing reads in the order the sessions happened, and the justifying comment is now true rather than merely confident. The hook's comment names the coupling explicitly so a future rename that puts the session id back in front has to confront it.

auerbachb:
@codeant-ai review

auerbachb:
@cursor review

auerbachb:
Fixed in 4874903, and the right answer turned out to be the one you suggested two rounds ago rather than the cap I reached for instead.

You are right that the cap does not make the lookup bounded — find still enumerates the directory and sort still materializes the set. And no cap can, because finding the newest of N files is inherently O(N). The ordering is what actually removes the risk.

Phase 1 now builds and writes the record with no pointer, exactly as this hook behaved before the feature existed. Phase 2 releases the lock, does the lookup, and refines usage-limit-last.json alone. Killed anywhere in phase 2, the user still has the record — which is the one thing this hook exists to produce.

Two deliberate details: only `last` is refined, because the events log is a history of what happened while `last` is the pointer a resuming session reads first, and rewriting history to add a field the reader gets from `last` anyway is not worth a second append. And the lock is dropped before phase 2, since every session hitting the same limit resolves the same newest handoff, so last-writer-wins on an atomic rename is correct and none of them should wait on another's enrichment.

A structural test asserts the ordering by line number, so moving the lookup back in front of the append fails the suite rather than silently reintroducing this.

auerbachb:
@codeant-ai review

auerbachb:
@cursor review

auerbachb:
Fixed in 808d929 — you and cursor[bot] found this independently, which is a fair signal it was real. It was a regression I introduced with the phase split, and the concurrency case is the expected one here rather than an edge case, since one account limit fails every active session at once.

Phase 2 now re-takes the lock and replaces only its OWN record: if usage-limit-last.json no longer holds the exact bytes this invocation wrote, a newer one owns the file and this one skips — that newer invocation runs its own phase 2, so nothing is lost by yielding. The lock is still dropped between the phases, which is what buys the concurrency the split was for; it just no longer leaves the publish unserialized.

A regression test runs eight concurrent invocations with a handoff present and asserts both that no record is lost and that last agrees with the newest event.

auerbachb:
Fixed in 808d929 — you and cursor[bot] found this independently, which is a fair signal it was real. It was a regression I introduced with the phase split, and the concurrency case is the expected one here rather than an edge case, since one account limit fails every active session at once.

Phase 2 now re-takes the lock and replaces only its OWN record: if usage-limit-last.json no longer holds the exact bytes this invocation wrote, a newer one owns the file and this one skips — that newer invocation runs its own phase 2, so nothing is lost by yielding. The lock is still dropped between the phases, which is what buys the concurrency the split was for; it just no longer leaves the publish unserialized.

A regression test runs eight concurrent invocations with a handoff present and asserts both that no record is lost and that last agrees with the newest event.

auerbachb:
Fixed in 808d929. The check aggregated content by heading NAME, so a populated second copy vouched for an empty first one — while the reader hits the empty one first, which is the copy that matters. Duplicated required headings are now rejected outright rather than guessing which copy counts, with a fixture for the empty-then-populated shape.

auerbachb:
@codeant-ai review

auerbachb:
@cursor review

auerbachb:
Fixed in 0fcd88a. You are right that I left the hole half-closed: take_lock's result was ignored, so on contention phase 2 ran unserialized and another invocation could replace the file between the guard read and the rename. Without the lock the guard is not a guard.

Phase 2 now exits when the lock cannot be taken. That is the correct asymmetry here — enrichment is optional, `last` being correct is not, and the invocation that won the lock publishes its own pointer anyway, so nothing is lost by yielding.

auerbachb:
Fixed in 0fcd88a. Agreed — a capital letter cannot be what decides whether the jargon ships, and 'phase b' carries the identical internal meaning.

Now case-insensitive, including the Merge_Ready and phase-A-fixer variants. I had originally kept it case-sensitive to avoid a false positive on prose like 'in a later phase a decision was made', and that risk is real — but it is the cheap side of the trade: a false positive costs the author a reword, while a false negative costs the reader their ability to act on the document at all. Fixtures added for the lowercase, mixed-case, and Merge_Ready forms, and the existing no-false-positive fixture ('phased rollout', 'phases') still passes.

auerbachb:
Fixed in 0fcd88a — and it is the same failure the collector text warns about, which makes leaving it in place worse than the original bug. Step 0 now resolves the collector document alongside the helper scripts (testing readability rather than the executable bit, since it is a document), and Step 3 reads it by that resolved path. If it cannot be found at all, the skill says so in its report and gathers what it can directly, rather than producing a document that reports nothing in flight because it could not look.

auerbachb:
@codeant-ai review

auerbachb:
@cursor review

auerbachb:
@coderabbitai Fixed in ebb2196. You are right and it was worse than having no test — an assertion ending in `|| true` reads as coverage while verifying nothing. Every line of the events log is now checked to carry the base hint and a null pointer, which is what the phase split actually promises.

auerbachb:
@coderabbitai Fixed in ebb2196. The code contradicted a rule stated two lines below it, which is the worse half of the problem — a reader following the prose would assume the guard existed. `mv` is now tested, and on failure the step names the verified draft path instead of a destination that was never created.

auerbachb:
@coderabbitai Fixed in ebb2196 — the line was stale from an earlier draft where the collector invoked the helper. Reworded to say what is actually true: `session-state.sh` is invoked, per-PR handoff payloads are read directly as files in section 2, and a consumer reaching for `handoff-state.sh` should resolve it the same way.

auerbachb:
@coderabbitai Fixed in ebb2196. Exit 4 is now in both the EXIT STATUS header and the --help text, with the reason attached — it is the fail-closed path for a line whose scan buffer came back empty, meaning the rules could not have run against it.

auerbachb:
@coderabbitai Fixed in ebb2196. The test now asserts the EXIT STATUS section, the --repo-root flag, and the newly documented exit code, so a truncated or broken help text fails rather than passing on its exit status alone.

auerbachb:
Declining this one, with reasoning — and the asymmetry you have spotted between phase 1 and phase 2 is deliberate rather than an oversight.

Phase 1 proceeding without the lock is the hook's original, pre-existing design, and its rationale is in the comment above the loop: 'if the lock cannot be taken we proceed anyway, because losing the record is worse than a rare interleave.' The lock is bounded at ~1s, and the alternative on contention is to write nothing — in the exact scenario where every active session just died at once and the record is the only thing the user will have. An interleaved line in an append-only log is recoverable; a missing record is not. Rotation, which genuinely can destroy history, is already skipped on the unlocked path for precisely this reason.

Phase 2 is the opposite trade, which is why it aborts instead: the enrichment is optional, `last` being correct is not, and the invocation holding the lock publishes its own pointer regardless. Nothing is lost by yielding there.

So both paths resolve the same question — 'what does the user lose if this step does not happen?' — and reach opposite answers because the answers genuinely differ. Changing phase 1 to abort would trade a rare interleave for a missing record, which is the wrong direction for this hook. Happy to revisit if you see a case where an unlocked append loses data rather than interleaving it.

auerbachb:
@codeant-ai review

auerbachb:
Fixed in 0bc1564. Correct — and the template itself uses bold in the Open work section, so this was a false positive waiting to happen on a document that was entirely fine, which is the failure mode that gets a checker bypassed rather than fixed.

Both masking paths now strip every leading delimiter (backtick, brackets, quotes, asterisk, underscore, tilde) rather than a single character. Fixtures cover a bolded absolute worktree path and a bolded repository URL.

auerbachb:
@codeant-ai review

auerbachb:
Fixed in 3e1fabb. Right — the leading run in a Markdown link is link TEXT, not delimiters, so a one-character-at-a-time strip never reaches the destination no matter how many characters it removes. The scan now starts after "](". Fixture added for a linked absolute worktree path.

auerbachb:
Fixed in 3e1fabb. Agreed — a section holding only a fence pair or an HTML comment is the empty-shell failure wearing a disguise: it satisfies a non-whitespace count while answering nothing. Structural-only lines no longer count toward the content tally, with a fixture for a fence-only section.

auerbachb:
@codeant-ai review

auerbachb:
Fixed in 9f57088, by deleting the cap rather than defending the window.

You are right that the discrepancy remained: timestamp-first names make the newest-by-mtime file *unlikely* to fall outside the lexically greatest 50, not impossible — an mtime perturbed by a copy or a restore parts company with the name.

The cap only ever existed to keep a slow lookup from consuming the 5s budget before the durable append ran, and the phase split removed that risk at its root: the record is on disk before the lookup happens at all. So it was guarding against something that can no longer occur, at the cost of the contract it was documented to uphold. Without it, newest-by-mtime is exactly true.

A test now creates 120 handoffs and gives the newest one the lexically SMALLEST name, so any name-ordered truncation fails it rather than passing by luck.

auerbachb:
@codeant-ai review

auerbachb:
@codeant-ai review

auerbachb:
Fixed in a7f61e8 — and this one is not hypothetical for this skill specifically. Step 6 prints the finished document inside a fence, so a render that captured its own fence would put every heading in a code block and satisfy all the structural rules while the reader sees one undifferentiated blob.

Heading detection and the working-directory field now ignore fenced lines. Content inside a fence still counts as content, deliberately: a command block is real material for the reader, and what this closes is 'all the STRUCTURE is fenced', not 'a fence exists'. The fixture wraps the golden document in a fence and expects every required section to be reported missing.

auerbachb:
Re-declining, same reasoning as the earlier thread on this exact code path — it is the hook's original, pre-existing design rather than something this PR introduced.

The comment above the lock loop states the trade: 'if the lock cannot be taken we proceed anyway, because losing the record is worse than a rare interleave.' The wait is bounded at ~1s, and the alternative on contention is to write nothing at all — in precisely the scenario where every active session just died at once and this record is the only thing the user will have. An interleaved line in an append-only JSONL log is recoverable by a reader; a missing record is not recoverable by anyone.

The specific risks you list are already handled asymmetrically for that reason. Rotation — the one operation that can genuinely destroy history — is skipped entirely on the unlocked path, which is why an unlocked run cannot 'append across a rotation'. And phase 2, where the trade runs the other way, now aborts outright without the lock: the enrichment is optional, so yielding costs nothing.

So both paths answer the same question — what does the user lose if this step does not happen — and land differently because the answers genuinely differ. I would change phase 1 if you can point to a case where an unlocked append loses data rather than interleaving it; short of that, making it abort would trade a rare, readable interleave for an absent record, which is the wrong direction for a hook whose entire purpose is leaving something behind.

auerbachb:
@codeant-ai review

auerbachb:
Declining, because this and the ordering are mutually exclusive and the current order is the safer one.

Publishing the enriched record as a single atomic operation requires knowing the handoff path before the write, which puts the directory scan back in front of the durable append — the exact Major issue raised twice on this file (that a slow lookup can consume the 5s budget and cost the user the record entirely). Fixing this finding reintroduces that one. There is no ordering that avoids both.

Given the choice, the intermediate state is the cheaper cost, and by some distance. A reader that catches usage-limit-last.json between the phases sees a complete, valid record whose portable_handoff is null — which is precisely the behaviour this hook had before #901, and is still fully useful: session id, cwd, transcript path, and the resume hint that shipped for two releases. The alternative failure is no record at all, in the one moment the user most needs one.

On the second half of your suggestion — making the intermediate state explicitly distinguishable — that is already true rather than something to add. The base record carries portable_handoff: null and the unmodified base resume_hint, so a reader can tell a not-yet-enriched record from an enriched one by inspection. The only thing it cannot distinguish is 'not yet enriched' from 'no handoff exists', and those call for the same action from the reader: proceed from the transcript. Adding a marker to separate them would encode a distinction nobody acts on.

Happy to revisit if there is an ordering that keeps the append first AND publishes once — I could not find one.

PR #913, opened 2026-08-01
fix(#874): scope session-scheduling-reconcile purge to verifiably-dead sessions

### LINKED ISSUE #874: session-scheduling-reconcile: purge assumes one Claude session per HOME
Surfaced by CodeAnt on PR #867 (issue #827).

`session-scheduling-reconcile.sh` purges when `SessionStart` reports `source=startup`. That proves *this* session is new; it does not prove no **other** Claude session is running against the same `~/.claude/session-state.json`. Two concurrent sessions therefore let a newly-started one clear bookkeeping owned by a live one.

Current exposure is small, and deliberately so:
- **Watchers** are already protected by the freshness window (`max(3x cadence, 30m)`) — a live watcher in another session has a recent `last_tick_at` and is not reaped.
- **`polling_jobs[]` / `auto_wake_cron_id`** are the unprotected fields, but since #827 no skill writes them, so in practice they are empty.

A real fix needs session-ownership metadata (a session id or generation stamped on each record) so reconciliation can distinguish "left by a dead session" from "owned by a live one". That is a broader change to the state contract than #827 should carry.

_Filed via /wrap._

---

## Implementation Plan

**Scope (narrow — per parent task):** Touch only `.claude/scripts/session-scheduling-reconcile.sh` and its test file. No corpus changes, no new skills, no stamp writes to other skills.

### Root cause
`session-scheduling-reconcile.sh` purges `.polling_jobs[]` and `.pmm.auto_wake_cron_id` unconditionally. A sibling session that wrote those records may still be live, so the purge corrupts its bookkeeping.

### Fix: per-record liveness check (mirrors `state-lock.sh`)

Each record optionally carries an `owner_session: {pid, host}` stamp (written by the skill at creation time — that write path is out of scope here; for now the stamp is optional). The reconcile script:

1. **Bash precompute** — for each `.polling_jobs[]` entry and for `.pmm.auto_wake_cron_id`, run `kill -0 <pid>` (same dead-holder test as `state-lock.sh`):
   - `dead` — stamp present, same host, `kill -0` fails
   - `alive` — stamp present, same host, `kill -0` succeeds
   - `unknown` — no stamp, cross-host, or malformed pid

2. **jq PLAN** — filter by verdict:
   - `dead` → purge (existing behavior, now guarded)
   - `alive` or `unknown` → skip purge, surface a notice (**fail-closed**)

3. **Babysit watchers** — no change; the freshness window (`max(3× cadence, 30m)`) already protects live watchers in sibling sessions.

4. **Writes** — all through `session-state.sh --raw-path` (no raw jq+mv); same as before.

### Files changed
- `.claude/scripts/session-scheduling-reconcile.sh` — add `session_is_dead()` helper, bash precompute, updated jq PLAN with `--argjson pj_verdicts` / `--arg aw_verdict`
- `.claude/scripts/tests/session-scheduling-reconcile.test.sh` — add tests 1a (unstamped not purged), 1b (live PID not purged); update test 1 to use dead-stamped records; update test 10 to check `polling_jobs_skip`

*CR plan (Tasks 1–2) deferred: stamping at write time and the shared session-owner primitive are a broader contract change for a follow-up issue.*

auerbachb:
## Summary

- Adds `session_is_dead()` helper (mirrors `state-lock.sh`'s `kill -0` dead-holder test) to determine per-record liveness before purging scheduling bookkeeping
- Precomputes verdict (`dead` / `alive` / `unknown`) for each `.polling_jobs[]` entry and for `.pmm.auto_wake_cron_id` in bash, then passes the verdicts into the jq PLAN via `--argjson`
- **Fail-closed:** records with no `owner_session` stamp, a cross-host stamp, or a live-PID stamp are skipped — never purged — and a `scheduling-reconcile:` notice is surfaced
- Babysit freshness-window path unchanged; live watchers in sibling sessions are already protected there

Closes #874

## Test plan

- [x] Test 1: dead-owner-stamped `polling_jobs` and `auto_wake_cron_id` are still purged (existing behavior preserved)
- [x] Test 1a: unstamped `polling_jobs` / `auto_wake_cron_id` are NOT purged (fail-closed regression guard)
- [x] Test 1b: live-PID-stamped records are NOT purged
- [x] Test 0: no apostrophe inside the single-quoted jq program
- [x] All 33 existing + new tests pass (`bash .claude/scripts/tests/session-scheduling-reconcile.test.sh`)
- [x] Both CLIs unavailable (CR CLI absent; CodeAnt 403 after one retry — daily cap, not auth); self-review performed; no issues found

[COVERAGE: none — CR CLI absent; CodeAnt 403 (daily cap); self-review only]

### DIFF
--- a/.claude/hooks/tests/session-start-sync-reconcile.test.sh
+++ b/.claude/hooks/tests/session-start-sync-reconcile.test.sh
@@ -16,12 +16,27 @@ trap cleanup EXIT
 fail() { echo "FAIL: $*" >&2; exit 1; }
 ok() { echo "ok   — $*"; }
 
+# Build state fixtures.  The reconcile script is fail-closed: only records with
+# a verifiable dead-owner stamp are purged.  The STAMPED variant carries a dead
+# PID so startup can provably purge it; UNSTAMPED has no stamp and must survive
+# startup unchanged (fail-closed guarantee).
+( : ) &
+_DEAD_PID=$!
+wait "$_DEAD_PID" 2>/dev/null || true
+_THIS_HOST="${HOSTNAME:-$(hostname 2>/dev/null || echo "localhost")}"
+STAMPED_STATE="$(printf \
+  '{"polling_jobs":[{"id":"live_job","owner_session":{"pid":%s,"host":"%s"}}],"pmm":{"auto_wake_cron_id":"live_cron","auto_wake_owner_session":{"pid":%s,"host":"%s"}}}' \
+  "$_DEAD_PID" "$_THIS_HOST" "$_DEAD_PID" "$_THIS_HOST")"
+UNSTAMPED_STATE='{"polling_jobs":[{"id":"live_job"}],"pmm":{"auto_wake_cron_id":"live_cron"}}'
+
 CASE_N=0
 OUT=""
 STATE=""
-# run_hook <source-json> — fresh HOME seeded with purgeable state.
+# run_hook <source-json> [initial-state] — fresh HOME seeded with scheduling state.
 # Sets $STATE and $OUT in the CALLER's scope: a command substitution would run
 # this in a subshell and the $STATE assignment would never escape it.
+# The optional second argument overrides the initial state written to $STATE;
+# defaults to UNSTAMPED_STATE when omitted.
 run_hook() {
   CASE_N=$((CASE_N + 1))
   export HOME="$TMP_DIR/home$CASE_N"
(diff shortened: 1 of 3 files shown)


auerbachb:
Fixed in d937bb5: changed CURRENT_HOST to use ${HOSTNAME:-$(hostname 2>/dev/null || echo "")} to match the convention used in state-lock.sh, so stamps created by processes that prefer the HOSTNAME env var are matched correctly.

auerbachb:
Fixed in d937bb5: added PJ_SNAPSHOT variable that captures .polling_jobs once from the state file in bash. The snapshot is passed to the jq plan via --argjson pj_snapshot so both the verdict precompute (the while loop) and the jq plan operate on the same frozen copy. This eliminates the race where a concurrent --set could shift the array between reads, causing verdict index misalignment.

auerbachb:
@cursor review

auerbachb:
Fixed in b4ec6e1: replaced the snapshot kept-list with dead fingerprints (pid+host pairs). The jq plan now filters the live .polling_jobs array using these fingerprints, so jobs added by sibling sessions after the bash snapshot was taken are preserved.

auerbachb:
@cursor review

PR #937, opened 2026-08-02
fix(#930): make state-lock give real mutual exclusion on Linux

### LINKED ISSUE #930: state-lock.sh loses updates on Linux: two writers enter the critical section, both exit 0
## Summary

`state-lock.sh` does not reliably provide mutual exclusion on Linux. Under contention, **two or more writers enter the critical section at once**, each completes its read-modify-write, and **each exits 0** — while one writer's change is silently discarded.

This surfaced as an intermittent CI failure in `.claude/scripts/tests/handoff-state.test.sh` (the "20 concurrent `--append` writers" block, `FAIL — all 20 unique filenames present (expected '20', got '19')`) on PRs with markdown-only diffs. It is **not** a flaky test. The test is correctly asserting an invariant the lock fails to uphold.

**Blast radius is not limited to tests.** The same library serializes every writer of `~/.claude/session-state.json` and the per-PR handoff files: `session-state.sh`, `handoff-state.sh`, `polling-state-gate.sh`, `dismiss-stale-bot-changes.sh`, `greptile-budget.sh`, `cr-review-hourly.sh`, `session-scheduling-reconcile.sh`, `session-state-audit.sh`. Concurrent orchestrators (`/babysit-pr`, `/pr-monitor-and-manage`, parallel Phase A/B subagents) can silently lose watermark, reviewer, and findings writes — the exact class of bug #639 and #682 were opened to prevent.

## Evidence

Reproduced in a Linux container (`ubuntu:latest`, `--cpus=1`) against the scripts as they exist on `main`:

- **Lost updates:** 12/12 rounds of the 20-writer append race lost at least one append (observed lengths 12–19 of 20), with **zero non-zero exit codes**.
- **Direct mutual-exclusion probe** (20 processes take the lock, log IN/OUT): **up to 15 overlapping entries per round**, zero acquisition failures.
- **Ownership-token probe** (each holder drops a unique token file inside the lock dir): two holders' tokens observed coexisting in one lock directory, each holder reporting its own token present — i.e. both genuinely believed they held the lock.
- **Not a container artifact:** violations reproduce on `tmpfs` (18 over 6 rounds) and on a bind mount (59 over 6 rounds), not just overlayfs. GitHub Actions `ubuntu-latest` runs on ext4, so the CI failures are the same defect.
- Passes locally on macOS because root cause A below is **GNU-coreutils-specific**.

## Root causes

**A. `_state_lock_age()` uses BSD `stat` syntax; on GNU coreutils it always reports 999999.**

```bash
mtime="$(stat -f %m "$lock_dir" 2>/dev/null || stat -c %Y "$lock_dir" 2>/dev/null || true)"
```

On GNU coreutils `-f` is `--file-system` (a boolean), so `%m` is parsed as a *filename*. The command fails on `%m` but still prints a multi-line filesystem dump to stdout and exits 1, so the `||` fallback runs and *appends* the real mtime. The result is a multi-line, non-numeric string, so the `^[0-9]+$` test fails and the function returns the `999999` sentinel. Verified:

```
mtime=[  File: "/tmp/lk"
    ID: 570889ec493abae5 Namelen: 255     Type: overlayfs
...
1785634740]
NON-NUMERIC -> age becomes 999999 (ALWAYS STALE)
```

Consequence: on Linux, **any lock whose `owner` file is momentarily unreadable is instantly classified stale**, regardless of its true age.

**B. Two routine windows per lock cycle leave `owner` unreadable — so A fires constantly.**

1. *Acquire:* `mkdir` wins, then the owner file is written by five separate `printf`s into one redirect. It is not atomic, so readers observe the file absent or partially written (no `epoch=` line yet).
2. *Release:* `rm -rf "$lock_dir"` unlinks `owner` **before** it removes the directory, so the lock exists with no owner for the duration.

**C. The double-steal guard is a 0.2 s sleep, which CPU contention defeats.**

`_state_lock_break()`'s only protection against breaking a *live* holder is sampling the owner bytes, sleeping 0.2 s, and re-sampling. With 20 writers on a 1–2 vCPU runner, a holder is routinely descheduled longer than 0.2 s inside window B, so `before == after == "__no_owner__"`, the lock is still "stale" per A, and the break proceeds.

**D. A robbed holder never finds out, and commits anyway.**

The break renames the directory aside, but the victim refers to the lock only *by path* and holds an already-open fd for its owner write. It therefore does not error, keeps `STATE_LOCK_DIR` set, and proceeds through its read-modify-write **while the breaker acquires a fresh lock on the same path**. Both write, both exit 0, one update is lost. At release the victim reads the new holder's `owner`, sees a pid mismatch, and silently declines to delete — which is why the failure leaves no trace in exit codes.

**E. Minor, latent:** in `_state_lock_meta`, `done < "$file" 2>/dev/null` applies redirections left-to-right, so a failed input redirect prints to the real stderr despite the intended suppression. Observed leaking as `state-lock.sh: line 141: .../owner: No such file or directory`.

## Acceptance criteria

- [ ] `_state_lock_age` computes a correct age on both GNU and BSD `stat`, and an indeterminate age fails **safe** (not-stale) rather than declaring the lock stale.
- [ ] The owner metadata becomes visible atomically — no reader ever observes a partially-written `owner`.
- [ ] Release removes the lock without leaving an owner-less window, and only when the caller still owns it (verified by a unique token, not a bare pid).
- [ ] A holder whose lock was broken **fails closed**: it must not commit its read-modify-write. Breaking a stale lock can never be perfectly safe, so the victim must detect the theft and return a non-zero, retryable status.
- [ ] The 20-writer concurrency test passes repeatedly under CPU contention on Linux; the assertion stays `== 20` and is **not** loosened.
- [ ] Deterministic (non-probabilistic) regression tests cover each defect above, and fail against the pre-fix library.
- [ ] `session-state.sh`'s existing test suite still passes (shared library).

## Repro

```bash
docker run --rm --cpus=1 -v "$PWD:/repo" ubuntu:latest bash -c '
  apt-get update -qq && apt-get install -y -qq jq git >/dev/null
  # 20 concurrent --append writers against one handoff file; expect length 20
  ...'
```

Full repro harness (append race, mutual-exclusion probe, token probe, filesystem control) is attached to the PR.

---

## Implementation Plan

*Source: Claude's analysis. CodeRabbit plan unavailable — the auto-enrichment comment on this issue contains only the "Create Plan" boilerplate, no substantive plan (`cr-plan.sh` exit 1).*

### `.claude/scripts/state-lock.sh`

1. **Per-acquisition token.** Generate `STATE_LOCK_TOKEN` (pid + epoch + random) at acquire and record it as `token=` in the owner file. Ownership decisions key off the token, not the pid — a pid alone cannot distinguish this acquisition from a later one.
2. **Atomic owner publication.** Write the metadata to `"$lock_dir/.owner.$$"` and `mv` it onto `"$lock_dir/owner"`. Same-directory `rename()` is atomic, so `owner` is never observed partially written (root cause B1).
3. **Correct, validated age.** Try GNU `stat -c %Y` then BSD `stat -f %m`, accepting only numeric output (root cause A).
4. **Indeterminate age fails safe.** A missing/unparseable `owner` no longer implies staleness. Age falls back to the lock directory's mtime; if no age can be established at all, the lock is treated as **not stale**. A genuinely wedged lock is still bounded — by the acquire timeout, and by the directory-mtime rule once it exceeds `STALE_AGE` (root causes A+B).
5. **Atomic release.** Replace `rm -rf "$lock_dir"` with `mv "$lock_dir" "$lock_dir.rel.<token>"` followed by `rm -rf` on the aside path. The lock path goes from fully-present to absent in one atomic step, closing the owner-less release window (root cause B2). Release still requires a positive ownership match — now on the token.
6. **Fail-closed steal detection.** New `state_lock_assert_held`: returns non-zero when `owner`'s token no longer matches `STATE_LOCK_TOKEN`. Breaking a stale lock can never be perfectly safe, so the victim must be able to detect the theft rather than commit blindly (root cause D).
7. **Break guard keyed on the token** and re-verified with the corrected age function after the resample.
8. Fix `_state_lock_meta`'s redirect ordering so a failed input redirect cannot leak to the real stderr (root cause E).

### Callers — fail closed at the commit point

- `.claude/scripts/handoff-state.sh` — call `state_lock_assert_held` in `_atomic_write` immediately before the `mv`; on failure discard the temp file and exit `6` (retryable, unchanged file) instead of committing.
- `.claude/scripts/session-state.sh` — same guard before its `mv "$OUT_TMP" "$STATE_FILE"`.

### Tests

Deterministic regression tests (each fails against the pre-fix library, none timing-dependent):

- `_state_lock_age` returns a correct small age for a lock directory with no owner file — the direct GNU-`stat` regression.
- A young lock whose owner file is absent is **not** classified stale.
- `owner` is never observable in a partial state.
- Release leaves no owner-less intermediate state.
- **Steal detection:** a holder whose lock directory is renamed away and re-acquired by another process detects it, and `handoff-state.sh --append` exits `6` in that state without clobbering the file.

The existing 20-writer assertion stays exactly `== 20` — not loosened — and is additionally verified under sustained CPU contention on Linux (containerised `ubuntu:latest --cpus=1`, many rounds) rather than only on macOS. `state-lock.test.sh` and `session-state.sh`'s suites must continue to pass, since the library is shared by nine production writers.

auerbachb:
## What this fixes

`state-lock.sh` did not reliably provide mutual exclusion on Linux. Under contention, two or more writers entered the critical section at once, each completed its read-modify-write, and **each exited 0** while one update was silently discarded.

The reported symptom was an intermittent `handoff-state.test.sh` failure on unrelated PRs (`FAIL — all 20 unique filenames present (expected '20', got '19')`). **This is a real bug, not a flaky test** — the assertion was correctly reporting an invariant the lock failed to uphold, which is why it also fails on markdown-only diffs and passes on a plain re-run.

The blast radius is wider than the test: the same library serializes `session-state.json` and the per-PR handoff files for nine production writers (`session-state.sh`, `handoff-state.sh`, `polling-state-gate.sh`, `dismiss-stale-bot-changes.sh`, `greptile-budget.sh`, `cr-review-hourly.sh`, `session-scheduling-reconcile.sh`, `session-state-audit.sh`). Concurrent orchestrators could silently lose watermark, reviewer, and findings writes — the exact failure #639 and #682 were opened to prevent.

## Root causes

**A — `_state_lock_age` used BSD `stat` syntax, and the wrong spelling does not fail cleanly.** On GNU coreutils `-f` is `--file-system` and `%m` is parsed as a *file operand*: the command fails on `%m` but still prints a multi-line filesystem dump to stdout, so the `stat -f %m … || stat -c %Y …` chain **concatenated** the dump with the real mtime. The result never matched `^[0-9]+$`, so the function returned its `999999` "ancient" sentinel. On Linux, **any lock whose `owner` file was momentarily unreadable was instantly classified stale.** macOS took the working branch — hence local-pass / CI-fail.

**B — two routine windows per lock cycle leave `owner` unreadable,** so A fired constantly: acquire wrote the owner file with five separate `printf`s (observable half-written), and release's `rm -rf` unlinked `owner` *before* removing the directory.

**C — the double-steal guard was a 0.2 s resample,** which CPU contention defeats: `before == after == "__no_owner__"` and the break proceeds against a live holder.

**D — a robbed holder never found out.** It referred to the lock only by path, so it did not error, kept `STATE_LOCK_DIR` set, and committed anyway — both writers exit 0, one update vanishes.

**E — minor/latent:** `done < "$file" 2>/dev/null` applies redirections left-to-right, so a failed input redirect printed to the real stderr despite the intended suppression.

## Approach

- **Portable, validated age lookup** — try `stat -c %Y` then `stat -f %m`, accept only numeric output.
- **Indeterminate age fails safe.** Both staleness rules now require *positive* evidence; absent evidence means "not stale". A lock orphaned before publishing `owner` still recovers via the directory's mtime once it passes `STALE_AGE`, so nothing wedges the fleet.
- **Atomic owner publication** — build under a temp name inside the lock dir, then `rename` into place. `owner` is never observable partially written.
- **Per-acquisition token.** A pid cannot distinguish this acquisition from a later one, so release and the new assert key off `token=`.
- **Atomic teardown** — `rename` the directory aside, *then* delete, so the lock is never visible-but-ownerless.
- **`state_lock_assert_held` (fail closed).** Breaking a stale lock can never be perfectly safe, so the victim must be able to detect the theft. `handoff-state.sh` and `session-state.sh` call it immediately before their commit and redo the read-modify-write from a fresh read (bounded retry, then exit 6) rather than committing a stale snapshot.

## Verification

Linux container (`ubuntu:latest`, `--cpus=1`) reproducing CI's contention:

| | before | after |
|---|---|---|
| 20-writer append race, rounds losing an append | **12 / 12** | **0 / 12** |
| Observed simultaneous lock holders (direct probe) | up to **15** | — |
| Non-zero exits while updates were lost | **0** (silent) | n/a |

Not a container artifact: violations also reproduced on `tmpfs` and a bind mount, and GitHub Actions `ubuntu-latest` runs on ext4.

Full auto-discovered suite on Linux — `origin/main` fails **5** suites *including both* `handoff-state.test.sh` and `state-lock.test.sh`; this branch fails **3**. The remaining 3 (`babysit-tick-watchdog`, `merge-gate-review-substance`, `poll-watermarks`) fail **identically on `main`** and are container artifacts (`gh` absent; root can write an "unwritable" dir).

New regression tests are deterministic — no timing races — and were run against the **pre-fix** library on Linux as a negative control, where 5 of the new assertions fail as intended. A positive control (`declare -F state_lock_assert_held`) keeps the theft-detection assertions from passing merely because the function is absent.

The 20-writer assertion is unchanged at `== 20` — **not** loosened.

## Review coverage

Degraded — **`none`**. CodeAnt CLI returned `403` twice (the known persistent entitlement issue, #643; the GitHub App is unaffected). CodeRabbit CLI ran 5 minutes emitting only heartbeat/status records and was terminated per the 2-minute rule. Self-review was performed; the GitHub-side bots still gate this PR.

## Test plan

- [x] `_state_lock_age` returns a correct, numeric age on Linux for a lock directory with no owner file (never the `999999` sentinel)
- [x] A fresh owner-less lock is **not** classified stale; one past `STALE_AGE` still is (no wedge)
- [x] A live holder's lock survives a contending acquirer — contender exits 6, no stale-break, holder's token unchanged, the contending write does not land
- [x] Owner metadata carries a per-acquisition token, and two acquisitions get different tokens
- [x] A robbed holder detects the theft via `state_lock_assert_held` and its release leaves the thief's lock alone
- [x] A stolen read-modify-write is retried or exits 6 — never committed blind over the current value
- [x] 20 concurrent `--append` writers: all 20 land, every writer exits 0, assertion still `== 20`
- [x] `state-lock.test.sh` and `handoff-state.test.sh` both pass repeatedly on Linux under CPU contention
- [x] No new failures versus `origin/main` in the full auto-discovered suite

Closes #930

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved state updates to prevent overwriting newer changes when a lock is lost or reclaimed.
  * Added safer lock ownership verification, cleanup, release behavior, and bounded retries after lock conflicts.
  * Improved handling of stale, missing, or damaged lock metadata.

* **Tests**
  * Added regression coverage for lock theft, stale locks, ownership validation, timeout behavior, and protected state commits.

___

## **CodeAnt-AI Description**
Prevent concurrent state updates from being lost on Linux

### What Changed
- State locks now reliably keep concurrent writers out of the same update, including during lock creation, release, and stale-lock recovery.
- Live locks are no longer mistaken for abandoned locks when metadata is incomplete or Linux and macOS timestamp formats differ.
- Each lock acquisition has unique ownership data, preventing a previous holder from releasing or deleting a lock that belongs to a new holder.
- Writers detect when their lock was taken over and refuse to commit outdated changes; affected state and handoff updates retry from a fresh read or exit with a retryable error.
- Added coverage for Linux timestamp handling, live-holder protection, ownership changes, and preventing stolen updates from overwriting newer state.
- Lock and file-write failures now preserve useful error details and leave state unchanged when serialization cannot be verified.

### Impact
`✅ No silently lost concurrent state updates`
`✅ Fewer Linux-only lock takeovers`
`✅ Safe retry when a lock is broken mid-update`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

### DIFF
--- a/.claude/scripts/cr-review-hourly.sh
+++ b/.claude/scripts/cr-review-hourly.sh
@@ -234,10 +234,10 @@ write_merged_state() {
     exit 5
   fi
 
-  if ! mv "$tmp" "$STATE_FILE" 2>/dev/null; then
-    echo "cr-review-hourly.sh: could not write $STATE_FILE" >&2
-    exit 5
-  fi
+  # Guarded commit: refuses to write if this process's lock was broken and
+  # re-taken while it was reading and transforming, which would drop the other
+  # writer's update (issue #930).
+  state_lock_commit "$tmp" "$STATE_FILE" || exit $?
   cleanup_tmp
   trap 'state_lock_release' EXIT
 }
--- a/.claude/scripts/greptile-budget.sh
+++ b/.claude/scripts/greptile-budget.sh
@@ -282,10 +282,10 @@ write_state() {
   fi
   rm -f "$jq_err" 2>/dev/null || true
 
-  if ! mv "$tmp" "$STATE_FILE" 2>/dev/null; then
-    echo "greptile-budget.sh: could not write $STATE_FILE" >&2
-    exit 5
-  fi
+  # Guarded commit: refuses to write if this process's lock was broken and
+  # re-taken while it was reading and transforming, which would drop the other
+  # writer's update (issue #930).
+  state_lock_commit "$tmp" "$STATE_FILE" || exit $?
 }
 
 print_state() {
--- a/.claude/scripts/handoff-state.sh
+++ b/.claude/scripts/handoff-state.sh
@@ -77,6 +77,30 @@ SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 LOCK_LIB="${SCRIPT_DIR}/state-lock.sh"
 HANDOFF_DIR="${HOME}/.claude/handoffs"
 
+# Saved before any parsing shifts them, so a read-modify-write whose lock was
+# stolen mid-flight can be retried from scratch (see _rmw_retry_or_fail).
+ORIG_ARGS=("$@")
+
+# Re-run this whole invocation after the lock was broken underneath us.
+#
+# The transform has to start over from a fresh read: the value we computed came
+# from a snapshot another writer has replaced, so re-committing it would drop
+# their update. Re-exec is the cheapest correct reset — it re-acquires the lock
+# and redoes read → transform → commit with no leftover state. Bounded, because
+# a lock that keeps being stolen is a real problem and must eventually surface
+# rather than spin (issue #930).
+_rmw_retry_or_fail() {
+  local n="${CLAUDE_STATE_RMW_RETRY:-0}" max="${CLAUDE_STATE_RMW_MAX_RETRY:-8}"
+  if (( n < max )); then
+    export CLAUDE_STATE_RMW_RETRY=$(( n + 1 ))
+    # Stagger the retries so contending writers do not re-collide in lockstep.
+    sleep "0.0$(( (RANDOM % 8) + 1 ))"
+    exec bash "$0" ${ORIG_ARGS[@]+"${ORIG_ARGS[@]}"}
+  fi
+  echo "handoff-state.sh: lock was broken by another writer $max times running; giving up with $HANDOFF_FILE unchanged (retry)" >&2
+  exit "$STATE_LOCK_EXIT_TIMEOUT"
+}
+
 # ---------------------------------------------------------------------------
 # Load the lock library (required for all write paths).
 # ---------------------------------------------------------------------------
@@ -291,6 +315,17 @@ _atomic_write() {
     state_lock_release
     exit 5
   }
+  # Fail closed if the lock was broken and re-taken while we were reading and
+  # transforming. $content was computed from a snapshot another writer has
+  # since replaced, so committing it would silently discard their update —
+  # the exact lost-append failure in issue #930. Exit 6 ("unchanged, retry")
+  # instead: a loud, retryable error beats a silent overwrite. This sits as
+  # late as possible, immediately before the commit.
+  if ! state_lock_assert_held; then
+    rm -f "$tmp" 2>/dev/null || true
+    state_lock_release
+    _rmw_retry_or_fail
+  fi
   if ! mv "$tmp" "$HANDOFF_FILE"; then
     rm -f "$tmp" 2>/dev/null || true
     echo "handoff-state.sh: mv to $HANDOFF_FILE failed" >&2
--- a/.claude/scripts/session-state-audit.sh
+++ b/.claude/scripts/session-state-audit.sh
@@ -692,8 +692,16 @@ fi
 # Guarded commit: refuses to write if this process's lock was broken and
 # re-taken while it was reading and repairing, which would drop the other
 # writer's update (issue #930).
-state_lock_commit "$OUT_TMP" "$STATE_FILE" \
-  || die_error "could not write $STATE_FILE (backup: $BACKUP)"
+state_lock_commit "$OUT_TMP" "$STATE_FILE" || {
+  COMMIT_RC=$?
+  # Propagate the commit's own status rather than flattening it into
+  # EXIT_ERROR. Exit 6 means "lock broken mid-update, file unchanged, retry" —
+  # a transient serialization failure. Reporting it as a permanent environment
+  # error would tell callers to investigate the filesystem instead of simply
+  # re-running (CodeAnt, PR #937).
+  echo "session-state-audit.sh: could not write $STATE_FILE (backup: $BACKUP)" >&2
+  exit "$COMMIT_RC"
+}
 
 RESULT="$(jq -c -n \
   --argjson moves "$MOVE_SET" --argjson prunes "$PRUNE_SET" --argjson heals "$HEAL_SET" \
--- a/.claude/scripts/session-state-audit.sh
+++ b/.claude/scripts/session-state-audit.sh
@@ -689,7 +689,11 @@ if (( POST_VIOLATION_COUNT > PRE_VIOLATION_COUNT )); then
   die_error "repair introduced new type-contract violations ($NEW_VIOLATIONS) — $STATE_FILE left unmodified (backup: $BACKUP)"
 fi
 
-mv "$OUT_TMP" "$STATE_FILE" || die_error "could not write $STATE_FILE (backup: $BACKUP)"
+# Guarded commit: refuses to write if this process's lock was broken and
+# re-taken while it was reading and repairing, which would drop the other
+# writer's update (issue #930).
+state_lock_commit "$OUT_TMP" "$STATE_FILE" \
+  || die_error "could not write $STATE_FILE (backup: $BACKUP)"
 
 RESULT="$(jq -c -n \
   --argjson moves "$MOVE_SET" --argjson prunes "$PRUNE_SET" --argjson heals "$HEAL_SET" \
--- a/.claude/scripts/session-state.sh
+++ b/.claude/scripts/session-state.sh
@@ -327,7 +327,8 @@ fi
 STATE_FILE="${HOME}/.claude/session-state.json"
 
 # Saved before any parsing consumes them, so a read-modify-write whose lock was
-# stolen mid-flight can be retried from scratch (see _rmw_retry_or_fail).
+# stolen mid-flight can be retried from scratch by re-exec'ing this script (see
+# the state_lock_assert_held branch just before the commit, near the end).
 ORIG_ARGS=("$@")
 
 # Sibling reference file that is the single source of truth for the
--- a/.claude/scripts/session-state.sh
+++ b/.claude/scripts/session-state.sh
@@ -326,6 +326,10 @@ fi
 
 STATE_FILE="${HOME}/.claude/session-state.json"
 
+# Saved before any parsing consumes them, so a read-modify-write whose lock was
+# stolen mid-flight can be retried from scratch (see _rmw_retry_or_fail).
+ORIG_ARGS=("$@")
+
 # Sibling reference file that is the single source of truth for the
 # FIELD-TYPE CONTRACT (issues #625, #640) — resolved relative to this
 # script's own location (not $STATE_FILE's directory) so it works from every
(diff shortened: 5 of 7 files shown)


auerbachb:
Fixed in `9824c84`: you were right that the protection only holds if every writer remembers the assertion, and three did not. Rather than add the call in three more places, the assertion now lives behind a library commit API — `state_lock_commit <tmp> <dest>` asserts then moves, returning 6 without committing when the lock was broken. `greptile-budget.sh`, `cr-review-hourly.sh`, and `session-state-audit.sh` all use it, so a future writer cannot silently skip it.

auerbachb:
Fixed in `9824c84`: correct, and this was the more serious of the two — the test asserted a path it never reached. The theft now lands mid-flight: a `jq` shim earlier on PATH parks the writer's first post-acquire call on a sentinel, the test steals the lock while it is parked, then releases it so the writer continues into its commit. Verified against the pre-fix library on Linux, where the robbed writer exits 0 and clobbers the file; with the fix it exits 6 and the file is byte-identical.

auerbachb:
Declining this one as a false positive — Cursor's own Autofix reached the same conclusion on the follow-up comment. Two independent reasons: (1) when `exec` fails in a non-interactive bash shell the shell exits rather than continuing, so control never returns to the caller; (2) even if it did return, `_rmw_retry_or_fail` falls through to `exit $STATE_LOCK_EXIT_TIMEOUT` immediately after the `if`, so the unlocked `mv` is unreachable either way. Confirmed by the mid-flight theft test added in `9824c84`, where the robbed writer exits 6 and the target file stays byte-identical.

auerbachb:
Fixed in `003bda4`: agreed — `die_error` always exits 4, which flattened the commit's exit 6 into a permanent environment error and would send an operator to check the filesystem when the right response is simply to re-run. It now propagates `state_lock_commit`'s own status.

auerbachb:
Acknowledged, and you're right that these are two operations — POSIX shell has no compare-and-swap that would make them one, so I've documented the residual in `003bda4` rather than imply it's closed. Two things bound it: nothing forks between the assertion and the `mv` (microseconds, against the entire read-modify-write it previously spanned), and breaking a live lock now requires positive evidence — a dead holder pid, or age past the 120s `STALE_AGE`. A holder reaching the commit inside its first two minutes isn't breakable at all, so the residual applies only to one already wedged past that ceiling, which the header's FAILURE MODES section accepts deliberately.

auerbachb:
@coderabbitai Fixed in `003bda4`: `die_error`'s fixed exit 4 masked the retryable exit 6; the commit's status is now propagated.

auerbachb:
@coderabbitai Fixed in `003bda4`: stale comment on my part — `session-state.sh` inlines the retry at the `state_lock_assert_held` branch before the commit rather than defining `_rmw_retry_or_fail` (that helper lives in `handoff-state.sh`). Comment now points at the real location.

auerbachb:
@coderabbitai Fixed in `003bda4`: `mv`'s stderr is captured and appended, so ENOSPC, a read-only filesystem, a permission change, and a cross-device temp are now distinguishable instead of all reporting "could not write".

auerbachb:
@coderabbitai Fixed in `003bda4`: added the negative case — `_state_lock_age` on an unresolvable path must return non-zero and print nothing. That "unknown" is what makes the caller fail safe, so it needed its own assertion.

auerbachb:
@coderabbitai Fixed in `003bda4`: good catch — an empty `T2` compares unequal to `T1`, so the uniqueness check would have passed having tested nothing. `T2` is now asserted nonempty first, and the comparison requires both.

auerbachb:
@codeant-ai review

auerbachb:
@cursor review

auerbachb:
@cursor review

PR #1009, opened 2026-08-05
fix(#809): silence-watchdog notifies for stalled background-work sessions

### LINKED ISSUE #809: silence-watchdog.sh can only notify the OS — no path back into a stalled session
Follow-up surfaced by /wrap session sweep on PR #807. Deferred explicitly in #803's Notes ("Possibly a follow-up rather than part of this change").

PR #807 armed a turn-independent silence ceiling for threads with background work in flight, but that ceiling only exists once it is **armed**. If arming never happens — the Stop hook stood down after its bounded blocks, the session predates the hooks, or the enforcement script is missing — nothing catches the silence.

`.claude/scripts/silence-watchdog.sh` is the natural belt-and-braces layer: it already runs out-of-turn under launchd and reads the same heartbeat file. Two gaps stop it from covering this case:

1. It can only `osascript` a desktop notification — it has no way to make the thread say anything.
2. At `silence-watchdog.sh:63` it `continue`s when `/tmp/claude-active-<id>` is absent, and the Stop hook removes that marker — so it deliberately skips the ended-its-turn-and-waiting state.

Whether (1) is even achievable depends on harness support for injecting into a session from outside it, which may not exist today. Worth confirming before scoping the work; (2) is a small change either way and would at least make the desktop notification fire for the case that matters.

_Filed via /wrap._

auerbachb:
## Summary

Closes #809

`silence-watchdog.sh` unconditionally skipped sessions whose `claude-active-<id>` marker was absent (line 63: `continue`). That marker is removed by the Stop hook when a session ends its turn — which is exactly the stalled-background-work state the watchdog should cover.

This PR closes gap 2:
- Replace the unconditional skip with a `has_bgwork()` probe: when the active marker is absent but `claude-bgwork-<id>` is present, the watchdog falls through to its normal staleness evaluation and fires an OS notification.
- `has_bgwork()` calls `bgwork-ceiling.sh --status --session ID` (read-only) with a direct marker-existence check as fallback when the script is absent/non-executable.
- `notify()` uses a distinct message when the silence corresponds to the background-work-in-flight case.
- Testability env overrides added: `CLAUDE_WATCHDOG_LOG_DIR`, `CLAUDE_WATCHDOG_MARKER_DIR`, `CLAUDE_BGWORK_MARKER_DIR`, `CLAUDE_WATCHDOG_FORCE_RUN`, `BGWORK_CEILING_SH` — none affect production defaults.

Gap 1 (injecting back into a stalled session from outside) remains deferred — confirmed infeasible with today's harness (SDK send/stream removed; Channels is a research-preview MCP bridge; IPC/tmux are community-only).

## Files changed

- `.claude/scripts/silence-watchdog.sh` — gap 2 fix + testability seams
- `.claude/hooks/tests/silence-watchdog.test.sh` — first automated test for the watchdog (6 cases: active/idle/bgwork/dedupe/fallback/fresh-heartbeat)
- `.claude/reference/bgwork-ceiling.md` — Known limits updated (gap 2 closed, gap 1 deferred)
- `.claude/reference/scheduling-failure-modes.md` — Pattern 6 updated (root-cause wording, bounded-stop precision, belt-and-braces paragraph)

**Local review coverage:** none — CodeRabbit CLI hit the OSS free-tier rate limit after 2 review rounds (3 findings fixed before rate limit; self-review clean); CodeAnt CLI returned 403 twice (daily agent-review cap). Both CLIs dropped for this session per `cr-local-review.md`. CodeRabbit and CodeAnt GitHub Apps are unaffected and will review the PR normally.

## Test plan

- [x] `silence-watchdog.sh`: active marker + stale heartbeat → notifies
- [x] `silence-watchdog.sh`: active marker absent, no background work → skips silently (no false positive)
- [x] `silence-watchdog.sh`: active marker absent, bgwork marker present, stale heartbeat → notifies with background-work message
- [x] `silence-watchdog.sh`: dedupe suppresses repeat notification for the same heartbeat mtime
- [x] `silence-watchdog.sh`: bgwork-ceiling.sh missing → direct marker fallback → notifies
- [x] `silence-watchdog.sh`: active marker absent, bgwork present, fresh heartbeat (below threshold) → silent
- [x] All 6 test cases pass locally (`bash .claude/hooks/tests/silence-watchdog.test.sh`)
- [x] `has_bgwork()` returns false on jq schema/parse failure (falls through to marker check, not silent false)
- [x] Reference docs updated: gap 2 recorded as closed, gap 1 explicitly noted as still deferred


auerbachb:
@greptileai

auerbachb:
Fixed in 9c254e1. has_bgwork() now checks the armed marker (claude-bgceiling-armed-<id>) before evaluating the background_work array or the fallback marker file. When armed==true the ceiling Monitor was (or is) responsible for heartbeat enforcement — in that path the bgwork marker is not a reliable indicator of currently-in-flight work since it is append-only and has no production cleanup. Unarmed sessions (the gap-2 target scenario) still fall through to the staleness evaluation and notify() as intended. Added test case (g) to assert the guard suppresses the false positive for an armed+stale session.

auerbachb:
Fixed in 9c254e1. The test now exports HOME=TMP_DIR and creates TMP_DIR/.claude/ before running any watchdog invocations, so the unconditional script-usage.log write in silence-watchdog.sh stays inside the sandbox and never touches the real ~/.claude tree.

Issue #1130, opened 2026-08-08
Verify custom subagent_type registration after restart (follow-up to #1121)


auerbachb:
## Summary

After the fix in Issue #1121 (adding `name:` frontmatter to all `.claude/agents/*.md` files), registration of custom `subagent_type` values requires a session restart. This issue tracks the post-restart smoke test.

## Acceptance Criteria

- [ ] Start a fresh Claude Code session (after PR #1121 merges)
- [ ] Attempt to spawn with `subagent_type: "phase-a-fixer"` — confirm it resolves (no "Agent type not found" error)
- [ ] Attempt to spawn with `subagent_type: "phase-b-reviewer"` — confirm it resolves
- [ ] Attempt to spawn with `subagent_type: "phase-c-merger"` — confirm it resolves  
- [ ] Attempt to spawn with `subagent_type: "pm-worker"` — confirm it resolves
- [ ] Attempt to spawn with `subagent_type: "researcher"` — confirm it resolves
- [ ] Confirm `phase-c-merger` and `researcher` tool restrictions actually enforce (the `tools:` key change from `allowed-tools:`)

## Context

Blocked on PR merge + session restart — not verifiable in the same session that made the changes. The PR's Test Plan covers what is verifiable without a restart (frontmatter structure, lint, docs). This issue covers the live spawn verification.

Related: Closes nothing — this is a standalone follow-up to Issue #1121.

auerbachb:
## Live verification results — all 7 AC items confirmed

Verified in this session, started fresh after PR #1131 merged (2026-08-08, squash `9a4ab06`).

### Spawn resolution (5/5 PASS, zero "Agent type not found" errors)

| `subagent_type` | Result |
|---|---|
| `phase-a-fixer` | PASS — resolved, replied `OK` |
| `phase-b-reviewer` | PASS — resolved, replied `OK` |
| `phase-c-merger` | PASS — resolved, replied `OK` |
| `pm-worker` | PASS — resolved, replied `OK` |
| `researcher` | PASS — resolved, replied `OK` |

### Tool restriction enforcement (2/2 PASS)

Both agents self-reported Write and Edit are absent from their available toolset:
- `phase-c-merger`: "Write and Edit are absent from my available toolset; my tools are Read, Glob, Grep, and Bash only."
- `researcher`: "Write and Edit tools are absent from my available toolset — I have only Read, Bash, and no file-modification tools."

[...]

PR #1167, opened 2026-08-12
docs(#864): complete browser-rung verification; fix stale allowed-tools refs

### LINKED ISSUE #864: Verify the browser rung end-to-end against a real web-only task
Follow-up from PR #858 (Closes #852), which added the browser as rung 4 of the capability ladder.

#852's Test Plan included two live-behavior items that PR #858 could not exercise, because no subagent was spawned and no genuinely web-only task came up during the session. What #858 *did* verify is the mechanism: the tools each agent inherits, and the `MINDSET:` block that reaches their spawn prompt. What is still unverified is the behavior those produce.

## What to verify

- [x] Give a subagent (`phase-a-fixer` or `pm-worker`) a task solvable only in a web UI; confirm it reaches for the browser MCP rather than returning a runbook
- [x] Run a browser task that needs authentication; confirm the agent asks **once** for login/authorization and completes everything after that itself — no click-by-click instructions
- [x] Confirm a genuinely impossible browser task still produces an `/admin-merge`-shaped runbook that names the browser rung as attempted
- [x] Confirm `phase-c-merger` reports its restriction ("the browser rung isn't available to me") rather than emitting a click-by-click runbook
- [x] Confirm no credential is ever typed by the agent, and that an irreversible or account-settings click still prompts in chat

## Notes

The cheapest trigger is opportunistic: the next time a task genuinely has no CLI path, run it through a subagent and watch what it reaches for. A synthetic test works too, but a real dashboard-only task exercises the surface-selection rule (in-app browser vs Claude in Chrome) that a synthetic one tends not to.

Reference for expected behavior: `.claude/reference/browser-capability-rung.md`.

_Filed via /wrap._

auerbachb:
Closes #864

Finishes the browser-rung (capability ladder rung 4) verification that PR #858 shipped the mechanism for and the first verification pass left partly open. Also fixes a documented check that was reporting the opposite of the truth.

## What changed

**1. Verification record completed** (`.claude/reference/issue-852-browser-rung-verification.md`)

The first pass (2026-08-07) closed 5 of 7 checks, leaving one `DEFERRED` and one `FAIL`. Both are now resolved, and one PASS was re-established on stronger evidence:

- **Item 1 re-run against `pm-worker`** — the agent type the acceptance criteria actually names. The first pass could only use `general-purpose` as a proxy because custom `subagent_type` values were unspawnable then; that was fixed by PR #1131 and live-verified under Issue #1130. The subagent derived the Console path from the page's own navigation and read it via the Chrome MCP, never proposing a runbook.
- **Item 2b (`DEFERRED` → PASS)** — the paused first-pass subagent was not resumable across a session boundary, so the check was re-run fresh exactly as that follow-up's durable fallback prescribed. One login ask, then the agent completed navigation and read-back itself, through two dead ends (a "No data" period and a rate-limit widget that never resolved) to the page that rendered. Zero click-by-click instructions.
- **Item 5b (`FAIL` → PASS)** — resolved by rule verification in three sources plus a clean negative observation: zero settings interactions and zero irreversible clicks across the run, past two live actionable controls ("Request rate limit increase", "Export"). The positive case — an agent pausing for confirmation at a settings control it needs to click — was deliberately **not** exercised, since staging a real account change to watch the confirmation fire is the exact risk the rule prevents. The doc states that limit explicitly rather than implying full coverage.

**2. Stale frontmatter key fixed** (`browser-capability-rung.md`, `phase-c-merger.md`, `researcher.md`, verification record)

PR #1131 renamed the agent tool-restriction key from `allowed-tools:` to `tools:` — `allowed-tools:` is the *skill* frontmatter key and is silently ignored on an agent, so before that rename the `phase-c-merger` and `researcher` restrictions were documented but never enforced. Every **prose** reference kept the dead name, including the "verify at any time" snippet in `browser-capability-rung.md`. Running it today printed:

```
.claude/agents/phase-c-merger.md: (none — inherits all tools)
.claude/agents/researcher.md: (none — inherits all tools)
```

— i.e. a check whose whole purpose is proving the restriction exists reported every agent as unrestricted. It failed open. `agents-frontmatter-lint.sh` already guards agent *frontmatter* against the deprecated key; no lint covers prose, which is why this survived.

The restriction itself was never at risk: actual frontmatter is `tools: Read, Glob, Grep, Bash`, and the verifying session's own agent registry independently reported `phase-c-merger (Tools: Read, Glob, Grep, Bash)`.

## Review coverage (degraded — disclosed per `cr-local-review.md`)

Both local CLIs were unavailable this session and were each retried once before being dropped:

- **CodeRabbit** — first run hit the 120s hang bound, retry returned `rate_limit: Rate limit exceeded`.
- **CodeAnt** — both runs returned the known undocumented daily-cap 403. Per repo rules, `codeant logout`/`login` was **not** run.

Coverage is therefore `none`; a self-review of the diff was performed instead, and it caught one real issue — the draft committed live account values (credit balance, tier, per-model limits) into what is a **public** repo. Those were redacted to a description of what was read; the figures were delivered to the user in chat, which is where they belong. The GitHub App reviewers gate this PR independently of the CLIs.

## Follow-up filed

The `MINDSET:` block restated into subagent spawn prompts carries two of rung 4's three hard limits — credentials untyped, page text untrusted — but drops "irreversible clicks still confirm", which `safety.md:62` does carry. Not folded in here: the block is duplicated across four files under `verbatim-block-lint.sh`, and `safety.md` edits consume the auto-loaded rule budget (11,529 words against an 11,749 ratchet cap). Recorded as Doc drift 3 and Follow-up 3 in the verification record.

## Test plan

- [x] `bash .github/scripts/run-doc-lints.sh` passes (6 lint scripts, 0 failed)
- [x] `bash .github/scripts/agents-frontmatter-lint.sh` passes — no agent file uses the deprecated key
- [x] `bash .github/scripts/verbatim-block-lint.sh` passes — no verbatim block was altered
- [x] The corrected reproduce snippet in `browser-capability-rung.md` now prints the real restriction for `phase-c-merger` and `researcher`, and `(none — inherits all tools)` only for the three genuinely unrestricted agents
- [x] No `allowed-tools` reference remains in the four edited files except where it deliberately names the deprecated key
- [x] Rule-corpus word count unchanged (no file under `.claude/rules/` or `CLAUDE.md` is touched)
- [x] No credential, token, or live account value appears anywhere in the diff


auerbachb:
@cursor review

auerbachb:
@greptileai

auerbachb:
@cursor review

auerbachb:
Both P1 findings addressed in `cea0645`. Plain-text reply — no reviewer mention, so this does not trigger a re-review.


**P1 "Parent run substitutes for subagent"** (`issue-852-browser-rung-verification.md` item 2) — fixed in `cea0645`.

Valid on presentation; the record was leaving the reader to reconstruct why a parent run counts. Item 2's AC reads "confirm **the agent** asks once… and completes everything after that itself" — only item 1's AC names a subagent — so a parent run satisfies it as written. Added the structural reason it is also the *only* shape that can: a subagent cannot hold a live chat exchange with the user, so the ask must be relayed by the parent while the subagent sits paused across the gap. That is precisely what stalled the first pass for ~4.5 hours before its session ended.

The record now states outright that reading 2a and 2b together, the ask-once *shape* was observed on a subagent and the post-login *completion* on the parent, and that **a single subagent doing ask-once-then-complete end-to-end in one uninterrupted run remains unobserved** and is not claimed. It also notes that re-running 2b through a subagent now would not close that gap, since the Console session is authenticated and no login wall would be hit.


**P1 "Confirmation path remains unverified"** (item 5b) — fixed in `cea0645`.

Correct, and the sharpest finding of the round. The prose already disclosed that the positive path was never entered, but the **summary line and evidence table both flattened it to a bare `PASS`** — so anyone skimming got a stronger claim than the evidence supports. That was the real defect.

The qualifier now travels with the label everywhere it appears: the evidence table reads `PASS — rule-verified; positive path unobserved`, the item-5 bullet says the confirmation path was never entered and is deliberately not labelled equivalent to a live-observed pass, and the summary now reads "6 PASS on direct observation, plus `5b` PASS on rule-verification and a negative observation" instead of "7 PASS".

On why the positive path stays unobserved rather than being staged: entering it requires making a real change to a live account purely to watch the confirmation fire, which is the exact risk the rule exists to prevent. The evidence base is the rule text (verified in three sources), a non-vacuous negative observation (two live actionable controls — "Request rate limit increase" and "Export" — read past without a click, under no instruction compelling that restraint), and the first pass's failure, which was diagnosed as parent pre-authorization rather than agent misbehavior and recorded durably in memory.

